PR #113, opened 2020-07-15
add --password option to `profile create` and `profile update` commands

timabrmsn:
This PR allows the password to be passed on profile creation and update commands, bypassing the interactive prompt.

### DIFF
--- a/src/code42cli/cmds/profile.py
+++ b/src/code42cli/cmds/profile.py
@@ -83,7 +83,7 @@ def create(name, server, username, password, disable_ssl_errors):
 def update(name, server, username, password, disable_ssl_errors):
     """Update an existing profile."""
     c42profile = cliprofile.get_profile(name)
-    cliprofile.update_profile(profile.name, server, username, disable_ssl_errors)
+    cliprofile.update_profile(c42profile.name, server, username, disable_ssl_errors)
     if password:
         _set_pw(name, password)
     else:
--- a/src/code42cli/cmds/profile.py
+++ b/src/code42cli/cmds/profile.py
@@ -21,14 +21,18 @@ name_option = click.option(
     "-n",
     "--name",
     required=True,
-    type=str,
     help="The name of the Code42 CLI profile to use when executing this command.",
 )
 server_option = click.option(
-    "-s", "--server", required=True, type=str, help="The url and port of the Code42 server."
+    "-s", "--server", required=True, help="The url and port of the Code42 server."
 )
 username_option = click.option(
-    "-u", "--username", required=True, type=str, help="The username of the Code42 API user."
+    "-u", "--username", required=True, help="The username of the Code42 API user."
+)
+password_option = click.option(
+    "--password",
+    help="The password for the Code42 API user. Only recommended if command needs to be run without "
+    "prompt interactions.",
 )
 disable_ssl_option = click.option(
     "--disable-ssl-errors",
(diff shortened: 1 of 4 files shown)


src/code42cli/cmds/profile.py
```diff
@@ -57,32 +61,42 @@ def show(profile_name):
...
+        _set_pw(name, password)
+    else:
+        _prompt_for_allow_password_set(profile.name)
     echo("Profile '{}' has been updated.".format(profile.name))
 
 
 @profile.command()
 @profile_name_arg
-def reset_pw(profile_name=None):
+def reset_pw(profile_name):
```
alanag13, line 95: This is not directly related to your change, but I think we should clarify that this command does not actually reset your code42 password and instead simply changes the password that is stored with the profile.

src/code42cli/cmds/profile.py
```diff
@@ -21,14 +21,18 @@ def profile():
...
-    "-s", "--server", required=True, type=str, help="The url and port of the Code42 server."
+    "-s", "--server", required=True, help="The url and port of the Code42 server."
 )
 username_option = click.option(
-    "-u", "--username", required=True, type=str, help="The username of the Code42 API user."
+    "-u", "--username", required=True, help="The username of the Code42 API user."
+)
+password_option = click.option(
+    "--password",
+    help="The password for the Code42 API user. Only recommended if command needs to be run without "
```
alanag13, line 34: The second sentence here kind of sounds like we are recommending not saving a password if possible. I _think_ what we're trying to say is that you can avoid using the flag so that the command doesn't end up in your bash history, so I would maybe make the second sentence something like:

```
if this option is omitted, interactive prompts will be used to obtain the password.
```

alanag13:
@timabrmsn will the new `--assume-yes` work with the prompts that:
- ask to save your password
- ask to fall back to a flat file as the method of password storage?

src/code42cli/cmds/profile.py
timabrmsn: yeah, I struggled with the right way to word this. I like your sentence better.

timabrmsn:
> will the new `--assume-yes` work with the prompts that:
> 
> * ask to save your password
> * ask to fall back to a flat file as the method of password storage?

No, the `@yes_option` has not been added to `profile create`, as the `--password` flag does the bypassing of the password prompt (I just merged those changes in here). 

That means the flat file fallback prompt has to be handled interactively. Do you think we need the `-y` option for that?

alanag13:
@timabrmsn having `password` handle that makes perfect sense. I don't think having `yes_option` play nice with the fallback prompt is super important, I was more just curious.

src/code42cli/cmds/profile.py
```diff
@@ -58,32 +62,42 @@ def show(profile_name):
...
 @username_option
+@password_option
 @disable_ssl_option
-def update(name=None, server=None, username=None, disable_ssl_errors=None):
+def update(name, server, username, password, disable_ssl_errors):
     """Update an existing profile."""
     profile = cliprofile.get_profile(name)
     cliprofile.update_profile(profile.name, server, username, disable_ssl_errors)
-    _prompt_for_allow_password_set(profile.name)
+    if password:
```
unparalleled-js, line 87: Could be extracted into a private function and shared with lines 70-73

timabrmsn: this feels like too small of a unit to get any real benefit from pulling it out since it's only used twice and the logic of the code itself is as readable as any function name we could give it.

PR #116, opened 2020-07-20
Chore/linter tox

alanag13:
similar to the recent change in py42:
- no logic has changed,  the formatter and flake8 rules touched a lot of files
- I mostly would like this review to be of the `CONTRIBUTING` document to make sure it all makes sense
- one of the formatters that auto-sorts imports also breaks up `from` imports into multiple lines, and I think I would rather have this be `import module as m` and then use `m.do_the_thing(), m.do_the_other_thing()` instead of `from module import do_the_thing, do_the_other_thing`, but i'll worry about that in a separate PR.

- templates for issues and PRs and build / code coverage badges have also been added.


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,27 @@
...
+### Expected Behavior
+<!-- What did you expect to happen when you did the above? -->
+
+### Actual Behavior
+<!-- What actually happened when you did the above? -->
+
+### Basic Information
+
+- code42cli version: <!-- get the version using `pip show code42cli`-->
+- python version: <!-- get the version using `python --version`-->
```
timabrmsn, line 26: Since just running `python --version` isn't guaranteed to actually be what the cli was installed under, I wonder if we should have `code42 -h` also print the python version, then we could get accurate cli version, py42 version, and python version just by having them do `code42 -h` for us.

alanag13: good call, I like that

src/code42cli/cmds/legal_hold.py
```diff
@@ -140,8 +143,10 @@ def bulk(state):
 @sdk_options
 def add(state, csv_rows):
```
timabrmsn, line 144: These should also change to `bulk_add`/`bulk_remove` to match the others.

alanag13: In the other cases it was _needed_ because the modules had more than one function with the same name, but I agree consistency is good.

CONTRIBUTING.md
```diff
@@ -1,59 +1,124 @@
...
-$ pip install tox
-$ pyenv install 3.5.7
-$ pyenv install 3.6.9
-$ pyenv install 3.7.4
-$ pyenv local 3.5.7 3.6.9 3.7.4
-$ tox
+tox -e style
+```
+
+This will output a diff of the files that were changed as well a list of files / line numbers / error descriptions for any style problems that need to corrected manually. Once these have been corrected and re-pushed, the PR checks should pass.
```
unparalleled-js, line 106: Looks like the word "be" is missing between "to" and "corrected".

src/code42cli/cmds/legal_hold.py
```diff
@@ -138,7 +138,7 @@ def bulk(state):
     name="add",
     help="Bulk add users to legal hold matters from a csv file. CSV file format: {}".format(
         ",".join(LEGAL_HOLD_CSV_HEADERS)
-    )
+    ),
```
unparalleled-js, line 141: Nice demo

CONTRIBUTING.md
```diff
@@ -75,61 +140,29 @@ Example:
...
-
-    ```bash
-    $ code42 cmd oneval twoval --three threeval --four fourval
-    ```
+## Documentation
 
-4. To add descriptions to your cli arguments to appear in the help text, your command takes a function as the `arg_customizer` parameter.
-    The entire [`ArgConfigCollection`](src/code42cli/args.py) that was automatically created is supplied as the only argument to this function
-    and can be modified by it. See `code42cli.cmds.profile._load_profile_create_descriptions` for an example of this.
+Public functions, classes, and methods should have docstrings.
```
unparalleled-js, line 145: Should this be more like `command functions should have docstrings` ?

alanag13: should actually be something like `command functions should have accompanying documentation`

PR #117, opened 2020-07-22
Feature/Option to use OR queries on security-data/alerts search

timabrmsn:
Adds `--or-query` option to make the passed in filters be processed in an OR group for both alert and file event searches.


src/code42cli/cmds/securitydata.py
```diff
@@ -163,9 +163,20 @@ def clear_checkpoint(state, checkpoint_name):
 @security_data.command()
 @file_event_options
 @search_options
+@click.option(
+    "--or-query", is_flag=True, cls=searchopt.AdvancedQueryAndSavedSearchIncompatible
```
alanag13, line 167: I do think there is potential usefulness to not having this restriction, but it probably warrants further discussion and should just stay this way for now. We can discuss further once we pursue "merging" cli options with saved search/advanced search.

alanag13:
Needs tests and testing instructions, but otherwise LGTM.

src/code42cli/cmds/alerts.py
```diff
@@ -155,15 +155,22 @@ def clear_checkpoint(state, checkpoint_name):
...
+    cli_state, format, begin, end, advanced_query, use_checkpoint, or_query, **kwargs
+):
     """Search for alerts."""
     output_logger = logger_factory.get_logger_for_stdout(format)
     cursor = _get_alert_cursor_store(cli_state.profile.name) if use_checkpoint else None
     handlers = ext.create_handlers(
         cli_state.sdk, AlertExtractor, output_logger, cursor, use_checkpoint
     )
     extractor = _get_alert_extractor(cli_state.sdk, handlers)
+    if or_query:
```
unparalleled-js, line 172: any reason not to just do `extractor.use_or_query = or_query` ?

timabrmsn: nope, I think i just copied the logic from security-data and then removed the extra line.

src/code42cli/cmds/securitydata.py
```diff
@@ -176,6 +187,9 @@ def search(
         state.sdk, FileEventExtractor, output_logger, cursor, use_checkpoint
     )
     extractor = _get_file_event_extractor(state.sdk, handlers)
+    if or_query:
```
alanag13, line 190:
```python
extractor.use_or_query = or_query
```

PR #118, opened 2020-07-23
Chore/click autodoc

timabrmsn:
This pulls in the sphinx-click extension to auto-generate documentation for commands. 

Built docs for this branch are here: https://code42cli.readthedocs.io/en/chore-click_autodoc/commands.html

I opened a PR with the `sphinx-click` project to fix the formatting on the option choices when there are lots of them or they're long (see `high-risk-employee add --risk-tag` for example problem). 

Also, since it automatically documented the `--profile` and `--debug` options on every command group, I made a new option decorator to quiet those flags where it makes sense. So now if you do `code42 <cmd_group> -h` you won't get `--debug` or `--profile` descriptions in the help text, but it will still work to pass them in at any position.


timabrmsn:
cool, PR already accepted, so this should be fixed soon: https://github.com/click-contrib/sphinx-click/pull/64

src/code42cli/options.py
```diff
@@ -79,13 +79,44 @@ def set_debug(ctx, param, value):
...
 
 
+profile_option_quiet = click.option(
+    "--profile",
+    expose_value=False,
+    callback=set_profile,
+    hidden=True,
+    help="The name of the Code42 CLI profile to use when executing this command.",
+)
+debug_option_quiet = click.option(
```
alanag13, line 89: Instead of having to duplicate these definitions, I think we could make each of these options and `sdk_options` take a boolean as a parameter to indicate whether or not they are hidden. The default could be true since that's what we would want most of the time. Something like:

```python
def profile_option(hidden=True):
    return click.option(
        "--profile",
        expose_value=False,
        callback=set_profile,
        hidden=is_hidden,
        help="The name of the Code42 CLI profile to use when executing this command.",
    )

def sdk_options(hidden=True):
    f = profile_option(hidden)
    ...
    return f

@sdk_options(hidden=False)
def some_command()
    pass
```

timabrmsn: I like it, and went a step further and make `@sdk_options` take the argument instead of two separate decorators.

alanag13: 💥

PR #121, opened 2020-07-24
Update gettingstarted.md

annie-payseur:
We go into a lot more detail on credential storage (specifically for RHEL) at https://support.code42.com/Administrator/Cloud/Monitoring_and_managing/Set_up_the_Code42_command-line_interface#Credential_storage. Do we need to go into those details here?


alanag13:
@annie-payseur good catch, lets get that info added!

PR #124, opened 2020-07-27
Update detectionlists.md

annie-payseur:
Was it intentional to omit instructions for updating risk factors for multiple users? (Like what we have here: https://support.code42.com/Administrator/Cloud/Monitoring_and_managing/Manage_detection_list_users_with_the_Code42_command-line_interface#Update_risk_factors_for_multiple_users.) This isn't really covered in the command doc either.

### DIFF
--- a/docs/userguides/detectionlists.md
+++ b/docs/userguides/detectionlists.md
@@ -1,6 +1,6 @@
-# Managing Detection List Users
+# Manage Detection List Users
 
-Use the departing-employee commands to add or remove employees on that that list, or update the details for a user. To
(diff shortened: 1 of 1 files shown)


alanag13:
> Was it intentional to omit instructions for updating risk factors for multiple users? (Like what we have here: https://support.code42.com/Administrator/Cloud/Monitoring_and_managing/Manage_detection_list_users_with_the_Code42_command-line_interface#Update_risk_factors_for_multiple_users.) This isn't really covered in the command doc either.

No, I'm sure that was just missed somehow (was probably a timing thing where we created these and IX added the documentation later).

PR #125, opened 2020-07-27
Enable Nightly Builds

timabrmsn:
This creates a "nightly" tox target that installs `py42` and `c42eventextractor` directly from the latest master branch before installing `code42cli` and running unit tests. 

Since `c42eventextractor` is still a private repo, we had to create a deploy key for the repo and save the private key into `code42cli` repo secrets. 

In addition to the scheduled nightly (midnight CST) run, it's possible to trigger this action manually via api due to the `on: workflow_dispatch:` config. While the scheduled run will only build against the master branch, `workflow_dispatch` allows you to manually trigger against a different branch. For example, this is what I used to fire off on this branch to confirm the workflow/tests worked:

```
curl -H "Accept: application/vnd.github.everest-preview+json" \
-H "Authorization: token <your token>" \
--request POST \
--data '{"ref": "chore/nightly_builds"}' \
https://api.github.com/repos/code42/code42cli/actions/workflows/nightly.yml/dispatches
```


.idea/code42cli.iml
```diff
@@ -10,10 +10,10 @@
     <option name="requirementsPath" value="" />
   </component>
   <component name="PyDocumentationSettings">
+    <option name="format" value="GOOGLE" />
```
alanag13, line 13: would you mind .gitignoring this file and removing it from the repo? It unfortunately doesn't at all work the way I hoped and we ended up removing it from py42.

tox.ini
```diff
@@ -3,6 +3,7 @@ envlist =
     py{38,37,36,35,27}
     docs
     style
+    nightly
```
alanag13, line 6: you don't need to add this to `envlist`, this is just the list of things that get iterated through when you enter `tox` (with no `-e`). It will still be available to run without it.

tox.ini
```diff
@@ -31,3 +32,12 @@ commands = sphinx-build -W -b html -d {envtmpdir}/doctress docs {envtmpdir}/html
 deps = pre-commit
 skip_install = true
 commands = pre-commit run --all-files --show-diff-on-failure
+
+
+[testenv:nightly]
+deps =
```
alanag13: right now all this does is run the install and nothing else -- we should add a `commands` section and run the unit tests against this.

tox.ini
timabrmsn: ah, good to know!

tox.ini
timabrmsn: no, it does run them since it inherits the command from the base `testenv`

.idea/code42cli.iml
timabrmsn: sure

tox.ini
alanag13: ah, right you are, I have actually never made a `testenv` where a command wasn't defined so I didn't realize this was the behavior.

PR #128, opened 2020-07-29
Feature/output formats

kiran-chaudhary:
**Description of Change**

Added support for [global output formats](https://jira.corp.code42.com/browse/INTEG-1053)

Added option for command `code42 alert-rules list`

**Issues Resolved**
[1053](https://jira.corp.code42.com/browse/INTEG-1053)

**Testing guidelines:**
`code42 alert-rules list -f csv`
`code42 alert-rules list -f json`
`code42 alert-rules list -f ascii-table`
`code42 alert-rules list -f raw-json`

Added same option for commands:
`code42 security-data  saved-search list`
`code42 legal-hold list`
`code42 legal-hold show`

**PR Checklist**

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,6 @@ setup(
         "keyring==18.0.1",
         "keyrings.alt==3.2.0",
         "py42>=1.7.0",
-        "texttable>=1.6.2",
     ],
     extras_require={
         "dev": [
--- a/setup.py
+++ b/setup.py
@@ -37,6 +37,7 @@ setup(
         "keyring==18.0.1",
         "keyrings.alt==3.2.0",
         "py42>=1.7.0",
+        "texttable>=1.6.2",
     ],
     extras_require={
         "dev": [
--- a/src/code42cli/cmds/enums.py
+++ b/src/code42cli/cmds/enums.py
@@ -1,8 +1,8 @@
 class OutputFormat:
-    ASCII = "ASCII-TABLE"
+    TABLE = "TABLE"
     CSV = "CSV"
     JSON = "JSON"
     RAW = "RAW-JSON"
 
     def __iter__(self):
-        return iter([self.ASCII, self.CSV, self.JSON, self.RAW])
+        return iter([self.TABLE, self.CSV, self.JSON, self.RAW])
--- a/src/code42cli/cmds/enums.py
+++ b/src/code42cli/cmds/enums.py
@@ -1,0 +1,8 @@
+class OutputFormat:
+    ASCII = "ASCII-TABLE"
+    CSV = "CSV"
+    JSON = "JSON"
+    RAW = "RAW-JSON"
+
+    def __iter__(self):
+        return iter([self.ASCII, self.CSV, self.JSON, self.RAW])
--- a/src/code42cli/cmds/securitydata.py
+++ b/src/code42cli/cmds/securitydata.py
@@ -217,8 +217,7 @@ def _list(state, format=None):
     response = state.sdk.securitydata.savedsearches.get()
     header = {"name": "Name", "id": "Id"}
     output = format(response["searches"], header)
-    echo(output, nl=False)
-    echo("")
+    echo(output)
 
 
 @saved_search.command()
--- a/src/code42cli/cmds/securitydata.py
+++ b/src/code42cli/cmds/securitydata.py
@@ -15,7 +15,7 @@ from code42cli.logger import get_main_cli_logger
 from code42cli.options import incompatible_with
 from code42cli.options import OrderedGroup
 from code42cli.options import sdk_options
-from code42cli.output_formats import format_option
+from code42cli.output_formats import format_option as format_output
 
 
 logger = get_main_cli_logger()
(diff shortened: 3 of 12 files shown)


setup.py
```diff
@@ -37,6 +37,7 @@
         "keyring==18.0.1",
         "keyrings.alt==3.2.0",
         "py42>=1.7.0",
+        "texttable>=1.6.2",
```
alanag13, line 40: I would rather not take this on as a dependency. It is independently owned and hasn't been updated in almost a year -- as new versions of python continue to be released it's possible that this library will no longer work.

We already have functionality that we have written in this package for generically printing tables, see `format_to_table`.

src/code42cli/cmds/enums.py
```diff
@@ -0,0 +1,8 @@
+class OutputFormat:
+    ASCII = "ASCII-TABLE"
```
alanag13, line 2: for the user's sake, lets call this `table` instead of `ascii-table`.

src/code42cli/output_formats.py
```diff
@@ -0,0 +1,65 @@
...
+    lines = []
+    lines.append(columns)
+    for row in output:
+        items = [str(row[key]) for key in header.keys()]
+        line = ",".join(items)
+        lines.append(line)
+    return "\n".join(lines)
+
+
+def to_ascii_table(output, header):
```
alanag13, line 42: use `find_format_width` and `format_to_table` in here instead of using `Texttable`. It probably also makes sense to move the `find_format_width` and `format_to_table` functions into this module.

src/code42cli/cmds/enums.py
unparalleled-js: Many tables are ascii tables.

tests/test_output_formats.py
```diff
@@ -0,0 +1,127 @@
...
+
+TEST_HEADER = OrderedDict()
+TEST_HEADER["observerRuleId"] = "RuleId"
+TEST_HEADER["name"] = "Name"
+TEST_HEADER["severity"] = "Severity"
+TEST_HEADER["type"] = "Type"
+TEST_HEADER["ruleSource"] = "Source"
+TEST_HEADER["isEnabled"] = "Enabled"
+
+TABLE_OUTPUT = """RuleId                                 Name                     Severity   Type                          Source     Enabled
```
kiran-chaudhary, line 73: The style guide is complaining that there are trailing spaces at the end of line. Tried adding `"...\n"` approach too but then tests weren't happy. Need suggestions to fix this silly thing or shall I just change the function to not add any padding after the last column?

setup.py
kiran-chaudhary: Removed.

src/code42cli/cmds/enums.py
kiran-chaudhary: Done.

src/code42cli/output_formats.py
kiran-chaudhary: done.

tests/test_output_formats.py
kiran-chaudhary: Fixed now.

kiran-chaudhary:
We can either consider a logical end of the task once code review is approved and continue the work in another branch. Let me know your thoughts. 

Now the commands that can be considered for adding format options, either have nested json response or use `extractor`, hence we can pick them up separately as it needs discussion on what fields should be displayed, etc.

src/code42cli/cmds/securitydata.py
```diff
@@ -15,8 +15,8 @@
 from code42cli.options import incompatible_with
 from code42cli.options import OrderedGroup
 from code42cli.options import sdk_options
-from code42cli.util import find_format_width
-from code42cli.util import format_to_table
+from code42cli.output_formats import format_option as format_output
```
unparalleled-js, line 18: Why rename the import here?

src/code42cli/output_formats.py
```diff
@@ -0,0 +1,59 @@
...
+    rows, column_size = find_format_width(output, header)
+    return format_to_table(rows, column_size)
+
+
+def _filter(output, header):
+    return [{header[key]: row[key] for key in header.keys()} for row in output]
+
+
+def to_json(output, header=None):
+    return json.dumps(_filter(output, header))
```
unparalleled-js, line 55: Should we also use `indent` param on `dumps()` to make it prettier?

unparalleled-js: Nevermind, it looks indented somehow already

src/code42cli/util.py
```diff
@@ -69,10 +69,13 @@ def find_format_width(record, header):
 
 def format_to_table(rows, column_size):
     """Prints result in left justified format in a tabular form."""
```
unparalleled-js, line 71: This comment needs to be changed, as it no longer prints the results but just returns them instead

src/code42cli/cmds/securitydata.py
kiran-chaudhary: Another option exists here with same name, used in `search` subcommand. It doesn't support same formatting options as of now.
I am planning to touch those in a separate branch as I need more information on what fields should be considered to be displayed in the output for such commands.

src/code42cli/util.py
kiran-chaudhary: Changed.

PR #130, opened 2020-07-29
Feature/file cat

unparalleled-js:
* Add support for searching file events using `FileCategory`.
* Uses consts/enum from py42 and removes duplicated values in this project

This is a draft, as this will require releasing py42 and updating the minimum version requirement in `setup.py`.

closes INTEG-1072

To test, turn on debug mode and make a few queries using FileCategory and see the filter go through the wire. Also, could verify results using console.


src/code42cli/cmds/securitydata.py
```diff
@@ -58,7 +60,7 @@
     "-t",
     "--type",
     multiple=True,
-    type=click.Choice(list(enum.ExposureType())),
+    type=click.Choice(list(ExposureType.choices())),
```
timabrmsn, line 63: don't need to call `list()` on these anymore.

PR #133, opened 2020-08-07
Feature/output formats

kiran-chaudhary:
**Description of Change**

Added formatting options to extraction/search commands.
`code42 alerts search` 
`code42 security-data search`
`code42 security-data search --saved-search`

**Test Guidelines:** 

```
$ code42 security-data search -b 2020-07-10
$ code42 security-data search -b 2020-07-10 --include-all 
$ code42 security-data search -b 2020-07-10 --include-all -f json
$ code42 security-data search -b 2020-07-10 --include-all -f raw-json
$ code42 security-data search -b 2020-07-10 --include-all -f csv
$ code42 security-data search -b 2020-07-10 --include-all -f table
$ code42 security-data search -b 2020-07-10 -f table
$ code42 security-data search -b 2020-07-10 -f csv
$ code42 security-data search -b 2020-07-10 -f raw-json
$ code42 security-data search -b 2020-07-10 -f json
```

Similarly added same argument options for `alerts` and `saved-search` commands
```
$ code42 security-data search --saved-search a083f08d-8f33-4cbd-81c4-8d1820b61185
$ code42 alerts search -b 2020-07-10
```

**PR Checklist**

- [x]   Add unit tests to verify this change

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,11 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
     each time the CLI is run with a profile configured this way, as it is not recommended.
 
 - The `path` positional argument for bulk `generate-template` commands is now an option (`--p/-p`).
+- Below `search` subcommands accept argument `--format/-f` to display result in formats `csv`, `table`, `json`, `formatted-json`:
+    By default only root/top level items from the response will be displayed, to view all items, that contains nested data, pass `--include-all` option.
+    - `code42 alerts search`
+    - `code42 security-data search`
+    - `code42 security-data search --saved-search`
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,11 +32,10 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 - The `path` positional argument for bulk `generate-template` commands is now an option (`--p/-p`).
 - Below `search` subcommands accept argument `--format/-f` to display result in formats `csv`, `table`, `json`, `raw-json`:
-    - BREAKING CHANGE: Default output format is changed to `table` format from `raw-json`, returns a paginated response.
-    A predefined set of fields would be displayed by default, pass `--include-all` to view all non-nested top-level data.
+    - Default output format is changed to `table` format from `raw-json`, returns a paginated response.
+    A predefined properties would be displayed by default, pass `--include-all` to view all non-nested top-level properties.
     - `code42 alerts search`
     - `code42 security-data search`
-    - `code42 security-data search --saved-search`
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,7 +32,7 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 - The `path` positional argument for bulk `generate-template` commands is now an option (`--p/-p`).
 - Below `search` subcommands accept argument `--format/-f` to display result in formats `csv`, `table`, `json`, `raw-json`:
-    - BREAKING CHANGE: Default output format is changed to `table` format from `raw-json`, returns a paginated response
+    - BREAKING CHANGE: Default output format is changed to `table` format from `raw-json`, returns a paginated response.
     A predefined set of fields would be displayed by default, pass `--include-all` to view all non-nested top-level data.
     - `code42 alerts search`
     - `code42 security-data search`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,8 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
     each time the CLI is run with a profile configured this way, as it is not recommended.
 
 - The `path` positional argument for bulk `generate-template` commands is now an option (`--p/-p`).
-- Below `search` subcommands accept argument `--format/-f` to display result in formats `csv`, `table`, `json`, `formatted-json`:
+- Below `search` subcommands accept argument `--format/-f` to display result in formats `csv`, `table`, `json`, `raw-json`:
+    - BREAKING CHANGE: Default output format is changed to `table` format from `raw-json` 
     By default only root/top level items from the response will be displayed, to view all items, that contains nested data, pass `--include-all` option.
     - `code42 alerts search`
     - `code42 security-data search`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,7 +32,7 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 - The `path` positional argument for bulk `generate-template` commands is now an option (`--p/-p`).
 - Below `search` subcommands accept argument `--format/-f` to display result in formats `csv`, `table`, `json`, `raw-json`:
-    - BREAKING CHANGE: Default output format is changed to `table` format from `raw-json` 
+    - BREAKING CHANGE: Default output format is changed to `table` format from `raw-json`, returns a paginated response
     By default only root/top level items from the response will be displayed, to view all items, that contains nested data, pass `--include-all` option.
     - `code42 alerts search`
     - `code42 security-data search`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,7 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 - The `path` positional argument for bulk `generate-template` commands is now an option (`--p/-p`).
 - Below `search` subcommands accept argument `--format/-f` to display result in formats `csv`, `table`, `json`, `raw-json`:
     - BREAKING CHANGE: Default output format is changed to `table` format from `raw-json`, returns a paginated response
-    By default only root/top level items from the response will be displayed, to view all items, that contains nested data, pass `--include-all` option.
+    A predefined set of fields would be displayed by default, pass `--include-all` to view all non-nested top-level data.
     - `code42 alerts search`
     - `code42 security-data search`
     - `code42 security-data search --saved-search`
--- a/src/code42cli/cmds/alerts.py
+++ b/src/code42cli/cmds/alerts.py
@@ -11,7 +11,7 @@ import code42cli.cmds.search.options as searchopt
 import code42cli.errors as errors
 import code42cli.options as opt
 from code42cli.cmds.search.cursor_store import AlertCursorStore
-from code42cli.output_formats import format_option
+from code42cli.output_formats import extraction_format_option as format_option
 
 search_options = searchopt.create_search_options("alerts")
 
@@ -170,12 +170,7 @@ def clear_checkpoint(state, checkpoint_name):
     "--or-query", is_flag=True, cls=searchopt.AdvancedQueryAndSavedSearchIncompatible
 )
 @opt.sdk_options()
-@click.option(
-    "--display",
-    multiple=True,
-    default=_OPTIONAL_OPTIONS,
-    type=click.Choice(_OPTIONAL_OPTIONS),
-)
+@click.option("--include-all", default=False, is_flag=True)
 def search(
     cli_state,
     format,
@@ -184,7 +179,7 @@ def search(
     advanced_query,
     use_checkpoint,
     or_query,
-    display,
+    include_all,
     **kwargs
 ):
     """Search for alerts."""
@@ -195,8 +190,7 @@ def search(
         format,
         cursor,
         use_checkpoint,
-        format_header=_HEADERS_KEY_MAP,
-        optional_fields=_optionally_display(display),
+        display_all=include_all,
     )
     extractor = _get_alert_extractor(cli_state.sdk, handlers)
     extractor.use_or_query = or_query
--- a/src/code42cli/cmds/alerts.py
+++ b/src/code42cli/cmds/alerts.py
@@ -168,7 +168,7 @@ def clear_checkpoint(state, checkpoint_name):
     "--include-all",
     default=False,
     is_flag=True,
-    help="Display complete list of data items from the first level of the nested response.",
+    help="Display simple properties of the primary level of the nested response.",
 )
 def search(
     cli_state,
--- a/src/code42cli/cmds/alerts.py
+++ b/src/code42cli/cmds/alerts.py
@@ -158,7 +158,7 @@ def clear_checkpoint(state, checkpoint_name):
     "--include-all",
     default=False,
     is_flag=True,
-    help="Display complete list of data items or only from top level of nested response.",
+    help="Display complete list of data items from the first level of the nested response.",
 )
 def search(
     cli_state,
--- a/src/code42cli/cmds/alerts.py
+++ b/src/code42cli/cmds/alerts.py
@@ -22,7 +22,14 @@ _HEADERS_KEY_MAP["createdAt"] = "Observed Date"
 _HEADERS_KEY_MAP["state"] = "Status"
 _HEADERS_KEY_MAP["severity"] = "Severity"
 # _HEADERS_KEY_MAP["description"] = "Description"
-
+_OPTIONAL_OPTIONS = [
+    "sources",
+    "exposure_types",
+    "file_count",
+    "file_detail",
+    "file_size",
+    "ip",
+]
 
 severity_option = click.option(
     "--severity",
@@ -166,13 +173,19 @@ def clear_checkpoint(state, checkpoint_name):
 @click.option(
     "--display",
     multiple=True,
-    default=["sources", "exposure_types", "file_count", "file_detail", "file_size", "ip"],
-    type=click.Choice(
-        ["sources", "exposure_types", "file_count", "file_detail", "file_size", "ip"]
-    ),
+    default=_OPTIONAL_OPTIONS,
+    type=click.Choice(_OPTIONAL_OPTIONS),
 )
 def search(
-    cli_state, format, begin, end, advanced_query, use_checkpoint, or_query, display, **kwargs
+    cli_state,
+    format,
+    begin,
+    end,
+    advanced_query,
+    use_checkpoint,
+    or_query,
+    display,
+    **kwargs
 ):
     """Search for alerts."""
     cursor = _get_alert_cursor_store(cli_state.profile.name) if use_checkpoint else None
@@ -183,7 +196,7 @@ def search(
         cursor,
         use_checkpoint,
         format_header=_HEADERS_KEY_MAP,
-        optional_fields = _optionally_display(display)
+        optional_fields=_optionally_display(display),
     )
     extractor = _get_alert_extractor(cli_state.sdk, handlers)
     extractor.use_or_query = or_query
@@ -240,10 +253,8 @@ def _display_file_detail(event):
     data = event["observations"][0]["data"]
     event["file_detail"] = ""
     for file_detail in data["files"]:
-        event["file_detail"] += "{0}##{1}##{2}".format(
-            file_detail["name"],
-            file_detail["path"],
-            file_detail["category"]
+        event["file_detail"] += "{}##{}##{}".format(
+            file_detail["name"], file_detail["path"], file_detail["category"]
         )
     return event
 
@@ -261,7 +272,7 @@ _OPTIONAL_DISPLAY_FUNCTIONS = {
     "file_categories": _display_file_categories,
     "file_detail": _display_file_detail,
     "file_size": _display_file_size,
-    "ip": _display_ip
+    "ip": _display_ip,
 }
 
 
--- a/src/code42cli/cmds/alerts.py
+++ b/src/code42cli/cmds/alerts.py
@@ -154,7 +154,12 @@ def clear_checkpoint(state, checkpoint_name):
     "--or-query", is_flag=True, cls=searchopt.AdvancedQueryAndSavedSearchIncompatible
 )
 @opt.sdk_options()
-@click.option("--include-all", default=False, is_flag=True)
+@click.option(
+    "--include-all",
+    default=False,
+    is_flag=True,
+    help="Display complete list of data items or only from top level of nested response.",
+)
 def search(
     cli_state,
     format,
--- a/src/code42cli/cmds/alerts.py
+++ b/src/code42cli/cmds/alerts.py
@@ -21,7 +21,7 @@ _HEADERS_KEY_MAP["actor"] = "Username"
 _HEADERS_KEY_MAP["createdAt"] = "Observed Date"
 _HEADERS_KEY_MAP["state"] = "Status"
 _HEADERS_KEY_MAP["severity"] = "Severity"
-_HEADERS_KEY_MAP["description"] = "Description"
+# _HEADERS_KEY_MAP["description"] = "Description"
 
 
 severity_option = click.option(
@@ -163,8 +163,16 @@ def clear_checkpoint(state, checkpoint_name):
     "--or-query", is_flag=True, cls=searchopt.AdvancedQueryAndSavedSearchIncompatible
 )
 @opt.sdk_options()
+@click.option(
+    "--display",
+    multiple=True,
+    default=["sources", "exposure_types", "file_count", "file_detail", "file_size", "ip"],
+    type=click.Choice(
+        ["sources", "exposure_types", "file_count", "file_detail", "file_size", "ip"]
+    ),
+)
 def search(
-    cli_state, format, begin, end, advanced_query, use_checkpoint, or_query, **kwargs
+    cli_state, format, begin, end, advanced_query, use_checkpoint, or_query, display, **kwargs
 ):
     """Search for alerts."""
     cursor = _get_alert_cursor_store(cli_state.profile.name) if use_checkpoint else None
@@ -175,6 +183,7 @@ def search(
         cursor,
         use_checkpoint,
         format_header=_HEADERS_KEY_MAP,
+        optional_fields = _optionally_display(display)
     )
     extractor = _get_alert_extractor(cli_state.sdk, handlers)
     extractor.use_or_query = or_query
@@ -196,3 +205,67 @@ def _get_alert_extractor(sdk, handlers):
 
 def _get_alert_cursor_store(profile_name):
     return AlertCursorStore(profile_name)
+
+
+def _display_sources(event):
+    # Assuming here and all other `_include` function that only 'observations' contain a
+    # single record.
+    sources = event["observations"][0]["data"]["sources"]
+    event["sources"] = "##".join(sources)
+    return event
+
+
+def _display_exposure_types(event):
+
+    exposure_types = event["observations"][0]["data"]["exposureTypes"]
+    event["exposure_types"] = "##".join(exposure_types)
+    return event
+
+
+def _display_file_count(event):
+    event["file_count"] = event["observations"][0]["data"]["fileCount"]
+    return event
+
+
+def _display_file_categories(event):
+    pass
+
+
+def _display_file_size(event):
+    event["file_size"] = event["observations"][0]["data"]["totalFileSize"]
+    return event
+
+
+def _display_file_detail(event):
+    data = event["observations"][0]["data"]
+    event["file_detail"] = ""
+    for file_detail in data["files"]:
+        event["file_detail"] += "{0}##{1}##{2}".format(
+            file_detail["name"],
+            file_detail["path"],
+            file_detail["category"]
+        )
+    return event
+
+
+def _display_ip(event):
+    ips = event["observations"][0]["data"]["sendingIpAddresses"]
+    event["ip"] = "##".join(ips)
+    return event
+
+
+_OPTIONAL_DISPLAY_FUNCTIONS = {
+    "sources": _display_sources,
+    "exposure_types": _display_exposure_types,
+    "file_count": _display_file_count,
+    "file_categories": _display_file_categories,
+    "file_detail": _display_file_detail,
+    "file_size": _display_file_size,
+    "ip": _display_ip
+}
+
+
+def _optionally_display(display_options):
+    for option in display_options:
+        _HEADERS_KEY_MAP[option] = option.capitalize()
+    return [_OPTIONAL_DISPLAY_FUNCTIONS[option] for option in display_options]
--- a/src/code42cli/cmds/alerts.py
+++ b/src/code42cli/cmds/alerts.py
@@ -1,3 +1,5 @@
+from _collections import OrderedDict
+
 import click
 import py42.sdk.queries.alerts.filters as f
 from c42eventextractor.extractors import AlertExtractor
@@ -11,6 +13,14 @@ import code42cli.options as opt
 from code42cli.cmds.search.cursor_store import AlertCursorStore
 from code42cli.output_formats import extraction_format_option as format_option
 
+
+SEARCH_DEFAULT_HEADER = OrderedDict()
+SEARCH_DEFAULT_HEADER["name"] = "RuleName"
+SEARCH_DEFAULT_HEADER["actor"] = "Username"
+SEARCH_DEFAULT_HEADER["createdAt"] = "ObservedDate"
+SEARCH_DEFAULT_HEADER["state"] = "Status"
+SEARCH_DEFAULT_HEADER["severity"] = "Severity"
+SEARCH_DEFAULT_HEADER["description"] = "Description"
 
 search_options = searchopt.create_search_options("alerts")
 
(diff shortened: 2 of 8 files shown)


src/code42cli/cmds/search/extraction.py
```diff
@@ -61,12 +69,22 @@ def handle_response(response):
                 events = _get_alert_details(sdk, events)
             except Exception as ex:
                 handlers.handle_error(ex)
+
         handlers.TOTAL_EVENTS += len(events)
-        event = None
-        for event in events:
-            output_logger.info(event)
-        if event:
-            last_event_timestamp = extractor._get_timestamp_from_item(event)
```
kiran-chaudhary, line 69: I would like to restructure code here, instead of processing events here in the handler, they should be processed in their respective "feature" module. 
This will allow us to implement separate processing of `security-data` events and `alert` events if needed. Also with that lines 67 to 71 can move in `alerts` module.  

Implying, let this module be generic and maintain separation of concern, without being aware what type of data it is processing.

This can be achieved by: `handlers.EVENTS = events` without needing a return statement.

src/code42cli/cmds/search/extraction.py
```diff
@@ -61,12 +69,22 @@ def handle_response(response):
...
+                else {
+                    key: key.capitalize()
+                    for key in events[0].keys()
+                    if type(events[0][key]) == str
+                }
+            )
+
+            click.echo_via_pager(output_format(events, format_header))
+
+            last_event_timestamp = extractor._get_timestamp_from_item(events[-1])
```
kiran-chaudhary, line 87: is this understanding fine? last timestamp is based on the last observed event. Looking at the existing code it seems so, Irrespective of whether it was sorted by some other field, etc, etc.

src/code42cli/output_formats.py
```diff
@@ -30,15 +46,33 @@ def output_format(_, __, value):
...
+extraction_format_option = click.option(
+    "-f",
+    "--format",
+    type=click.Choice(OutputFormat(), case_sensitive=False),
+    help="The output format of the result. Defaults to table format.",
+    callback=extraction_output_format,
+)
+
+
+def to_dynamic_csv(output, header):
```
kiran-chaudhary, line 58: Created a separate function that creates `csv` output with 'csv' module. As I did not want to change existing function where we had flexibility to set custom column headers. I couldn't figure out how that would be achieved with `csv` module which takes `fieldnames` as column header by default. Also note that header is based on keys found in first record of `events`.

src/code42cli/cmds/search/extraction.py
```diff
@@ -61,12 +69,22 @@ def handle_response(response):
...
-        for event in events:
-            output_logger.info(event)
-        if event:
-            last_event_timestamp = extractor._get_timestamp_from_item(event)
+        if len(events):
+            format_header = (
+                {key: key.capitalize() for key in events[0].keys()}
+                if display_all
+                else {
+                    key: key.capitalize()
```
unparalleled-js, line 79: This whole part is a bit confusing to follow.

unparalleled-js: I don't really get the different between `display_all` and not `display_all` as far as this logic goes... maybe it'd be best to make some old-fashioned `if, else` statements here.

src/code42cli/cmds/search/extraction.py
unparalleled-js: I think this makes sense, yes, but I don't follow what you mean by `sorted by some other field`. I think this has to be sorted by a timestamp to work properly.

src/code42cli/cmds/search/extraction.py
kiran-chaudhary: I understand the pain in reading it, as this is the only way ternary operator is defined in python. 

Based on the `display_all` it initialises `format_header` variable. 
Otherwise I have to do either of below and former involves extra process and later involves extra declaration!! 
```
format_header =  {key: key.capitalize() for key in events[0].keys()}
if display_all:
    format_header = {key: key.capitalize() for key in events[0].keys() if...}
```

or

```
format_header = []
if display_all:
    format_header =  {key: key.capitalize() for key in events[0].keys()}
else:
   format_header =  {key: key.capitalize() for key in events[0].keys() if.....}
```

src/code42cli/output_formats.py
unparalleled-js: Custom headers look nice but I am wondering if it is just simply better to use response keys themselves.

They can vary though - `legalHoldUid` is `Matter ID`.

`Matter ID` looks nice but now we are maintaining a mapping _and_ obfuscating results.

src/code42cli/cmds/search/extraction.py
kiran-chaudhary: Yea, that is what I wanted to confirm, extractor always returns result sorted by timestamp, that it doesn't support sorting by other fields.

CHANGELOG.md
```diff
@@ -29,6 +29,12 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 - A profile created with the `--disable-ssl-errors` flag will now correctly not verify SSL certs when making requests. A warning message is printed
     each time the CLI is run with a profile configured this way, as it is not recommended.
 
+- Below `search` subcommands accept argument `--format/-f` to display result in formats `csv`, `table`, `json`, `formatted-json`:
```
unparalleled-js, line 32: Didn't the default output change for a lot of commands as well? We should also mention that... What before returned a JSON blob, now returns a terminal table without using the proper flag.

src/code42cli/cmds/alerts.py
```diff
@@ -159,14 +154,27 @@ def clear_checkpoint(state, checkpoint_name):
     "--or-query", is_flag=True, cls=searchopt.AdvancedQueryAndSavedSearchIncompatible
 )
 @opt.sdk_options()
+@click.option(
+    "--include-all",
+    default=False,
+    is_flag=True,
+    help="Display complete list of data items or only from top level of nested response.",
```
unparalleled-js, line 161: Should the word `or` be `of`?

src/code42cli/cmds/alerts.py
```diff
@@ -180,6 +188,9 @@ def search(
         extractor.extract(*cli_state.search_filters)
     if handlers.TOTAL_EVENTS == 0 and not errors.ERRORED:
         echo("No results found.")
+    else:
+        output = _process_events(format, include_all, handlers.EVENTS)
```
unparalleled-js, line 192: Shouldn't this happen in the handlers?

src/code42cli/cmds/search/extraction.py
```diff
@@ -61,12 +63,11 @@ def handle_response(response):
                 events = _get_alert_details(sdk, events)
             except Exception as ex:
                 handlers.handle_error(ex)
-        handlers.TOTAL_EVENTS += len(events)
-        event = None
-        for event in events:
-            output_logger.info(event)
```
unparalleled-js, line 67: Did miss a discussion where we decided to no longer output the data using the handlers?

src/code42cli/cmds/securitydata.py
```diff
@@ -172,6 +168,12 @@ def clear_checkpoint(state, checkpoint_name):
     "--or-query", is_flag=True, cls=searchopt.AdvancedQueryAndSavedSearchIncompatible
 )
 @sdk_options()
+@click.option(
+    "--include-all",
+    default=False,
+    is_flag=True,
+    help="Display complete list of data items or only from top level of nested response.",
```
unparalleled-js, line 175: I don't think `or only` here makes sense either...  

How about just `Display complete list of data items from the first level of the nested response`

CHANGELOG.md
kiran-chaudhary: Yea this should change. Wondering, do we want to go ahead with the breaking change or shall we default it to JSON blob, as earlier? 
Is it possible to get a confirmation on whether the default format can be changed? i.e there isn't any automation around this? or just declaring that this is a breaking change would suffice, any one who updates to latest version will do the needful.

src/code42cli/cmds/alerts.py
kiran-chaudhary: Need more improvisation here, specifying "--include-all" does the other way around, it display all items and defaults to top level. 

How about `Display complete list of data items, defaults to data items from the first level of the nested response`

src/code42cli/cmds/alerts.py
kiran-chaudhary: I did initiate a [discussion](https://github.com/code42/code42cli/pull/133#discussion_r466994317) ,  I wanted processing of events outside the extraction module because it gives us flexibility to process data in their respective module. 

`security-data` doesn't need much processing as all the items are at top level data, whereas `alerts` can be processed as below in `alerts` module. 

```
def _transform_nested_events(events):
    transformed_events = []
    for event in events:
        transformed_event = {key: event[key] for key in event.keys() if type(event[key]) == str}
        
        for observation in event["observations"]:
            for key in observation["data"]:
                transformed_event["observation_"+key] = observation["data"][key]
        transformed_events.append(transformed_event)
    return transformed_events


def _process_events(output_format, include_all, events):
    if not include_all:
        format_header = get_format_header(include_all, events[0])
        return output_format(events, format_header)
    
    transformed_events = _transform_nested_events(events)
    format_header = {key: key.capitalize() for key in transformed_events[0].keys()}
    return output_format(transformed_events, format_header)

```

src/code42cli/cmds/search/extraction.py
kiran-chaudhary: Same as above.

src/code42cli/cmds/securitydata.py
kiran-chaudhary: same as [above comment](https://github.com/code42/code42cli/pull/133#discussion_r468312462)

src/code42cli/cmds/alerts.py
kiran-chaudhary: I.e in `table` format and `csv` format we can display data by removing the nested structure.

kiran-chaudhary: I was referring to above code in the meeting, replace `_process_events` method with above code and observe the difference in output with `--include-all` option. 

Also this is just an idea, we can polish it further if team has better ideas. 

All nested data keys are prepended with "observation_", that's because some keys from top level data  like `type` etc do not get overwritten.

CHANGELOG.md
alanag13: The default output format should change to table. Calling this out as a breaking change should suffice in this case.

src/code42cli/cmds/alerts.py
alanag13: @unparalleled-js 's verbiage suggestion works here, there is not a way to get any items from past the first level of the response except for with json/csv.

src/code42cli/cmds/securitydata.py
alanag13: I would prefer @unparalleled-js 's proposed verbiage here.

src/code42cli/cmds/alerts.py
unparalleled-js: Perhaps, could you make them have / implement separate handlers instead?  

I am not a fan of holding onto `EVENTS`. They should be processed as they come through. Storing them in a class property to process later seems less functional and more of a work-around.

This is not a complete blocker for me since this all behind the scenes stuff ... but if we are going to have event handlers, they should handle the event

alanag13: I agree that `security-data` and `alerts` should have separate methods for processing data, but I think that should be handled by using separate handler instances (each with their own process logic defined) instead of moving the logic to be completely outside the handlers.

alanag13: also, since `observations` is a nested object and _especially_ because it could contain more than one item, I thought we had agreed that it should be excluded from being output in `alerts`, am I misunderstanding something?

timabrmsn: I also agree that the handlers should be defined separately, and we definitely want to process events as they come in, otherwise for large queries that might make multiple 10k search requests, the user would have to wait until they _all_ were retrieved before anything got output to the screen, which isn't a great experience.

src/code42cli/cmds/search/extraction.py
```diff
@@ -61,12 +63,11 @@ def handle_response(response):
...
-        if event:
-            last_event_timestamp = extractor._get_timestamp_from_item(event)
+        total_events = len(events)
+        handlers.TOTAL_EVENTS += total_events
+        handlers.EVENTS = events
```
timabrmsn, line 68: This will probably change given the other discussions about separating the handlers, but right now this is eating all but the last page of events, since you're overwriting `handlers.EVENTS` instead of appending them to a list.

src/code42cli/cmds/securitydata.py
```diff
@@ -206,6 +209,9 @@ def search(
         extractor.extract(*state.search_filters)
     if handlers.TOTAL_EVENTS == 0 and not errors.ERRORED:
         echo("No results found.")
+    else:
+        output = _process_events(format, include_all, handlers.EVENTS)
+        click.echo_via_pager(output)
```
alanag13, line 214: how does `echo_via_pager` behave when the command is run in the background?

timabrmsn: I'm pretty sure it detects if you're in a TTY and if so, attempts to pipe to your default pager, otherwise does the same as `echo()` does. So if you're piping the output to another process or file, it skips the pager.

src/code42cli/cmds/alerts.py
kiran-chaudhary: I agree with processing events as they come in and will get those changes done. 

I might have misread if [this](https://github.com/code42/code42cli/pull/133/files#r468700924) was finalised. So to clarify on it, current implementation is, we display all top level data by default. Now, if we do not display the nested object (`observations`) then do we need `--include-all` option!?

src/code42cli/cmds/securitydata.py
kiran-chaudhary: changed.

CHANGELOG.md
kiran-chaudhary: Done.

kiran-chaudhary:
Based on the above raised concerns and challenges I faced to get the desired output, I would like to discuss below points to get my understanding right:

- Processing single record at a time with `echo_via_pager` doesn't provide the intended behaviour we expect. It displays only one record at a time, so I reverted code back to `echo`.

- Processing one record  at a time, requires a workaround to print headers. As of now, I have skipped printing headers, which I feel is needed in `table` and `csv` format.

In `extraction` module statement `len(events)` indicate that events is not a `generator` but a `list` and I confirmed the same by checking `type(events)`, i.e all events are loaded in memory instead of 1 at a time.

src/code42cli/cmds/alerts.py
alanag13: The intention was not to output _all_ top-level fields by default, but rather to choose a _default_ set of top-level fields to output. Using `--include-all` would output all _top-level_ fields, but nested fields, lists, etc would never be output to the table.

alanag13:
> Based on the above raised concerns and challenges I faced to get the desired output, I would like to discuss below points to get my understanding right:
> 
> * Processing single record at a time with `echo_via_pager` doesn't provide the intended behaviour we expect. It displays only one record at a time, so I reverted code back to `echo`.
> * Processing one record  at a time, requires a workaround to print headers. As of now, I have skipped printing headers, which I feel is needed in `table` and `csv` format.
> * In `extraction` module statement `len(events)` indicate that events is not a `generator` but a `list` and I confirmed the same by checking `type(events)`, i.e all events are loaded in memory instead of 1 at a time.
> 
> Preferably discussing over a call would be better.

I think there's some confusion here, the intention is to process one _page_ of events at a time, not one _event_ at a time.
- `echo_via_pager`'s first argument is meant to be a generator. We will need to design a generator that returns a full page of events each time it is called, and use that. Per our discussion, the best way to handle this is probably to add a `extract_pages()` method to the event extractor that accepts a number of events to return per page and yields one page a time (it should otherwise behave exactly the same as `extract`).

src/code42cli/cmds/search/extraction.py
```diff
@@ -1,17 +1,20 @@
...
 import code42cli.errors as errors
 from code42cli.date_helper import verify_timestamp_order
 from code42cli.logger import get_main_cli_logger
+from code42cli.output_formats import get_dynamic_header
 from code42cli.util import warn_interrupt
 
 logger = get_main_cli_logger()
 
 _ALERT_DETAIL_BATCH_SIZE = 100
+_EVENT_COUNT_PER_PAGE = 10
```
kiran-chaudhary, line 17: Need to decide a proper value or it can be a command line option, allowing user to define how many results to view.

src/code42cli/cmds/search/extraction.py
kiran-chaudhary: As we are using `echo_via_pager`, this is no longer required. 

Although this will be needed for `send-to` but we will define a new function for it.

src/code42cli/cmds/securitydata.py
```diff
@@ -25,15 +26,21 @@
 _HEADER_KEYS_MAP["name"] = "Name"
 _HEADER_KEYS_MAP["id"] = "Id"
 
+SEARCH_DEFAULT_HEADER = OrderedDict()
```
kiran-chaudhary, line 29: Some of the fields can be removed since having so many fields doesn't fit in single terminal row. 

Added these fields based on the idea that,  users can specify these fields in search criteria ( advanced_query )

kiran-chaudhary:
> > Based on the above raised concerns and challenges I faced to get the desired output, I would like to discuss below points to get my understanding right:
> > 
> > * Processing single record at a time with `echo_via_pager` doesn't provide the intended behaviour we expect. It displays only one record at a time, so I reverted code back to `echo`.
> > * Processing one record  at a time, requires a workaround to print headers. As of now, I have skipped printing headers, which I feel is needed in `table` and `csv` format.
> > * In `extraction` module statement `len(events)` indicate that events is not a `generator` but a `list` and I confirmed the same by checking `type(events)`, i.e all events are loaded in memory instead of 1 at a time.
> > 
> > Preferably discussing over a call would be better.
> 
> I think there's some confusion here, the intention is to process one _page_ of events at a time, not one _event_ at a time.
> 
> `echo_via_pager`'s first argument is meant to be a generator. We will need to design a generator that returns a full page of events each time it is called, and use that. Per our discussion, the best way to handle this is probably to add a `extract_pages()` method to the event extractor that accepts a number of events to return per page and yields one page a time (it should otherwise behave exactly the same as `extract`).

@alanag13 While implementing `extract_pages` realised that same logic can be defined here as well, and so created a `_pagination` method to achieve the same.

src/code42cli/cmds/securitydata.py
kiran-chaudhary: Confirmed it skips the pager.

src/code42cli/cmds/search/extraction.py
kiran-chaudhary: Changed.

src/code42cli/cmds/alerts.py
kiran-chaudhary: Changed as discussed.

CHANGELOG.md
```diff
@@ -31,6 +31,12 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
     each time the CLI is run with a profile configured this way, as it is not recommended.
 
 - The `path` positional argument for bulk `generate-template` commands is now an option (`--p/-p`).
+- Below `search` subcommands accept argument `--format/-f` to display result in formats `csv`, `table`, `json`, `raw-json`:
+    - BREAKING CHANGE: Default output format is changed to `table` format from `raw-json`, returns a paginated response.
```
unparalleled-js, line 35: If it is under `changed` header in the Changelog, you don't need to declare `BREAKING CHANGE` here

CHANGELOG.md
```diff
@@ -31,6 +31,12 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
...
+    A predefined set of fields would be displayed by default, pass `--include-all` to view all non-nested top-level data.
+    - `code42 alerts search`
+    - `code42 security-data search`
+    - `code42 security-data search --saved-search`
```
unparalleled-js, line 39: `code42 security-data search` probably encompasses `code42 security-data search --saved-search`

src/code42cli/cmds/alerts.py
```diff
@@ -159,14 +164,33 @@ def clear_checkpoint(state, checkpoint_name):
     "--or-query", is_flag=True, cls=searchopt.AdvancedQueryAndSavedSearchIncompatible
 )
 @opt.sdk_options()
+@click.option(
+    "--include-all",
+    default=False,
+    is_flag=True,
+    help="Display complete list of data items from the first level of the nested response.",
```
unparalleled-js, line 171: Is `results` a better word for `data items` ?

unparalleled-js, line 171: Is `properties` a better word for `data items` ?

src/code42cli/cmds/search/extraction.py
```diff
@@ -61,12 +79,18 @@ def handle_response(response):
...
+        handlers.TOTAL_EVENTS += total_events
+
+        for events_per_page in _paginate(events, _EVENT_COUNT_PER_PAGE):
+            output = _process_events(
+                output_format, include_all, events_per_page, output_header
+            )
+            click.echo_via_pager(output)
+
+        if total_events:
+            last_event_timestamp = extractor._get_timestamp_from_item(events[-1])
```
unparalleled-js, line 93: This could be dangerous. We may change the extractor internal method not realizing or remembering it is being used here.

src/code42cli/cmds/securitydata.py
```diff
@@ -171,6 +178,12 @@ def clear_checkpoint(state, checkpoint_name):
     "--or-query", is_flag=True, cls=searchopt.AdvancedQueryAndSavedSearchIncompatible
 )
 @sdk_options()
+@click.option(
+    "--include-all",
+    default=False,
+    is_flag=True,
+    help="Display complete list of data items from the first level of the nested response.",
```
unparalleled-js, line 185: Similarly here, I think `properties` is the right word for `data items`

unparalleled-js: Another thing here, it is immediately clear to me what is at the first level.  Sometimes responses just have `data` at the first level.... Does this include the name of the nest objects with a value of `{ ... }` (like Chrome dev tools), or does it exclude it entirely.


Maybe it should say `Display simple properties of the primary level of the nested response`

src/code42cli/cmds/alerts.py
unparalleled-js: Actually, see this comment: https://github.com/code42/code42cli/pull/133/files#r470628459

src/code42cli/util.py
```diff
@@ -47,11 +47,14 @@ def find_format_width(record, header):
         record (list of dict), data to be formatted.
         header (dict), key-value where keys should map to keys of record dict and
           value is the corresponding column name to be displayed on the cli.
+        include_header (bool), include header in output, defaults to True.
```
unparalleled-js, line 50: Should be `(bool, optional)` here

src/code42cli/cmds/search/extraction.py
alanag13: I agree, but this was already in place. My hope is that once we have proper support for getting more than 10K events via the api that we can just kill the extractor altogether.

src/code42cli/cmds/securitydata.py
alanag13: The eventId is not particularly useful information and I would exclude it by default.
What order are the fields presented in? I think identifying information such as the file name and path should come first.

src/code42cli/cmds/search/extraction.py
unparalleled-js: This already happens in the extractor.

unparalleled-js: Any reason for that? @alanag13 @timabrmsn

kiran-chaudhary: We just found this part is duplicated, those two lines are already being executed in "extractor" and things work just fine event without it. So we have removed it.

src/code42cli/cmds/securitydata.py
kiran-chaudhary: The order is same as they are added in the dict.

kiran-chaudhary: "FilePath"
 "FileName"
 "Type"
 "EventTimestamp"
 "FileCategory"
 "FileSize"
 "FileOwner"
 "MD5Checksum"
"SHA256Checksum"

kiran-chaudhary: Just removed EventId and brought FilePath and FileName above.

src/code42cli/cmds/search/extraction.py
```diff
@@ -83,11 +75,10 @@ def handle_response(response):
         total_events = len(events)
         handlers.TOTAL_EVENTS += total_events
 
-        for events_per_page in _paginate(events, _EVENT_COUNT_PER_PAGE):
-            output = _process_events(
-                output_format, include_all, events_per_page, output_header
-            )
-            click.echo_via_pager(output)
+        def paginate():
```
unparalleled-js, line 78: I like this!

Note to others:  Kiran and I were on a call and found out that by passing the generator in, `echo_via_page()` will handle pagination for us. Meaning, it display records to fit in the current terminal size and then hitting enter allows you to page through them.

src/code42cli/cmds/search/extraction.py
```diff
@@ -96,3 +107,9 @@ def create_time_range_filter(filter_cls, begin_date=None, end_date=None):
 
     elif end_date and not begin_date:
         return filter_cls.on_or_before(end_date)
+
+
+def _process_events(output_format, include_all, events, output_header):
+    if include_all:
```
unparalleled-js, line 113: Now... I don't think we need to print the header again fir the next page.

The experience should be that we are paging through the records in the terminal and shouldn't be able to distinguish when the next set of events (the HTTP page) comes through... The header should happen only once I think.

That would required another condition checked here, like `not is_first_page`... I am pretty sure the HTTP response will tell us what page it is. If not, there are other ways.

unparalleled-js: Ah, nevermind about it being in the response. I forgot that this is not actually an API paging mechanism

unparalleled-js: Hmm... You can the extractor's `_self._previous_event_count` though

src/code42cli/cmds/search/extraction.py
unparalleled-js: Welll.... this isn't quite enough.... So, the END paging happens once the generator runs outs..... UGH!


This is a for sure finding:   The way we have the extraction handlers set up does not plug in well to `echo_via_pager()`

unparalleled-js:
I am really not sure what to do here. `echo_via_pager()` expect all of the data, either in a generator or as text.  The way we have to do it is to give 1 extractor page's worth of data, either through a generator or text (preferably generator for reasons stated above). 

We are unable to do this. Meaning, we are unable to provide a decent paging experience. So we have to make a choice and I can't quite figure out the best course yet.

If the extractor yields the pages instead of using the handlers, this would be much easier.

src/code42cli/cmds/search/extraction.py
timabrmsn: It was duplicated here because when you hit ctrl-c to interrupt the process, it will finish processing the _current_ batch (the existing `handle_response()` call), then exit. But the extractor only saves the checkpoint _after_ `handle_response()` is called. So if you ctrl-c you'd get a bunch of processed events but the checkpoint would be off and the next run you'd get duplicates.

src/code42cli/cmds/search/extraction.py
unparalleled-js: None of what I said here is relevant anymore.... https://github.com/code42/code42cli/pull/133#issuecomment-674169518

src/code42cli/cmds/search/extraction.py
unparalleled-js: OK maybe it is not exiting early, it just does the END thing multiple times, which is not ideal, but in that case it is fine to have the headers multiple times.

unparalleled-js:
This probably is not a big deal since there is going to be 10,000 records before the next extractor page.

src/code42cli/cmds/search/extraction.py
unparalleled-js, line 113: Note that this means every 10,000 records, we are going to print the headers again...

This is not a big deal because we are also going to reach the END line of terminal pagers.... which is weird, but probably the best way of doing this currently.

src/code42cli/cmds/search/extraction.py
alanag13: I was going to suggest 20, but I'm open to feedback on that. I think an option for specifying the number of lines to output is probably not necessary.

src/code42cli/cmds/securitydata.py
alanag13: what do you think about having fileName be first?

src/code42cli/cmds/search/extraction.py
alanag13: I don't see paging past 10K events in the terminal being a common occurrence so I wouldn't worry about this at the moment.

timabrmsn: Yeah, since `handle_response` is expected to do all the actual processing of the events calling `echo_via_pager()` multiple times will give us a pager process for each page, instead of it appending it all into a single pager process. 

I think our only option would be to have `extract()` be called on a separate thread, and `handle_response()` append the results to a queue. Then we could on the main thread have a generator function that pulled events off the queue as they came in to pass to `echo_via_pager`. I think that would get us updates in the pager in "real time" while being able to pass everything in a single generator.

unparalleled-js:
Could we get some tests in `test_securitydata.py` that make sure `echo_via_pager()` gets called per extractor page?

Also for alerts

Otherwise this looks good and I appreciate you spending much time on this! This was a tricky problem.

timabrmsn:
If the way the extractor is currently designed is too big of a hassle to use `echo_via_pager`, I'd rather just skip it and `echo` directly to stdout like we were doing, since a user can always just pipe the output into their own pager of choice to get everything in the pager, and the multiple instances of the pager would be a weird unexpected experience.

unparalleled-js:
> If the way the extractor is currently designed is too big of a hassle to use `echo_via_pager`, I'd rather just skip it and `echo` directly to stdout like we were doing, since a user can always just pipe the output into their own pager of choice to get everything in the pager, and the multiple instances of the pager would be a weird unexpected experience.

The only issue at this point is that it will END after 10,000 events and then suddenly start paging again, so you don't lose any events, but it looks like it restarts or something..... I spoke with @alanag13 on Slack and we were thinking that paging through that many events in the terminal is an edge case.

That being said, many times today I felt exactly like you are now @timabrmsn in that we should just let them pipe to `less`.  Commit https://github.com/code42/code42cli/pull/133/commits/445b986cd76b6ee1c183beaebc0c5e1478988544 is what made me start to sway to how it is currently.

alanag13:
@unparalled-js @timabrmsn if the intent of table output is to make something quickly and easily readable in a terminal it seems off to me to have it output 10K records to the screen at once. I recognize the imperfection of the current behavior but since it only occurs after 10K records I think that in virtually all cases it will be a better user experience to leave this as-is. We can solidify it once the change for properly supporting more than 10K results for file events is merged -- I'm pretty sure theres a PR open for it right now!

src/code42cli/cmds/securitydata.py
kiran-chaudhary: Changed.

src/code42cli/cmds/search/extraction.py
kiran-chaudhary: Added back.

src/code42cli/cmds/search/extraction.py
unparalleled-js: It should be terminal height, which it does automatically!

src/code42cli/cmds/search/extraction.py
```diff
@@ -61,12 +71,18 @@ def handle_response(response):
...
-        if event:
-            last_event_timestamp = extractor._get_timestamp_from_item(event)
+
+        total_events = len(events)
+        handlers.TOTAL_EVENTS += total_events
+
+        def paginate():
+            yield _process_events(output_format, include_all, events, output_header)
+
+        click.echo_via_pager(paginate)
```
unparalleled-js, line 81:
``` if len(events) > 10:
            click.echo_via_pager(paginate)
        else:
            event_text = [e for e in paginate()]
            click.echo(event_text)``` ?

PR #134, opened 2020-08-10
Add date validation for bulk add command in departing employee

kiran-chaudhary:
**Change Description**

Added `departure_date` validation in `bulk add` subcommand in `departing_employee`  command.

**Test Guidelines**
`code42 departing-employee bulk add <csv-file>`

**PR Checklist**

- [ n/a] Add docstrings for any new public parameters / methods / classes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 ### Fixed
 
 - Bug where `code42 legal-hold show` would error when terminal was too small.
+- Fixed bug in `departing_employee bulk add` command that allowed invalid dates to be passed without validation.
 
 ### Changed
 
--- a/src/code42cli/cmds/departing_employee.py
+++ b/src/code42cli/cmds/departing_employee.py
@@ -1,3 +1,5 @@
+from datetime import datetime
+
 import click
 from py42.exceptions import Py42BadRequestError
 
@@ -15,6 +17,9 @@ from code42cli.options import OrderedGroup
 from code42cli.options import sdk_options
 
 
+DATE_FORMAT = "%Y-%m-%d"
+
+
 @click.group(cls=OrderedGroup)
 @sdk_options(hidden=True)
 def departing_employee(state):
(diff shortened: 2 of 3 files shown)


kiran-chaudhary:
Used solution to invoke 'add' command from `bulk add` command though its discouraged [here](https://click.palletsprojects.com/en/6.x/advanced/#invoking-other-commands). 

The validation takes place in the bulk command itself since validation check in `add` command is ignored. 
The behaviour remains same, in case of invalid date `ValueError` is being raised and printed to log file.

unparalleled-js:
This PR seems good, but noticed a few things when testing it.

1.  The error message (`ValueError: time data '04-04-2021' does not match format '%Y-%m-%d'` should give an example along with the format string... it would be more human readable

2. There seems to be an extra space where the progress bar used to be for bulk commands, causing the error message output to look a little odd.

timabrmsn:
click's type classes have a `convert()` method that is used on the arg values when the click app is invoked, so you could use that instead of manually calling `.invoke()` in the bulk command:

```
@sdk_options()
def bulk_add(state, csv_rows):
    sdk = state.sdk

    def handle_row(username, cloud_alias, departure_date, notes):
        departure_date = click.DateTime(formats=[DATE_FORMAT]).convert(departure_date)
        _add_departing_employee(sdk, username, cloud_alias, departure_date, notes)
```

kiran-chaudhary:
> click's type classes have a `convert()` method that is used on the arg values when the click app is invoked, so you could use that instead of manually calling `.invoke()` in the bulk command:
> 
> ```
> @sdk_options()
> def bulk_add(state, csv_rows):
>     sdk = state.sdk
> 
>     def handle_row(username, cloud_alias, departure_date, notes):
>         departure_date = click.DateTime(formats=[DATE_FORMAT]).convert(departure_date)
>         _add_departing_employee(sdk, username, cloud_alias, departure_date, notes)
> ```

Used the `convert` method for validation.
Since `_add_departing_employee` expects `str` date format instead of `datetime`, we need to convert back `datetime` to `str` as its done in `add` command, to avoid repetition didn't change that part. Otherwise that condition should be moved to `_add_departing_employee`

Also `bulk add`  is a collection of `add` so it makes sense to call `add` from it.

kiran-chaudhary:
> This PR seems good, but noticed a few things when testing it.
> 
> 1. The error message (`ValueError: time data '04-04-2021' does not match format '%Y-%m-%d'` should give an example along with the format string... it would be more human readable
> 2. There seems to be an extra space where the progress bar used to be for bulk commands, causing the error message output to look a little odd.

1. Handled exception and added proper message. 
2. The extra space is observed when the progress bar isn't printed. When all the dates in the csv are incorrect then this case is observed.

unparalleled-js:
Consider it still approved on my end, but the exception still looks the same to me.

PR #135, opened 2020-08-11
implement doc feedback (INTEG-1135)

timabrmsn:


src/code42cli/bulk.py
```diff
@@ -52,9 +52,9 @@ def generate_template_cmd_factory(group_name, commands_dict):
     )
     def generate_template(cmd, path):
         """\b
-        Generate the csv template needed for bulk adding/removing users.
+        Generate the CSV template needed for bulk adding/removing users.
 
-        Optional PATH argument can be provided to write to a specific file path/name.
+        Add the optional PATH argument to write to a specific file path/name.
```
unparalleled-js, line 57: Why is PATH capitalized? Makes me think it is the environment variable.

timabrmsn: It's because click capitalizes argument names in the help docs:

```
# code42 departing-employee bulk generate-template -h
Usage: code42 departing-employee bulk generate-template [OPTIONS] [add|remove] [PATH]

  Generate the CSV template needed for bulk adding/removing users.

  Add the optional PATH argument to write to a specific file path/name.
```

unparalleled-js: Since it's positional anyway, I wonder if `SAVE_PATH` is a better name. The whole last three words of this command is a bit ambiguous:  `generate-template add path/to/file`. <--  I can't easily tell what this is doing.

unparalleled-js: I am starting to think that positional arguments should only be used when it's super obvious.

timabrmsn: I converted path to an option, as it does make more sense as one.

PR #137, opened 2020-08-17
Feature/send to

kiran-chaudhary:
**Description of Change**

code42 alerts search send-to
code42 security-data  search send-to

**Test Guidelines:**

$ code42 security-data search send-to localhost -b 2020-07-10
$ code42 alerts search send-to localhost -b 2020-07-10

**PR Checklist**

- [ ] Add docstrings for any new public parameters / methods / classes


unparalleled-js:
Should `send-to` be a flag? It will share all the same args as `search` but tacks on protocol and an address.

I propose it to be this way:

```
code42 security-data search --send TCP:sysylog.code42.com:456
```

where the protocol is `TCP`, the address is `syslog.code42.com` and the port is `456`.

src/code42cli/cmds/securitydata.py
```diff
@@ -234,9 +225,79 @@ def show(state, search_id):
...
+        extractor.extract_advanced(advanced_query)
+    elif saved_search:
+        extractor.extract(*saved_search._filter_group_list)
+    else:
+        if begin or end:
+            state.search_filters.append(
+                ext.create_time_range_filter(f.EventTimestamp, begin, end)
+            )
+        extractor.extract(*state.search_filters)
+    if handlers.TOTAL_EVENTS == 0 and not errors.ERRORED:
```
unparalleled-js, line 302: could simply to `not handlers.TOTAL_EVENTS`

kiran-chaudhary: `if not handlers.TOTAL_EVENTS and not errors.ERRORED`? missing `not`

kiran-chaudhary:
@unparalleled-js  Do we want to `send` behaviour similar to search, i.e have all format options with `--include-all` option for `send` as well??

alanag13:
> @unparalleled-js Do we want `send` behaviour similar to search, i.e have all format options with `--include-all` option for `send` as well??

If send is its own command and still defaults to tableoutput, then yes. `--include-all` is only relevent to table output, but should be honored regardless of it being printed to terminal or send over the network.



> Should `send-to` be a flag? It will share all the same args as `search` but tacks on protocol and an address.
> 
> I propose it to be this way:
> 
> ```
> code42 security-data search --send TCP:sysylog.code42.com:456
> ```
> 
> where the protocol is `TCP`, the address is `syslog.code42.com` and the port is `456`.

I like this idea a lot, but I think specifying the protocol crowds it a bit. Additionally, we might be adding support for TLS-based syslogging, which would mean we would also need to add the argument for a path to a cert. I think we should leave it as a separate command for now.

tests/cmds/test_alerts.py
```diff
@@ -632,3 +632,27 @@ def test_search_with_or_query_flag_produces_expected_query(runner, cli_state):
     }
     actual_query = json.loads(str(cli_state.sdk.alerts.search.call_args[0][0]))
     assert actual_query == expected_query
+
+
+def test_send_to_makes_call_to_the_extract_method(
```
alanag13, line 637: some tests that validate that the `send_events` methods get called with the expected data would be helpful.

src/code42cli/cmds/securitydata.py
```diff
@@ -253,9 +245,86 @@ def show(state, search_id):
...
+        saved_search,
+        or_query,
+        output_function=send_events,
+        **kwargs,
+    )
+    if not handlers.TOTAL_EVENTS and not errors.ERRORED:
+        echo("No results found.")
+    else:
+        output_func = handlers.output_func
+        output_func(format, hostname, protocol, SEARCH_DEFAULT_HEADER)
```
alanag13, line 285: I'm a little bit confused by this, I would think that outputting would be handled by `_extract_events`, especially since `send_events` is passed to it. Why does this get called here?

src/code42cli/cmds/search/output_processor.py
```diff
@@ -0,0 +1,51 @@
...
+        paginate = _process_events(output_format, events, output_header)
+
+        url_parts = get_url_parts(hostname)
+        port = url_parts[1] or 514
+        try:
+            handler = NoPrioritySysLogHandlerWrapper(
+                url_parts[0], port=port, protocol=protocol
+            ).handler
+        except Exception as e:
+            raise Exception(
```
alanag13, line 45: I'm pretty sure that no connection attempt is made during the construction of this object, so I would not expect this to result in an exception ever being thrown. @unparalleled-js made this wrapper, is that correct?

src/code42cli/cmds/securitydata.py
alanag13: I think `output_func` should be called from within `handle_response` -- unless I'm misunderstanding something, I think that this will only output the last page.

src/code42cli/cmds/search/output_processor.py
kiran-chaudhary: Changed.

src/code42cli/cmds/securitydata.py
kiran-chaudhary: Changed.

src/code42cli/cmds/search/output_processor.py
kiran-chaudhary: Yes, changed.

src/code42cli/cmds/search/output_processor.py
```diff
@@ -0,0 +1,52 @@
...
+        paginate = _process_events(output_format, events, output_header)
+
+        url_parts = get_url_parts(hostname)
+        port = url_parts[1] or 514
+        try:
+            handler = NoPrioritySysLogHandler(
+                url_parts[0], port=port, protocol=protocol
+            )
+            for page in paginate():
+                handler.emit(page)
```
unparalleled-js, line 45: Instead of using the handler, just use the logger.

Look at the old code for `logger_factory` in the CLI search directory and find the method that creates a logger. THis will probably be the easier route.

CHANGELOG.md
```diff
@@ -48,6 +48,7 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
     - `code42 legal-hold list`
     - `code42 legal-hold show`
     - `code42 security-data saved-search list`
+- Added back `send-to` subcommand to `alerts` and `security-data` command to output to server via UDP or TCP.
```
unparalleled-js, line 51: I'd mention the exact command and available args.

```
Re-added `send-to` command to `alerts` and `security-data` that accepts a host address and a `--protocol` option with choices UDP or TCP.
```

src/code42cli/cmds/securitydata.py
```diff
@@ -196,33 +199,19 @@ def search(
...
+        begin,
+        end,
+        advanced_query,
         use_checkpoint,
-        include_all,
-        output_format=format,
-        output_header=SEARCH_DEFAULT_HEADER,
+        saved_search,
+        or_query,
+        output_function=print_events_decorator,
```
unparalleled-js, line 211: Shouldn't need an output function, only the handlers need to know how to output the events.

Create different handlers for `send-to` and have it use the SyslogLogger rather than echo... There will likely be logic sharing there

src/code42cli/cmds/search/output_processor.py
```diff
@@ -0,0 +1,34 @@
...
+    """Sends events to server/hostname"""
+
+    def decorator(events):
+        def paginate():
+            yield format_function(events, output_header)
+
+        logger = get_logger_for_server(hostname, protocol, output_format)
+
+        for page in paginate():
+            logger.handlers[0].emit(page)
```
kiran-chaudhary, line 32: `logger.info` doesn't fail but doesn't even transfer the data so I had managed it somehow. I will continue work on the same.

src/code42cli/cmds/alerts.py
```diff
@@ -186,14 +192,96 @@ def search(
         include_all, SEARCH_DEFAULT_HEADER, format
     )
     format_func = get_output_format_func(format)
+    print_events_decorator = print_events(format_func, output_header)
+    handlers = _extract_events(
```
kiran-chaudhary, line 196: This is just a refactor, no big change in search. Created `_extract_events` to make sure `search` and `send-to` use the same code.

src/code42cli/cmds/search/output_processor.py
kiran-chaudhary: Changes done, just one minor fix remaining, will investigate the same tomorrow.

src/code42cli/cmds/search/extraction.py
```diff
@@ -1,6 +1,5 @@
 import json
```
unparalleled-js, line 1: Kiran, this was totally my fault, but can you fix line 26?  The error message is backwards.  It should say `--include-all only allowed for Table output format`.

src/code42cli/logger.py
```diff
@@ -14,6 +21,21 @@
...
+    else:
+        return FileEventDictToRawJSONFormatter()
+
+
+def _init_logger(logger, handler, output_format):
+    formatter = _get_formatter(output_format)
+    logger.setLevel(logging.INFO)
+    return add_handler_to_logger(logger, handler, formatter)
+
+
```
unparalleled-js, line 38: Can you delete `get_logger_for_stdout()` while you are at it? We don't use that anymore

src/code42cli/logger.py
```diff
@@ -40,6 +62,36 @@ def get_logger_for_stdout(name_suffix="main", formatter=None):
     return logger
 
 
+def get_logger_for_server(hostname, protocol, output_format):
```
unparalleled-js, line 65: Figured out so far that RAW-JSON works perfectly but none of the other formats do, which tells me this is an error with the Formatter that is causing this not to function.... will keep digging

src/code42cli/cmds/search/output_processor.py
```diff
@@ -0,0 +1,34 @@
...
+
+    return decorator
+
+
+def send_events(output_format, hostname, protocol, output_header, format_function):
+    """Sends events to server/hostname"""
+
+    def decorator(events):
+        def paginate():
+            yield format_function(events, output_header)
```
unparalleled-js, line 27: This is the reason for your bug.

Do this `yield events` instead.

The formatting takes place at the logging level.  We are instead doing it twice here which causes the second time to error out.

src/code42cli/cmds/search/output_processor.py
unparalleled-js: Make this `logger.info()` and follow other advice

unparalleled-js, line 32: For this part, do it this way instead        
```
for page in paginate():
            for event in page:
                logger.info(event):
```

unparalleled-js: We are supposed to log each individual event, not each page of events.

unparalleled-js:
Figured out why your code wasn't working. See comments!

src/code42cli/cmds/search/output_processor.py
unparalleled-js: However, if you temporarily change this part to be `yield events` and line 32 to 
```
for page in paginate():
    for event in page:
        logger.info(event)
```
you would see it working

src/code42cli/cmds/securitydata.py
```diff
@@ -267,9 +262,98 @@ def show(state, search_id):
...
+
+def _extract_events(
+    state,
+    begin,
+    end,
+    advanced_query,
+    use_checkpoint,
+    saved_search,
+    or_query,
+    output_function,
```
unparalleled-js, line 333: See https://github.com/code42/code42cli/pull/137#discussion_r474188218

Instead of taking output_function`, make `send_to()` create different handlers using `extraction.py`. It will be the implementation much more straightforward with less decorators /passing function around.  I will add a comment in `extraction.py` will additional info

src/code42cli/cmds/search/extraction.py
```diff
@@ -45,7 +44,7 @@ def _get_alert_details(sdk, alert_summary_list):
 
 
 def create_handlers(
```
unparalleled-js, line 46: Instead of changing this implementation, make a new method for `create_server_handlers()` or `create_send_to_handlers()` and create the logger inside there and all you need to do it call `logger.info()` on one of the event dicts, and the behavior is then restored.

unparalleled-js:
Give everything a shot, let me know how it goes, push it all, and then I will look at it again.

src/code42cli/cmds/search/output_processor.py
kiran-chaudhary: Done.

src/code42cli/cmds/securitydata.py
kiran-chaudhary: Done.

src/code42cli/cmds/search/extraction.py
kiran-chaudhary: Done

src/code42cli/cmds/securitydata_output_formats.py
```diff
@@ -19,7 +19,9 @@ def get_file_events_output_format_func(value):
 
 
 def to_cef(output, header):
-    return [_convert_event_to_cef(e) for e in output]
```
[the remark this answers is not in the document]
kiran-chaudhary: Done.

tests/cmds/test_alerts.py
kiran-chaudhary: added.

src/code42cli/cmds/search/output_processor.py
```diff
@@ -0,0 +1,17 @@
+import click
```
unparalleled-js, line 1: This should no longer be needed

unparalleled-js:
@kiran-chaudhary This is looking a lot better. We should merge in the rest of output formats and then it will be easier to finish this review

src/code42cli/cmds/securitydata.py
```diff
@@ -211,31 +216,21 @@ def search(
...
-    cursor = (
-        _get_file_event_cursor_store(state.profile.name) if use_checkpoint else None
-    )
-    handlers = ext.create_handlers(
-        state.sdk,
-        FileEventExtractor,
-        cursor,
+    print_events_decorator = print_events(format_func, output_header)
+
+    handlers = _extract_events(
```
unparalleled-js, line 221: Since this is only used for `search`, we don't need a separate method like this... Mostly, I'd like to get rid of `output_processor`

src/code42cli/cmds/search/extraction.py
```diff
@@ -124,3 +116,54 @@ def create_time_range_filter(filter_cls, begin_date=None, end_date=None):
 
     elif end_date and not begin_date:
         return filter_cls.on_or_before(end_date)
+
+
+def create_send_to_handlers(
```
unparalleled-js, line 121: This could be more dry..

Some shared code between both methods that create handlers

src/code42cli/cmds/alerts.py
```diff
@@ -186,14 +192,53 @@ def search(
...
+)
+@opt.sdk_options()
+@server_options
+@click.option(
+    "--include-all",
+    default=False,
+    is_flag=True,
+    help="Display simple properties of the primary level of the nested response.",
+)
+@send_to_options
```
unparalleled-js, line 224: Alerts does not support CEF

src/code42cli/cmds/search/extraction.py
kiran-chaudhary: I created `_extract_events` and `output_processor` to use shared code but then that is not acceptable [here](https://github.com/code42/code42cli/pull/137#discussion_r475769356) and I was advised to separate them out.

Anyways I will be doing the suggested changes.

src/code42cli/cmds/alerts.py
kiran-chaudhary: Removed.

src/code42cli/cmds/securitydata.py
kiran-chaudhary: Changed. Removed.

src/code42cli/cmds/search/extraction.py
kiran-chaudhary: Done some changes.

src/code42cli/cmds/search/extraction.py
```diff
@@ -72,9 +75,7 @@ def create_handlers(
     extractor = extractor_class(sdk, ExtractionHandlers())
     handlers = _set_handlers(cursor_store, checkpoint_name)
 
-    @warn_interrupt(
-        warning="Attempting to cancel cleanly to keep checkpoint data accurate. One moment..."
-    )
+    @warn_interrupt(warning=INTERRUPT_WARNING)
```
kiran-chaudhary, line 78: An observation. If the interrupt warning would be printed when "CTRL-C" is pressed after command execution then it isn't, atleast for me.

src/code42cli/output_formats.py
```diff
@@ -10,16 +10,29 @@
...
+
+class OutputFormat(JsonOutputFormat):
+    TABLE = "TABLE"
+    CSV = "CSV"
+
     def __iter__(self):
         return iter([self.TABLE, self.CSV, self.JSON, self.RAW])
 
 
+class SendToOutputFormat(JsonOutputFormat):
```
unparalleled-js, line 29: Perhaps this should be called `SendToFileEventsOutputFormat` since it doesn't apply to alerts

src/code42cli/output_formats.py
```diff
@@ -38,8 +51,11 @@ def get_output_format_func(value):
 def to_csv(output, header):
     if not output:
         return
+    fieldnames = output[0].keys()
+    # if header:
```
unparalleled-js, line 55: commented code

src/code42cli/cmds/search/extraction.py
unparalleled-js: Like if you hit CTRL-C twice? I think that is expected

src/code42cli/cmds/search/extraction.py
unparalleled-js: I think it is better here-- less args needed

unparalleled-js: Thanks

src/code42cli/cmds/search/extraction.py
```diff
@@ -124,3 +129,34 @@ def create_time_range_filter(filter_cls, begin_date=None, end_date=None):
...
+            except Exception as ex:
+                handlers.handle_error(ex)
+
+        total_events = len(events)
+        handlers.TOTAL_EVENTS += total_events
+
+        for event in events:
+            logger.info(event)
+
+        # To make sure the extractor records correct timestamp event when `CTRL-C` is pressed.
```
unparalleled-js, line 156: This part can be a separate function shared between both handlers

unparalleled-js: lines 156 - 159

src/code42cli/cmds/search/extraction.py
```diff
@@ -124,3 +129,34 @@ def create_time_range_filter(filter_cls, begin_date=None, end_date=None):
...
+    sdk, extractor_class, cursor_store, checkpoint_name, logger
+):
+    extractor = extractor_class(sdk, ExtractionHandlers())
+    handlers = _set_handlers(cursor_store, checkpoint_name)
+
+    @warn_interrupt(warning=INTERRUPT_WARNING)
+    def handle_response(response):
+        response_dict = json.loads(response.text)
+        events = response_dict.get(extractor._key)
+        if extractor._key == "alerts":
```
unparalleled-js, line 144: Lines 144 - 148 seems like they can be moved to a separate function and called by both `create_handlers()` methods

src/code42cli/cmds/search/extraction.py
```diff
@@ -124,3 +129,34 @@ def create_time_range_filter(filter_cls, begin_date=None, end_date=None):
...
+            try:
+                events = _get_alert_details(sdk, events)
+            except Exception as ex:
+                handlers.handle_error(ex)
+
+        total_events = len(events)
```
unparalleled-js, line 150: Recording the totla events can be moved and shared between both `create_handlers()` methods

src/code42cli/cmds/securitydata.py
```diff
@@ -267,6 +282,59 @@ def show(state, search_id):
...
+    end,
+    advanced_query,
+    use_checkpoint,
+    saved_search,
+    or_query,
+    **kwargs
+):
+    """Send events to the given server address."""
+    logger = get_logger_for_server(hostname, protocol, format)
+    cursor = (
```
unparalleled-js, line 315: Might be worth making a shared internal method for getting the cursor that can be shared between both `send_to` and `search`

src/code42cli/cmds/securitydata.py
```diff
@@ -267,6 +282,59 @@ def show(state, search_id):
...
+        _get_file_event_cursor_store(state.profile.name) if use_checkpoint else None
+    )
+    handlers = ext.create_send_to_handlers(
+        state.sdk, FileEventExtractor, cursor, use_checkpoint, logger
+    )
+    extractor = _get_file_event_extractor(state.sdk, handlers)
+    extractor.use_or_query = or_query
```
unparalleled-js, line 322: A lot of this can be shared.

src/code42cli/cmds/search/extraction.py
unparalleled-js: When I made that original comment, that shared method was not actually being shared. Creating the separate handlers makes it so that we don't need to pass in an `output_processor`. I found that to be confusing because `handle_response` is an output processor, and if you need a different output processor, you should create extra handlers, which you did (thank you). That does not mean the methods that create handlers can't share code or that the methods like `send_to` and `search` can use shared functions either.  I am sorry for that confusion, not trying to give you contradicting advice..

unparalleled-js:
Some code could be drier but otherwise looks good! Especially once the conflicts are resolved.

kiran-chaudhary:
Please test command `code42 security-data search -f json --saved-search a083f08d-8f33-4cbd-81c4-8d1820b61185` to confirm paging behaviour when record size is quite large. 

Also need to confirm whether `saved-search` is applicable to `send-to` command??

src/code42cli/cmds/alerts.py
```diff
@@ -195,17 +222,46 @@ def search(
...
+    **kwargs
+):
+    """Send alerts to the given server address."""
+    logger = get_logger_for_server(hostname, protocol, format)
+    cursor = _get_alert_cursor_store(cli_state.profile.name) if use_checkpoint else None
+    handlers = ext.create_send_to_handlers(
+        cli_state.sdk, AlertExtractor, cursor, use_checkpoint, logger,
+    )
+    _call_extractor(cli_state, handlers, begin, end, or_query, advanced_query, **kwargs)
+    if not handlers.TOTAL_EVENTS and not errors.ERRORED:
```
unparalleled-js, line 264: NOT A BLOCKER, but this could also be a separate shared method like `_handle_if_no_events()`

kiran-chaudhary: I would like to make this method as below, should it be in `util` or if it is specific to search then shall I create a new `util` in `search`? This is so that it can apply both to alerts and security-data. 

```
def _handle_no_events(no_events:bool):
      if no_events:
          echo("No results found.")
```


and then it will be called as `_handle_no_events(not handlers.TOTAL_EVENTS and not errors.ERRORED)`

unparalleled-js: I'd put it in `cmds.search.extraction` , which is like the util for only search related things, but yeah that looks like a good implementation

src/code42cli/cmds/search/extraction.py
kiran-chaudhary: found this in while executing `search` command with  `saved-search` option. It works.

src/code42cli/cmds/alerts.py
kiran-chaudhary: Done.

unparalleled-js:
Screen Shot 2020-08-26 at 15 03 13

^ this is what happens when I do `code42 security-data search -f json --saved-search a083f08d-8f33-4cbd-81c4-8d1820b61185`.  It does not occur from Master however... I will look at it for the next hour to try and diagnose.

And yes, saved-search should work with `send-to`.. Is that going to be a lot of re-work, or should that be do-able?

unparalleled-js:
And yes, I think we do need saved-search to work for send-to. Is that going to be any re-work? How hard would that be?

src/code42cli/cmds/search/extraction.py
unparalleled-js: I never see the interrupt warning, weird... probably address this outside this PR

unparalleled-js (approved):

kiran-chaudhary:
> And yes, I think we do need saved-search to work for send-to. Is that going to be any re-work? How hard would that be?

it works right now.

PR #138, opened 2020-08-17
Bugfix/put back cef

unparalleled-js:
We had lost CEF when we changed how did output formats. They were different in that we were using loggers before and switched to using `echo()` and `echo_via_pager()`.  However, all the CEF mapping worked best by importing logger formatters from the extractor module.  Therefore, I ported over the logic for doing CEF conversions here.

* Ports over CEF logic
* Moves around methods/classes to be more understandable
* Adds a few missing tests


unparalleled-js:
Does anyone do any scripting with CEF format for events? If so, perhaps the CEF stuff should move to py42.  I am not exactly sure what to do I guess.  Keep in mind that the original CEF implementation worked with `LogRecord` objects, not lists.

src/code42cli/cmds/search/enums.py
```diff
@@ -1,21 +1,13 @@
+from code42cli.output_formats import OutputFormat
+
 IS_CHECKPOINT_KEY = "use_checkpoint"
 
 
-class OutputFormat:
+class SecurityDataOutputFormat(OutputFormat):
```
unparalleled-js, line 6: Do we have CEF alerts?

src/code42cli/cmds/securitydata.py
```diff
@@ -1,9 +1,15 @@
 from _collections import OrderedDict
+from datetime import datetime
 from pprint import pformat
 
 import click
 import py42.sdk.queries.fileevents.filters as f
 from c42eventextractor.extractors import FileEventExtractor
+from c42eventextractor.logging.formatters import CEF_TEMPLATE
```
unparalleled-js, line 8: This stuff is still in the extractor... but perhaps should also be ported over

unparalleled-js:
Is there a way to do this without porting over the CEF mapping logic?  Porting just seemed like the best course of action to me since we need to change functions from expecting a log record to expecting a list and also because we have to do it anyway

src/code42cli/cmds/search/enums.py
alanag13: No, just file events

alanag13:
> Is there a way to do this without porting over the CEF mapping logic? Porting just seemed like the best course of action to me since we need to change functions from expecting a log record to expecting a list and also because we have to do it anyway

This is fine for now since the extractor will probably die soon-ish anyway. One idea I've been thinking about is making a "file events mapping" package, there are a lot of other formats besides CEF that SST is interested in mapping to, and a mapping-based repo would be easy enough for just about anyone to contribute to.

This has exposed that doing an implementation that only allowed for this conversion to happen via loggers was probably a mistake. If make something like that in the future, we should probably have it output bare strings.

src/code42cli/output_formats.py
```diff
@@ -4,29 +4,29 @@
...
+    def __iter__(self):
+        return iter([self.TABLE, self.CSV, self.JSON, self.RAW])
+
+
+def output_format(_, __, value):
     if value is not None:
+        value = value.upper()
         if value == OutputFormat.CSV:
-            return to_dynamic_csv
+            return to_csv
```
unparalleled-js, line 29: This looks like a change, but the diff is just being funny. The regular `output_format()` uses `to_csv`, and the `extraciton_output_format` uses `to_dynamic_csv`

src/code42cli/output_formats.py
```diff
@@ -4,29 +4,29 @@
...
-            return to_table
-        if value == OutputFormat.JSON:
-            return to_formatted_json
-    # default option
-    return to_table
+CEF_DEFAULT_PRODUCT_NAME = "Advanced Exfiltration Detection"
+CEF_DEFAULT_SEVERITY_LEVEL = "5"
 
 
-def extraction_output_format(_, __, value):
```
unparalleled-js, line 26: Moved `extraction_output_format` to a new module `cmds/search/output_format`

src/code42cli/cmds/search/enums.py
```diff
@@ -1,21 +1,13 @@
...
-        return iter([self.CEF, self.JSON, self.RAW])
-
-
-class AlertOutputFormat:
-    JSON = "JSON"
-    RAW = "RAW-JSON"
 
     def __iter__(self):
-        return iter([self.JSON, self.RAW])
+        return iter([self.TABLE, self.CSV, self.JSON, self.RAW, self.CEF])
```
kiran-chaudhary, line 10: Output formats are not only restricted to `alerts/security-data` commands. They were also used elsewhere, so CEF would also be supported in those. I am ok with it, making sure everyone is on same page.

Naming it as `AlertOutputFormat` in that case might not suffice.

alanag13: CEF is a format that is only implemented for / only makes sense for security events -- it isn't a data format that can be applied to everything like JSON or CSV can, so we would want this to remain specific to security events.

unparalleled-js: This is only used by File Events

unparalleled-js: I can move it somewhere to be more obviois

src/code42cli/cmds/securitydata_output_formats.py
```diff
@@ -0,0 +1,100 @@
...
+from c42eventextractor.maps import FILE_EVENT_TO_SIGNATURE_ID_MAP
+from c42eventextractor.maps import JSON_TO_CEF_MAP
+
+import code42cli.cmds.search.enums as enum
+from code42cli.cmds.search.output_formats import extraction_output_format
+from code42cli.output_formats import CEF_DEFAULT_PRODUCT_NAME
+from code42cli.output_formats import CEF_DEFAULT_SEVERITY_LEVEL
+
+
+def file_events_output_format(_, __, value):
```
alanag13, line 15: are there any tests around `to_cef`? I know this is largely just a copy/paste over from the extractor -- the tests from there could also be moved over, or we could make our own here. They probably don't have to be exhaustive as what we have in the extractor, but something that at least does a sanity check that proves that converting to CEF doesn't bomb completely is probably a good idea.

unparalleled-js: Yes, there are. I copied all the tests over.

unparalleled-js: `cmds/test_securitydata_output_formats.py`

alanag13: ah thanks, i missed it because the diff was large so github didn't show it.

unparalleled-js:
`Include-all` can now only be used with table format.

src/code42cli/cmds/search/extraction.py
```diff
@@ -118,7 +130,7 @@ def create_time_range_filter(filter_cls, begin_date=None, end_date=None):
         return filter_cls.on_or_before(end_date)
 
 
-def _process_events(output_format, include_all, events, output_header):
-    if include_all:
-        output_header = get_dynamic_header(events[0])
-    return output_format(events, output_header)
+# def _process_events(output_format, include_all, events, output_header):
```
kiran-chaudhary, line 133: if this is not needed anymore it can be removed.

src/code42cli/cmds/securitydata_output_formats.py
```diff
@@ -0,0 +1,100 @@
+from datetime import datetime
```
kiran-chaudhary, line 1: filename `securitydata_...` doesn't seem apt, it should be `CEF_...`  because this format even though today is supported only for security data, it can be supported elsewhere. or if we want proper modularisation then `cmd/output_formats/CEF.py`

src/code42cli/output_formats.py
```diff
@@ -94,6 +77,7 @@ def to_formatted_json(output, header=None):
 
 
 def get_dynamic_header(header_items):
+    print(header_items)
```
kiran-chaudhary, line 80: print

kiran-chaudhary (changes requested):
few minor stuff like print statements, etc and suggested a naming convention change.

src/code42cli/cmds/securitydata_output_formats.py
unparalleled-js: This specific module only has to do with file events.  Mapping to CEF is very business-logic-y and will vary a lot with each data set.

src/code42cli/cmds/securitydata_output_formats.py
unparalleled-js: That is because we are so good at writing tons of tests haha

src/code42cli/cmds/securitydata_output_formats.py
unparalleled-js: Also contains `file_events_output_format` which could potentially include other variances of file event-realted output formats

kiran-chaudhary: Then its fine.

src/code42cli/cmds/securitydata.py
```diff
@@ -37,6 +36,15 @@
 SEARCH_DEFAULT_HEADER["sha256Checksum"] = "SHA256Checksum"
 
 
+file_events_format_option = click.option(
+    "-f",
+    "--format",
+    type=click.Choice(enum.FileEventsOutputFormat(), case_sensitive=False),
+    help="The output format of the result. Defaults to table format.",
+    callback=file_events_output_format,
```
kiran-chaudhary, line 44: callback

src/code42cli/cmds/search/extraction.py
```diff
@@ -8,14 +8,26 @@
...
 from code42cli.util import warn_interrupt
 
 logger = get_main_cli_logger()
 
 _ALERT_DETAIL_BATCH_SIZE = 100
 
 
+def handle_include_all(include_all, default_header, format_func):
+    output_header = None
+    if format_func.__name__ == to_table.__name__:
```
kiran-chaudhary, line 21: if we want to do away with using `__name__`, we can get rid of [callback](https://github.com/code42/code42cli/pull/138/files#r473037679) argument and define another method that maps the function to the options.

Just an alternative suggestion.

unparalleled-js: I will try this

unparalleled-js: I did this

CONTRIBUTING.md
```diff
@@ -183,6 +183,6 @@ Document all notable consumer-affecting changes in CHANGELOG.md per principles a
 
 ## Opening a PR
 
-When you're satisified with your changes, open a PR and fill out the pull request template file. We recommend prefixing the name of your branch and/or PR title with `bugfix`, `chore`, or `feature` to help quickly categorize your change. Your unit tests and other checks will run against all supported python versions when you do this.
+When you're satisfied with your changes, open a PR and fill out the pull request template file. We recommend prefixing the name of your branch and/or PR title with `bugfix`, `chore`, or `feature` to help quickly categorize your change. Your unit tests and other checks will run against all supported python versions when you do this.
```
unparalleled-js, line 186: I feel like I've fixed this sp. at least four times now

src/code42cli/cmds/securitydata.py
```diff
@@ -149,7 +160,7 @@ def file_event_options(f):
     f = process_owner_option(f)
     f = tab_url_option(f)
     f = include_non_exposure_option(f)
-    f = extraction_format_option(f)
+    f = file_events_format_option(f)
```
unparalleled-js, line 163: Alerts uses the standard format option (doesn't support CEF), so that is why this is renamed

src/code42cli/output_formats.py
```diff
@@ -2,31 +2,29 @@
...
 import json
 
-import click
-
-from code42cli.cmds.enums import OutputFormat
 from code42cli.util import find_format_width
 from code42cli.util import format_to_table
 
 
-def output_format(_, __, value):
```
unparalleled-js, line 12: moved below and given a more function-sounding name

src/code42cli/output_formats.py
```diff
@@ -37,46 +35,20 @@ def extraction_output_format(_, __, value):
...
-extraction_format_option = click.option(
-    "-f",
-    "--format",
-    type=click.Choice(OutputFormat(), case_sensitive=False),
-    help="The output format of the result. Defaults to table format.",
-    callback=extraction_output_format,
-)
-
-
-def to_dynamic_csv(output, header):
```
unparalleled-js, line 58: @kiran-chaudhary No need for dynamic_csv, as csv always is dynamic now (no custom headers, always gets all the data).  To get a subset of the data, you have to use TABLE format

tests/cmds/test_alert_rules.py
```diff
@@ -303,4 +303,9 @@ def test_list_cmd_formats_to_csv_when_format_is_passed(runner, cli_state):
     cli_state.sdk.alerts.rules.get_all.return_value = [TEST_RULE_RESPONSE]
     result = runner.invoke(cli, ["alert-rules", "list", "-f", "csv"], obj=cli_state)
     assert cli_state.sdk.alerts.rules.get_all.call_count == 1
-    assert "RuleId,Name,Severity,Type,Source,Enabled" in result.output
+    assert "observerRuleId" in result.output
```
unparalleled-js, line 306: have to do it this way because of cursed python 3.5

tests/cmds/search/test_extraction.py
```diff
@@ -47,33 +70,3 @@ def _get_timestamp_from_item(self, item):
     py42_response = Py42Response(http_response)
     handlers.handle_response(py42_response)
     output_format.assert_called_once_with(events, header)
-
-
-def test_include_all_creates_dynamic_header_from_events_to_output_format(
```
unparalleled-js, line 52: Creating handlers no longer needs to know about include-all

kiran-chaudhary (approved):
LGTM.

PR #139, opened 2020-08-19
Update gettingstarted.md

annie-payseur:
Added more details on Keyring/credential storage per OS.

### DIFF
--- a/docs/userguides/gettingstarted.md
+++ b/docs/userguides/gettingstarted.md
@@ -69,14 +69,47 @@ python3 -m pip install code42cli --upgrade
 ## Authentication
 
 ```eval_rst
-.. important:: the Code42 CLI currently only supports token-based authentication.
+.. important:: The Code42 CLI currently only supports token-based authentication.
 ```
 
-To use the CLI, you must provide your credentials (basic authentication). The CLI uses keyring when storing passwords.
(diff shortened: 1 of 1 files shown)


docs/userguides/gettingstarted.md
```diff
@@ -69,15 +69,48 @@ python3 -m pip install code42cli --upgrade
...
+```bash
+yum -y install python-pip python3 dbus-python gnome-keyring libsecret dbus-x11
+pip3 install code42cli
+```
+If the following directories do not already exist, create them:
+```bash
+mkdir -p ~/.cache
+mkdir -p ~/.local/share/keyring
+```
+In the following commands, replace the example value `\n` with the Keyring password (if the default Keyring already exists).
```
unparalleled-js, line 96: Why do we use the endline as the example value?

unparalleled-js (approved):
Had a small question about \n, throws me off a little why that is the default value in the example but maybe I just don't understand the convention.

Anyway, thank you!

alanag13:
> Had a small question about \n, throws me off a little why that is the default value in the example but maybe I just don't understand the convention.

`\n` is treated the same as simply hitting return, which would create (or unlock) the keyring without a password.

PR #143, opened 2020-08-21
Bugfix/json output

unparalleled-js:
Fixes:

* fixes bug where formatted JSON was not always outputting all of the data
* Bug where JSON format would simply not work.
* Bug where we were not outputting lists very well (was showing brackets)
* General non-realistic legal hold tests fixed

I have tested this quite a bit but am tired and will test more on Monday.

### DIFF
--- a/src/code42cli/cmds/search/extraction.py
+++ b/src/code42cli/cmds/search/extraction.py
@@ -53,7 +53,7 @@ def create_handlers(
         if hasattr(exception, "response") and hasattr(exception.response, "text"):
             message = "{}: {}".format(exception, exception.response.text)
         else:
-            message = exception
+            message = exception()
         logger.log_error(message)
         secho(str(message), err=True, fg="red")
 
@@ -87,7 +87,9 @@ def create_handlers(
             click.echo_via_pager(_format_output())
         else:
             for page in _format_output():
-                click.echo(page)
+                click.echo(page, nl=False)
+            if formatter.output_format == OutputFormat.TABLE:
+                click.echo()
 
         # To make sure the extractor records correct timestamp event when `CTRL-C` is pressed.
         if total_events:
--- a/src/code42cli/cmds/search/extraction.py
+++ b/src/code42cli/cmds/search/extraction.py
@@ -19,11 +19,8 @@ _ALERT_DETAIL_BATCH_SIZE = 100
 def try_get_default_header(include_all, default_header, output_format):
     """Returns appropriate header based on include-all and output format. If returns None,
     the CLI format option will figure out the header based on the data keys."""
-    output_header = None
-
-    if output_format == OutputFormat.TABLE and not include_all:
-        output_header = default_header
-    elif output_format != OutputFormat.TABLE and include_all:
+    output_header = None if include_all else default_header
+    if output_format != OutputFormat.TABLE and include_all:
         err_text = "--include-all only allowed for non-Table output formats."
         logger.log_error(err_text)
         raise errors.Code42CLIError(err_text)
--- a/src/code42cli/cmds/search/extraction.py
+++ b/src/code42cli/cmds/search/extraction.py
@@ -42,7 +42,7 @@ def _get_alert_details(sdk, alert_summary_list):
 
 
 def create_handlers(
-    sdk, extractor_class, cursor_store, checkpoint_name, formatter,
+    sdk, extractor_class, cursor_store, checkpoint_name, formatter, force_pager,
 ):
     extractor = extractor_class(sdk, ExtractionHandlers())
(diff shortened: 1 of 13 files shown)


tests/cmds/test_securitydata_output_formats.py
```diff
@@ -106,42 +106,61 @@
...
+        self,
+    ):
+        formatter = FileEventsOutputFormatter(FileEventsOutputFormat.RAW)
+        assert id(formatter._format_func) == id(to_json)
+
+    def test_init_sets_format_func_to_cef_function_when_cef_format_option_is_passed(
+        self,
+    ):
+        formatter = FileEventsOutputFormatter(FileEventsOutputFormat.CEF)
+        print(formatter._format_func.__name__)
```
kiran-chaudhary, line 150: print

kiran-chaudhary (commented):
apart from the print LGTM. Tested few commands too.

src/code42cli/output_formats.py
```diff
@@ -20,35 +20,59 @@ def __iter__(self):
...
-
-def to_csv(output, header):
+class OutputFormatter:
+    def __init__(self, output_format, header=None):
+        output_format = output_format.upper() if output_format else OutputFormat.TABLE
+        self.output_format = output_format
+        self._format_func = to_table
+        self.header = header
+        if output_format is None:
+            return
```
timabrmsn, line 30: I'm not seeing a way `output_format` could ever be `None` here.

src/code42cli/output_formats.py
```diff
@@ -20,35 +20,59 @@ def __iter__(self):
...
+        else:
+            for item in output:
+                yield self._format_output(item)
+
+    @property
+    def _requires_list_output(self):
+        return (
+            self.output_format == OutputFormat.TABLE
+            or self.output_format == OutputFormat.CSV
+        )
```
timabrmsn, line 56: could do: `return self.output_format in (OutputFormat.TABLE, OutputFormat.CSV)`

src/code42cli/output_formats.py
```diff
@@ -20,54 +20,78 @@ def __iter__(self):
...
+    """Output is a single record"""
     return json.dumps(output)
 
 
-def to_formatted_json(output, header=None):
-    return json.dumps(_filter(output, header), indent=4)
+def to_formatted_json(output, header):
+    """Output is a single record"""
+    json_str = json.dumps(output, indent=4)
+    return json_str
```
timabrmsn, line 83: We should add a newline to both this output and the `to_json` output, as right now all events are being concat-ed together without a separator.

src/code42cli/output_formats.py
```diff
@@ -20,54 +20,78 @@ def __iter__(self):
...
-    return {
-        key: key.capitalize()
-        for key in header_items.keys()
-        if type(header_items[key]) == str
-    }
+    for item in header_items:
+        keys = item.keys()
+        for key in keys:
+            if key not in header and isinstance(key, str):
+                header[key] = key.capitalize()
```
timabrmsn, line 96: Just calling capitalize on the string here causes the mid-word capital letters to get lowered. So the headers right now without the `--include-all` option look like `FileName, FilePath, EventTimestamp`, but with `--include-all` they're: `Filename, Filepath, Eventtimestamp`, etc.

I'd rather us not try to transform them in either place, so if you're using the table headers to pull items from json, you know exactly what keys to look for.

src/code42cli/cmds/search/extraction.py
```diff
@@ -83,13 +80,13 @@ def handle_response(response):
...
-            yield format_func(events, output_header)
+        def _format_output():
+            return formatter.get_formatted_output(events)
 
         if len(events) > 10:
-            click.echo_via_pager(paginate)
+            click.echo_via_pager(_format_output())
         else:
-            for page in paginate():
+            for page in _format_output():
```
timabrmsn, line 89: This is what's causing the extra newlines I mentioned before. It's not just alerts, it's if there are less than 10 events since we're `echo`-ing them individually `echo` is adding a newline like `print` does. 

Changing to `click.echo(page, nl=False)` fixes that.

unparalleled-js: Ah, thanks for finding this!

unparalleled-js: We need to echo a new line at the very end though

unparalleled-js: Oh wait... only if it's table do we need a new line at the end... ugh

unparalleled-js:
@timabrmsn So are we good now?

unparalleled-js:
Tim said we are good!

PR #144, opened 2020-08-25
pin py42 for cli versions < 1.0 to < 1.8.0

alanag13:
py42 version 1.8 will introduce new exception types that will be thrown in py42 rather than in the cli. Because the existing releases do not catch these exceptions, some existing error situations would be logged to error files instead of having their messages output to the console. Pinning the py42 version until we create a release that properly catches the new py42 errors corrects this.


src/code42cli/__version__.py
```diff
@@ -1 +1 @@
-__version__ = "0.7.3"
+__version__ = "0.7.4"
```
unparalleled-js, line 1: Should we add an empty changelog?

alanag13: We'll add that entry to master separately since don't intend to merge this there.

unparalleled-js (approved):

PR #148, opened 2020-08-28
Fix/remove output formats from lg show

unparalleled-js:


src/code42cli/output_formats.py
```diff
@@ -95,5 +95,5 @@ def to_json(output):
 
 def to_formatted_json(output):
     """Output is a single record"""
-    json_str = "{}\n".format(json.dumps(output, indent=4))
```
unparalleled-js, line 98: @timabrmsn For some commands, this is needed, some not.... see the problem still in certain places

unparalleled-js: Trying to figure out how to fix this now

PR #149, opened 2020-08-28
Output formats

unparalleled-js:
* Fixes handling newlines
* Removed format options from `show`
* Corrects changelog entries
* Makes default send-to RAW

### DIFF
--- a/src/code42cli/cmds/search/extraction.py
+++ b/src/code42cli/cmds/search/extraction.py
@@ -100,10 +100,7 @@ def create_handlers(
         if total_events > 10 or force_pager:
             click.echo_via_pager(_format_output())
         else:
-            for page in _format_output():
-                click.echo(page, nl=False)
-            if formatter.output_format in [OutputFormat.TABLE]:
-                click.echo()
+            formatter.echo_formatters_output(_format_output())
 
         # To make sure the extractor records correct timestamp event when `CTRL-C` is pressed.
         if total_events:
--- a/src/code42cli/cmds/search/extraction.py
+++ b/src/code42cli/cmds/search/extraction.py
@@ -95,7 +95,7 @@ def create_handlers(
         handlers.TOTAL_EVENTS += total_events
 
         if total_events > 10 or force_pager:
-            click.echo_via_pager(events)
+            click.echo_via_pager(formatter.get_formatted_output(events))
         else:
             formatter.echo_formatted_list(events)
 
--- a/src/code42cli/cmds/search/extraction.py
+++ b/src/code42cli/cmds/search/extraction.py
@@ -94,13 +94,10 @@ def create_handlers(
         total_events = len(events)
         handlers.TOTAL_EVENTS += total_events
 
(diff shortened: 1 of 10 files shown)


src/code42cli/cmds/search/extraction.py
```diff
@@ -100,10 +100,7 @@ def _format_output():
         if total_events > 10 or force_pager:
             click.echo_via_pager(_format_output())
         else:
-            for page in _format_output():
-                click.echo(page, nl=False)
-            if formatter.output_format == OutputFormat.TABLE:
-                click.echo()
+            formatter.echo_formatters_output(_format_output())
```
timabrmsn, line 103: I think you mean to use `echo_formatted_list()` here?

alanag13: agreed, this doesn't appear to be a defined function

timabrmsn: and we're now calling the format_output function twice on this data, which is causing it to escape the quotations and newlines

src/code42cli/cmds/search/extraction.py
```diff
@@ -94,16 +94,10 @@ def handle_response(response):
         total_events = len(events)
         handlers.TOTAL_EVENTS += total_events
 
-        def _format_output():
-            return formatter.get_formatted_output(events)
-
         if total_events > 10 or force_pager:
-            click.echo_via_pager(_format_output())
+            click.echo_via_pager(events)
```
timabrmsn, line 98: This should still call format on the events, otherwise it's echoing the dicts instead of json.

alanag13:
it looks like most of our tests verify that py42 gets called with the right variables, but don't verify that anything gets output to the terminal for situations when they're supposed to. I'm not saying that we need to add those all right now because I've tested enough to know that the outputs work, but it's an observation I want to put out there because those are probably good tests to add.

unparalleled-js:
> it looks like most of our tests verify that py42 gets called with the right variables, but don't verify that anything gets output to the terminal for situations when they're supposed to. I'm not saying that we need to add those all right now because I've tested enough to know that the outputs work, but it's an observation I want to put out there because those are probably good tests to add.

There are countless tests that check for strings inside of `results.output`.

The tests would be even better if we didn't have to support python 3.5

alanag13:
> > it looks like most of our tests verify that py42 gets called with the right variables, but don't verify that anything gets output to the terminal for situations when they're supposed to. I'm not saying that we need to add those all right now because I've tested enough to know that the outputs work, but it's an observation I want to put out there because those are probably good tests to add.
> 
> There are countless tests that check for strings inside of `results.output`.
> 
> The tests would be even better if we didn't have to support python 3.5

You're right, thanks -- Its been a while since I spent some time in this code base so I was looking for tests that called `echo`, `echo_formatted_list`, etc and didn't see any but I see what you're talking about now.

unparalleled-js:
I agree there are missing tests though

PR #151, opened 2020-08-31
remove nc/powershell examples and replace with send-to in docs/readme

timabrmsn:


README.md
```diff
@@ -132,37 +132,19 @@ To write events to a file, just redirect your output:
 code42 security-data search -b 2020-02-02 > filename.txt
 ```
 
-To send events to an external server using `netcat` on Linux/Mac:
+To send events to an external server, use the `send-to` command, which behaves exactly the same as `search` but sends
```
unparalleled-js, line 135: The one difference between Search is that the default output format is RAW-JSON for send-to and I think it is just JSON for search

PR #153, opened 2020-08-31
handle detection list removal errors

timabrmsn:


src/code42cli/cmds/alert_rules.py
```diff
@@ -68,7 +69,12 @@ def add_user(state, rule_id, username):
 @sdk_options()
 def remove_user(state, rule_id, username):
     """Remove a user from an alert rule."""
-    _remove_user(state.sdk, rule_id, username)
+    try:
+        _remove_user(state.sdk, rule_id, username)
+    except Py42BadRequestError:
```
unparalleled-js, line 74: Is there a chance that the rule ID or username could somehow cause a bad request (from the CLI) for other reasons?

unparalleled-js:
Should mention this in the Changelog

src/code42cli/cmds/departing_employee.py
```diff
@@ -46,7 +47,14 @@ def add(state, username, cloud_alias, departure_date, notes):
 @sdk_options()
 def remove(state, username):
     """Remove a user from the departing-employee detection list."""
-    _remove_departing_employee(state.sdk, username)
+    try:
+        _remove_departing_employee(state.sdk, username)
+    except Py42NotFoundError:
```
unparalleled-js, line 52: in `err.response.text`, it says `Could not find item for user: <id>`... Probably want to leverage that in py42 at least.

unparalleled-js:
Testing passed

src/code42cli/cmds/alert_rules.py
timabrmsn: I just tested these cases and they both provide correct error messages.

PR #155, opened 2020-09-03
Handle updating cloud aliases when adding to a list

unparalleled-js:
Currently, when trying to use `--cloud-alias` when adding a Departing Employee or a High Risk Employee, it fails if the employee already has a cloud alias..
This will remove the other cloud aliases before adding the new one


src/code42cli/cmds/detectionlists/__init__.py
```diff
@@ -13,6 +13,11 @@ def update_user(sdk, username, cloud_alias=None, risk_tag=None, notes=None):
     """
     user_id = get_user_id(sdk, username)
     if cloud_alias:
```
unparalleled-js, line 15: The UI only allows you to have one additional alias... This removes all "additional" aliases before adding the new one, to mirror the Code42 UI.

timabrmsn (approved):

timabrmsn (approved):

PR #160, opened 2020-09-16
added audit-logs command

kiran-chaudhary:
**Description of Change**

Added `audit-logs` command in cli.

closes  #1179

**Testing Procedure**

`code42 audit-logs search -b 2020-08-09 --usernames 'username1' --usernames 'username2'`

**PR Checklist**


src/code42cli/cmds/auditlogs.py
```diff
@@ -0,0 +1,123 @@
+import click
+
+from code42cli.date_helper import parse_max_timestamp
+from code42cli.date_helper import parse_min_timestamp
+from code42cli.options import OrderedGroup
+from code42cli.options import sdk_options
+
+
+filter_option_usernames = click.option(
+    "--usernames", required=False, help="Filter results by usernames.", multiple=True,
```
kiran-chaudhary, line 10: since we set `multiple=True`, it allows us to specify more than one usernames in format....
`--usernames "u1" --usernames "u2" ` in such case should the argument name be singular or plural?? e.g `--usernames` or `--username`

src/code42cli/cmds/auditlogs.py
```diff
@@ -0,0 +1,123 @@
...
+        user_ids=user_ids,
+        user_ip_addresses=user_ips,
+        affected_user_ids=affected_user_ids,
+        affected_usernames=affected_usernames,
+    )
+
+
+def _search(sdk, format, **filter_args):
+
+    if format.upper() != "JSON":
```
kiran-chaudhary, line 118: Need to remove this case based on discussion in standup.

tests/cmds/test_auditlogs.py
```diff
@@ -0,0 +1,87 @@
...
+):
+    http_response = mocker.MagicMock(spec=Response)
+    http_response.text = '{"events": [{"property": "bar"}]}'
+    py42_response = Py42Response(http_response)
+    cli_state.sdk.auditlogs.get_all.return_value = [py42_response["events"]]
+    runner.invoke(
+        cli, ["audit-logs", "send-to", "localhost", "--begin", "1d"], obj=cli_state
+    )
+    assert cli_state.sdk.auditlogs.get_all.call_count == 1
+    # assert event_extractor_logger.info.assert_called_once_with({"property": "bar"})
```
kiran-chaudhary, line 87: `info` method is called but still it says called 0 times. Looking into it.

src/code42cli/cmds/search/extraction.py
```diff
@@ -162,3 +162,18 @@ def handle_response(response):
 def handle_no_events(no_events):
     if no_events:
         click.echo("No results found.")
+
+
+def create_simple_send_to_hanlder(logger, func, key, **kwargs):
```
kiran-chaudhary, line 167: Open to suggestions for an apt name here.

src/code42cli/cmds/auditlogs.py
kiran-chaudhary: removed.

src/code42cli/cmds/auditlogs.py
alanag13: I would stick with the same convention that we have on other commands. I believe everything else uses singular names, so we should be consistent and do that here, too.

src/code42cli/cmds/auditlogs.py
```diff
@@ -0,0 +1,159 @@
...
+    multiple=True,
+)
+filter_option_event_types = click.option(
+    "--event-types",
+    required=False,
+    help="Filter results by event types.",
+    multiple=True,
+)
+
+filter_option_begin_time = click.option(
```
alanag13, line 46: i think that begin / end time arguments exist in another place in the codebase, can we import them and share those definitions instead of copying them over here? Are the time formats for audit log the same as what is used for file events / alerts?

alanag13:
@kiran-chaudhary the minimum version of py42 will need to be bumped to be whatever the version is once the audit log functionality is released there.

src/code42cli/cmds/auditlogs.py
kiran-chaudhary: Refactored, there are subtle differences in usage so had skipped it in the first iteration. Thanks.

src/code42cli/cmds/auditlogs.py
kiran-chaudhary: changed.

kiran-chaudhary:
> @kiran-chaudhary the minimum version of py42 will need to be bumped to be whatever the version is once the audit log functionality is released there.

Thanks, yes I will merge after approval, until py42 is released.
> 
> one thing that I understand but is still pretty scary to me: this build does not use a version of py42 that has audit log in it, but all the tests still passed. I think that's a sign that we're relying on mocking a little too heavily.

If we add unit tests that require the necessary build dependencies available, it will block development.  I tested the same by pulling the required py42 code from the audit-logs branch. We shouldn't move it to testing until the version is bumped and integration test should be written to test such cases and dependencies.

unparalleled-js:
Should we add a required begin date like we do for file events?

unparalleled-js:
Should we tack on the `csv` format option?

unparalleled-js:
I am wondering why there is no Table output format as the default, rather than just json objects?

kiran-chaudhary:
> Should we tack on the `csv` format option?

`csv` and `cef` format are supported by the API itself, While implementing those in py42, we observed those options do not return paginated response, etc and hence had decided to skip them for now. I thought we would be following same approach over here.

kiran-chaudhary:
> There are a lot of results sometimes, should we use `echo_via_pager()` again if the result count is more than 10?

Yes, I can add that.

kiran-chaudhary:
> I am wondering why there is no Table output format as the default, rather than just json objects?

Yes, I am open to supporting them as we have done for other commands if we do not want to use the API supported format options.

alanag13:
I agree with adding all the items that @unparalleled-js called out.
Let's plan to cut a py42 release soon so that we can get the appropriate version set on this.

unparalleled-js:
I can get a strange error when I use `-e`  on its own without `-b`.  This action does not appear to be allowed by the server.  We should not allow this in the CLI and echo some error text when the user tries.

unparalleled-js:
When I try to use a format such as `2020-02-02` (not time), I can a strange error and we send `null` in the reqest... We should tack on zero seconds in this case so that this easy format is allowed.

unparalleled-js:
It appears that the `-b` and `-e` flags do not work. It always sends null in the request when I try this.

kiran-chaudhary:
> There are a lot of results sometimes, should we use `echo_via_pager()` again if the result count is more than 10?

Added.

kiran-chaudhary:
> Should we tack on the `csv` format option?

Added.

kiran-chaudhary:
> I am wondering why there is no Table output format as the default, rather than just json objects?

Added, have not defined headers, let me know the fields if needed, I think the "request json" can be done away with.

kiran-chaudhary:
> It appears that the `-b` and `-e` flags do not work. It always sends null in the request when I try this.

This observation might be because the py42 version might be pointing to `timestamp-formats` branch which didn't support float format then, so this must be fixed now.

alanag13:
> This observation might be because the py42 version might be pointing to timestamp-formats branch which didn't support float format then, so this must be fixed now.

Are you saying that this PR has a dependency on https://github.com/code42/py42/pull/205 ?
This PR already also has a dependency on audit log being available in py42 (it is in master but not merged).

I'm going to convert this PR to draft so that we don't risk merging this before the necessary changes in py42 have been released.

kiran-chaudhary:
> > This observation might be because the py42 version might be pointing to timestamp-formats branch which didn't support float format then, so this must be fixed now.
> 
> Are you saying that this PR has a dependency on [code42/py42#205](https://github.com/code42/py42/pull/205) ?
> This PR already also has a dependency on audit log being available in py42 (it is in master but not merged).
> 
> I'm going to convert this PR to draft so that we don't risk merging this before the necessary changes in py42 have been released.

No @alanag13 , I intended to imply that  @unparalleled-js might have checked against some other Py42 branch which didn't support the timestamp format, as I was aware she had just reviewed my other branch.  Since I was not able to reproduce it, I had a discussion with her on the same yesterday and she confirmed it was some error on her side. 

As per my understanding, once review is completed we are only waiting for Py42 version to be bumped so that we can make appropriate change here.

unparalleled-js:
When there are no results, can we print `no results.` like we do for file events?

unparalleled-js:
The field `requestJson` show up in the output and it really bogs it down and feels unnecessary... Not sure what's it for exactly, but I vote we remove it... It is just a JSON string, so it's like JSON within an ascii Table, kind of weird and kind of like nested fields.

unparalleled-js:
> The field `requestJson` show up in the output and it really bogs it down and feels unnecessary... Not sure what's it for exactly, but I vote we remove it... It is just a JSON string, so it's like JSON within an ascii Table, kind of weird and kind of like nested fields.

Maybe we don't need to remove it.... Not a very tight looking table though. Curious what others think.

timabrmsn:
The `requestJson` gives you the actual FFS query the user made for the search event, which is generally going to be an important part of the auditing. It might be possible to pull out a human-readable form of the filter clauses when showing the table view? Like converting this:

```
{"filters":[{"term":"osHostName","operator":"IS","value":"DESKTOP-H88BEKO","display":null},{"term":"osHostName","operator":"IS","value":"W10E-X64-FALLCR","display":null}],"filterClause":"OR","display":null}
```
to:
```
osHostName is ('DESKTOP-H88BEKO' OR 'W10E-X64-FALLCR')
```

Although that's probably too much effort for our given workload right now, and would likely fit better into py42.

alanag13:
my vote would be to have the table format be the same as it is in the ui and only show the `requestJson` in the csv.
The reason for this is that while audit log _currently_ only is used for ffs searches / downloads, many more types of events that will probably _not_ have a relevant `requestJson` property are planned to be added. The fields displayed in the UI currently are those most likely to be relevant for all current and future event types.

We could consider adding a `detail` option (like what we have for alerts) that would output only the extra fields relevant to the specific type of event.

kiran-chaudhary:
Added message when no results are found, Added default header items for table format.

unparalleled-js:
Looks like this can be ready for review now. Py42 released

kiran-chaudhary:
On a day off, this is the most I could do, just toggled the version and made it available for review :)

alanag13:
@unparalleled-js @kiran-chaudhary the version of py42 that was released yesterday does not contain the audit log functionality -- v1.8.2 was a branch off of v1.8.1 that _only_ added a fix for a paging issue that was in py42.

It does look like the other items have completed testing though, so we should be able to release py42 soon. We will need to bump the py42 version of this again once that happens. I'm going to mark this as draft again until it does to ensure that we don't release too early.

setup.py
```diff
@@ -36,7 +36,7 @@
         "c42eventextractor==0.4.0",
         "keyring==18.0.1",
         "keyrings.alt==3.2.0",
-        "py42>=1.8.1",
+        "py42>=1.8.2",
```
unparalleled-js, line 39: 1.9

unparalleled-js:
Looks good to me, just bump py42 vers

integration/test_alerts.py
```diff
@@ -1,89 +1,31 @@
-import json
```
kiran-chaudhary, line 1: Had to fix this, as it used a static date which fails when current date has difference of more than 90 days and removed old sub-commands like `write-to` and `print`. Since the response is dynamic, altered validation to just check the success status of command execution, to make sure at-least the version dependency is verified by real command execution.

src/code42cli/cmds/auditlogs.py
```diff
@@ -0,0 +1,185 @@
...
+        click.echo("No results found.")
+    elif event_count > 10:
+        click.echo_via_pager(formatter.get_formatted_output(events))
+    else:
+        formatter.echo_formatted_list(events)
+
+
+def _send_to(sdk, hostname, protocol, format, **filter_args):
+    logger = get_logger_for_server(hostname, protocol, format)
+    response_handler = create_simple_send_to_handler(
```
timabrmsn, line 182: I don't really see the benefit of pulling this out into a separate function (`create_simple_send_to_handler`), at least not yet since there's only one place this is used. And there's no need to load the response.text to json as py42 already has already done it in the `Py42Response` object. Unless there's something I'm missing `_send_to()` could be simplified to: 

```python
def _send_to(sdk, hostname, protocol, format, **filter_args):
    logger = get_logger_for_server(hostname, protocol, format)
    with warn_interrupt():
        response_gen = sdk.auditlogs.get_all(**filter_args)
        for response in response_gen:
            if EVENT_KEY in response:
                for event in response[EVENT_KEY]:
                    logger.info(event)
            else:
                click.echo("No results found.")
```

timabrmsn (changes requested):

src/code42cli/cmds/auditlogs.py
```diff
@@ -0,0 +1,185 @@
...
+    formatter = OutputFormatter(format, AUDIT_LOGS_DEFAULT_HEADER)
+    response_gen = sdk.auditlogs.get_all(**filter_args)
+
+    events = []
+    try:
+        for response in response_gen:
+            response_dict = json.loads(response.text)
+            if EVENT_KEY in response_dict:
+                events.extend(response_dict.get(EVENT_KEY))
+    except KeyError:
```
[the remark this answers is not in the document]
kiran-chaudhary: this `try..except` block is needed for the generator ( `for response in response_gen` ) which does a dict lookup after the `yield` statement in `get_all_pages` method.

src/code42cli/cmds/auditlogs.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+@click.group(cls=OrderedGroup)
+@sdk_options(hidden=True)
+def audit_logs(state):
+    """Retrieve audit logs."""
+    pass
+
+
+@audit_logs.command()
```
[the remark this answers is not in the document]
kiran-chaudhary: added.

src/code42cli/cmds/auditlogs.py
kiran-chaudhary: I intended to be consistent here with all "extraction/send-to-server" functions in a single module. Refactored, as requested. Are we fine here with `warn_interrupt` not having a message?

src/code42cli/options.py
```diff
@@ -3,9 +3,22 @@
 from code42cli.cmds.search.enums import ServerProtocol
 from code42cli.errors import Code42CLIError
 from code42cli.output_formats import OutputFormat
+from code42cli.output_formats import SendToFileEventsOutputFormat
 from code42cli.profile import get_profile
 from code42cli.sdk_client import create_sdk
 
+BEGIN_OPTION_HEP_MESSAGE = (
```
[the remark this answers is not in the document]
kiran-chaudhary: Changed.

src/code42cli/cmds/auditlogs.py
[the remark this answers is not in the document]
timabrmsn: Ah, I see. I didn't realize the exception would be raised from py42.. Maybe we should add a check in py42 for the missing key and if it's not there insert an empty list to make it consistent with other endpoints? 

We should also file a bug with the audit logs team about that, I'll look into that part.

src/code42cli/cmds/auditlogs.py
timabrmsn: Yeah, but this function doesn't actually use the extractor, so it doesn't quite fit there. 

I'd just rather not try to force abstracting out a function like this until we're sure we'll need that exact functionality in multiple places. 

And warn_interrupt does have a default message (`Cancelling operation cleanly, one moment...`) if none is passed in.

src/code42cli/cmds/auditlogs.py
```diff
@@ -0,0 +1,193 @@
...
+    begin,
+    end,
+    event_type,
+    username,
+    user_id,
+    user_ip,
+    affected_user_id,
+    affected_username,
+):
+    """Send audit logs to given server address."""
```
timabrmsn, line 139: Can we say `to _the_ given server address` to be consistent with the security-data send-to command? And I also noticed when checking to see what the help text was on security-data that `saved-search` is missing help text:

```
Commands:
  clear-checkpoint  Remove the saved file event checkpoint from `--use-checkpoint/-c` mode.
  search            Search for file events.
  saved-search
  send-to           Send events to the given server address.
```

Could you update that too?

kiran-chaudhary: Changed.

src/code42cli/cmds/auditlogs.py
```diff
@@ -0,0 +1,193 @@
...
+EVENT_KEY = "events"
+AUDIT_LOGS_KEYWORD = "audit-logs"
+
+AUDIT_LOGS_DEFAULT_HEADER = OrderedDict()
+AUDIT_LOGS_DEFAULT_HEADER["timestamp"] = "Timestamp"
+AUDIT_LOGS_DEFAULT_HEADER["type"] = "Type"
+AUDIT_LOGS_DEFAULT_HEADER["actorName"] = "ActorName"
+AUDIT_LOGS_DEFAULT_HEADER["actorIpAddress"] = "ActorIpAddress"
+AUDIT_LOGS_DEFAULT_HEADER["success"] = "Success"
+AUDIT_LOGS_DEFAULT_HEADER["resultCount"] = "ResultCount"
```
timabrmsn, line 28: Can we add affected username to the table output list? I think anything that we can filter for in the cmd options should be visible in the table, otherwise you can get stuff like this: 

```
# code42 audit-logs search --affected-username "tim.abramson+support@code42.com" -b 90d
Timestamp                  Type   ActorName   ActorIpAddress   Success   ResultCount   
2020-11-19T03:16:18.239Z   None   None        None             None      None          

# code42 audit-logs search --affected-username "tim.abramson+support@code42.com" -b 90d -f JSON
{
    "type$": "audit_log::local_auth_only_changed/1",
    "timestamp": "2020-11-19T03:16:18.239Z",
    "userId": "942564422882759874",
    "userName": "tim.abramson+support@code42.com",
    "newValue": false
}
```

And I see that there are two "type" fields in the results `type$` and `type`, but we're only showing one in the table, but pretty much all events that are not of type 'query' don't have the `type` field. And from the looks of it, Console is using the `type$` field to pull out the event type to display, so I wonder if all new events will only have `type$` going forward, and if we should just switch to displaying that instead of `type`.

timabrmsn: Yeah, it also looks like when we use `--event-type` to filter, it only works on the `type$` values, so we should definitely switch to displaying that.

src/code42cli/cmds/auditlogs.py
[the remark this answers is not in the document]
timabrmsn: Looks like the bug has already been filed for the empty events result: PL-93211

So I think it's ok to just catch the exception here in the CLI and once the backend is fixed we can revisit if needed.

src/code42cli/cmds/auditlogs.py
kiran-chaudhary: Changed, since we can't accommodate all columns in table format, removed `resultCount` and `success` columns and added the requested one.

timabrmsn (approved):
LGTM

PR #161, opened 2020-09-17
Feature/advanced query from file

timabrmsn:
## Description of change

Implements the ability for `alerts search` and `security-data (search|send-to)` commands to read `--advanced-query` arg from a file. 

If the argument string for `--advanced-query` does not begin with an opening brace `{`, the CLI assumes the argument is a filename, and will try to open and parse the file as JSON. Helpful error messages are printed if the file can't be read or the JSON can't be parsed. 

`--advanced-query` is still incompatible with all other query arguments, but it can now be used with a checkpoint.


.github/workflows/cla.yml
```diff
@@ -0,0 +1,33 @@
+name: "CLA Assistant"
```
kiran-chaudhary, line 1: This file should not be committed in this branch, its already part of another PR!!

CONTRIBUTING.md
```diff
@@ -188,4 +188,8 @@ Document all notable consumer-affecting changes in CHANGELOG.md per principles a
 
 When you're satisfied with your changes, open a PR and fill out the pull request template file. We recommend prefixing the name of your branch and/or PR title with `bugfix`, `chore`, or `feature` to help quickly categorize your change. Your unit tests and other checks will run against all supported python versions when you do this.
 
+For contributions from non-Code42 employees, we require you to agree to our [Contributor License Agreement](https://code42.github.io/code42-cla/Code42_Individual_Contributor_License_Agreement).
```
kiran-chaudhary, line 191: Part of another PR!!!

kiran-chaudhary (changes requested):

.github/workflows/cla.yml
alanag13: @timabrmsn this is usually a sign that your working branch is not caught up with master, could you merge master into this?

src/code42cli/click_ext/groups.py
```diff
@@ -0,0 +1,110 @@
+import difflib
```
alanag13, line 1: are all these `click_ext` files the same as they were before, just moved to another directory? What, if anything, has changed here?

src/code42cli/cmds/search/options.py
```diff
@@ -127,6 +116,16 @@ def handle_parse_result(self, ctx, opts, args):
...
 
 
+def _convert_advanced_query_dict_to_filter_groups(ctx, param, arg):
+    if arg is None:
+        return
+    try:
+        filter_groups = [FilterGroup.from_dict(group) for group in arg["groups"]]
+        return filter_groups
+    except KeyError as e:
+        raise click.ClickException("Failed to validate query from JSON. {}".format(e))
```
alanag13, line 126: `parse` would probably be a better description of what's happening here than `validate`.

src/code42cli/cmds/securitydata.py
```diff
@@ -206,12 +206,12 @@ def clear_checkpoint(state, checkpoint_name):
...
 ):
+    if advanced_query:
+        state.search_filters = advanced_query
     extractor = _get_file_event_extractor(state.sdk, handlers)
     extractor.use_or_query = or_query
     extractor.or_query_exempt_filters.append(f.ExposureType.exists())
-    if advanced_query:
-        extractor.extract_advanced(advanced_query)
-    elif saved_search:
+    if saved_search:
```
alanag13, line 214: use `elif` here.

src/code42cli/click_ext/types.py
```diff
@@ -0,0 +1,39 @@
...
+    If the argument's first non-whitespace character is a "{", assumes it's JSON and tries to parse
+    it, otherwise assumes it's a filename and tries to open file then parse contents as JSON.
+    """
+
+    def __init__(self):
+        super().__init__("r")
+
+    def convert(self, value, param, ctx):
+        if re.match(
+            r"\s*\{", value
```
alanag13, line 20: `curl` does this by treating arguments that have the first character as `@` as files. Could be a good opportunity to borrow a familiar cli pattern.

src/code42cli/click_ext/types.py
```diff
@@ -0,0 +1,39 @@
...
+        ):  # assume JSON string since it begins with a bracket
+            try:
```
alanag13, line 22: these two try/except blocks can merged into a single function that take a string argument (either `value` or file.read()`). It's okay if the exception message is the same for both of those cases, imo.

src/code42cli/click_ext/options.py
```diff
@@ -0,0 +1,36 @@
+import click
```
alanag13, line 1: has anything changed here?

src/code42cli/cmds/alerts.py
```diff
@@ -177,13 +178,12 @@ def _call_extractor(
...
     if advanced_query:
-        extractor.extract_advanced(advanced_query)
-    else:
-        if begin or end:
-            cli_state.search_filters.append(
-                ext.create_time_range_filter(f.DateObserved, begin, end)
-            )
-        extractor.extract(*cli_state.search_filters)
+        cli_state.search_filters = advanced_query
+    if begin or end:
```
alanag13, line 182: I would expect to see some sort of comparable/equivalent change made for file events for this, am I missing it somewhere?

src/code42cli/click_ext/groups.py
timabrmsn: Yeah, just reorganized things a bit. Since I was adding the new `click.File` subclass (`JSONFileOrString`) and it didn't really fit anywhere, I decided to move all the re-usable custom click classes here.

src/code42cli/click_ext/options.py
unparalleled-js: Does not look like this - this was in `options.py` before, if that helps.

src/code42cli/cmds/securitydata.py
timabrmsn: Nope, the `saved-search` uses `cls=incompatible_with("advanced_query")`.

src/code42cli/cmds/alerts.py
timabrmsn: The change for file events is in the `_call_extractor` function, which is used by both `search` and `send-to`.

src/code42cli/cmds/search/options.py
timabrmsn: Well, we've already parsed the json at this point, now we're trying to build the query, so I wanted it to be clear that it's not a json error but that we just couldn't build the query from the data we got. 

Should I just say "failed to build query"?

src/code42cli/click_ext/types.py
timabrmsn: Yeah, at first I was thinking it would be simpler for the user to just pass a bare filename or query and not make it explicit, but I think it is actually probably better to be.

src/code42cli/click_ext/types.py
timabrmsn: I separated the file reading from the json parsing which cleaned this logic up.

src/code42cli/cmds/search/options.py
alanag13: Something like `failed to build query from input JSON` seems to fit that, yeah.

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 ### Changed
 
+- The `--advanced-query` option on `alerts search` and `security-data (search|send-to)` commands has been updated:
```
alanag13, line 15: we should call out the `@` prefix here as well as in the help documentation string.

timabrmsn: done

alanag13 (approved):

alanag13:
@kiran-chaudhary when you get a chance can you re-review/approve this, assuming that the changes you requested have been addressed?

As an FYI, if you request changes, this makes it so that the change cannot be merged until you approve (even if someone else approves).

kiran-chaudhary (approved):
My bad, apologies. LGTM. I have not tested it, the work has some conflicts with my changes so once merged I will test it along with my work.

Issue #166, opened 2020-10-13
[Enhancement] End support for python 3.5


alanag13:
## Summary

Python 3.5 reached official end-of-life in September. Ending support for it allows us to take advantage of new language features and enables us to take on modern dependencies that could be useful to the cli such as `pandas`.

A large part of the migration effort for this can probably be greatly simplified by changing our `pyupgrade` style hook to be 3.6+.

timabrmsn:
This was implemented in #167.

PR #167, opened 2020-10-13
Add devices command

### LINKED ISSUE #165: [Enhancement] Deactivate Devices command
## Summary
One of the most used scripts currently maintained by tech services is the device deactivations script. This script can be used either as scheduled task or for one-time cleanup efforts. It supports a variety of settings to customize the functionality.

## Proposed API

The existing standalone script has a lot of functionality. In conversation with Paul Hirst, we separated it out into the important things that should be in the initial implementation, and then a set of things that would be nice-to-have, but aren't necessarily essential.

Important/first-pass:
* Dry-run by default - require a parameter to actually deactivate anything
* GUID list - no validation against anything else
* Last connected date
* Update cold storage
* Rename devices, customizable prefix, date appended - check to make sure the device doesn't already have `_deactivated` in it
* Org scope (included/excluded, option to include child orgs)
* option to leave users with a minimum count of devices (i.e. don't deactivate a device if it's a user's last device) - save most recently connected devices
* Outputs results/summary file (see below)

Nice-to-have:
* Min/max archive size
* Client version
* OS type
* Don't try to deactivate legal hold devices to save time (only if user can see legal hold)
* Settings can be saved to and read from a file - don't worry about compat

Output - summary file:
* CSV of all devices and status (deactivated, failed, not deactivated b/c legal hold) - include owning user
* How many devices (total)
* How much data recovered (total)
* Any errors - details?
* Record of settings used

Some proposed APIs:
```
code42cli devices bulk deactivate --filename devicesToDeactivate.csv --execute
code42cli devices bulk deactivate --daysSinceActive 90 --maximumArchiveSize 100GB --versions 6.8.9,7.0.5 --minimumDeviceCount 1 --coldStorageDays 1 --org 3099 --includeChildOrgs --execute
code42cli devices bulk deactivate --settingsFile settings.txt --execute
```

## Intended Use Case
Environment cleanup and maintenance.

ceciliastevens:
**Description of Change**

This PR adds and tests the `devices` command, along with `devices deactivate`, `devices show`, and the bulk equivalents `devices bulk deactivate` and `devices list`.

Closes #165 and #171

This will also require adding pandas as a dependency, which isn't supported on python 3.5, so this should not be merged until support for 3.5 is ended.

**Testing Procedure**

```
code42 devices show 12345
code42 devices deactivate 12345 --change-device-name --purge-date 2020-10-13
code42 devices list --active --org-uid 123456789 --drop-most-recent 1 --include-backup-usage --include-settings
code42 devices bulk deactivate test_device_id_list.csv --change-device-name --purge-date 2020-10-13
```

**PR Checklist**

* [x]  Add unit tests to verify this change
* [x]  Add an entry to CHANGELOG.md describing this change
* [x]   Add docstrings for any new public parameters / methods / classes


alanag13:
This PR removes support for python 3.5. Can you please call that out in the description somewhere?

If we end support for 3.5, it will probably be as a separate PR and we wouldn't merge this until that. I'm still waiting to hear back from SST members about that, apparently there is some disagreement about how widespread 3.5 use is / potential impact of this.

ceciliastevens:
Ah thank you for reminding me to mention that. This will need to wait on the end of support for 3.5, and I'm still planning on adding some more things to this PR so that we have something approaching feature parity with the old standalone script, so I don't anticipate that this'll be ready for merging anytime soon.

ceciliastevens:
With the addition of the username inclusion I think we are more or less feature-complete for an initial implementation. I have a few items left to address however:

1. Is it possible to adjust the csv input file option to allow a CSV that contains extra columns? If I'm reading the built-in stuff correctly it will enforce that the csv only contains one column, whereas I would like it to work with csvs that contain other columns (working around this is pretty trivial with something like awk but it it'd be nice if we could support a `code42 devices bulk info | code42 devices bulk deactivate` workflow)
2. Currently any error reporting requires cross-referencing your input CSV with the errors logged to the CLI's log. Ideally we'd be able to output a results spreadsheet or print errors to the console as they occur or something like that. I'm not super familiar with how Click's error handling interacts with the way we're doing run_bulk_process, or what the best way is to have my row handler method (or something) catch and handle errors.

alanag13:
@ceciliastevens we definitely have functionality that uses csv inputs with more than one column, for example bulk adding to departing employee works this way (you can specify both the user and the departure date as part of the input). I would think that as long as the csv _contains_ all the necessary columns , you should be good (it shouldn't matter if you have _extra_ columns).

Printing a progress bar to the terminal _and_ error output to the screen at the same time during output probably is a non-starter, but I like the idea of a results csv. We could probably change the way that bulk processing works to optionally take a file name that would write the currently processing row over again, with additional column of "successful" or something to that effect.

ceciliastevens:
In order to allow the output of a results spreadsheet, this PR now also contains some changes to run_bulk_process and worker. This shouldn't impact any of their existing functionality, so existing consumers shouldn't be affected.

Additionally, I am going to work on implementing a device settings report (Issue #171), and that will likely re-use a lot of the functionality here. Would it be best to create a separate PR (with the understanding that a lot will be duplicated), or should I instead just add that work to this existing PR and update the name/description to indicate that it addresses both features?

ceciliastevens:
Because I already had `devices bulk info` in here, and I would've had to duplicate almost all of that work, I decided to address #171 here instead of in a separate PR

alanag13:
Thanks @ceciliastevens, that works!

src/code42cli/cmds/devices.py
```diff
@@ -0,0 +1,368 @@
...
+            )
+            row["deactivated"] = "True"
+        except Exception as e:
+            row["deactivated"] = "False: {}".format(e)
+        return row
+
+    result_rows = run_bulk_process(
+        handle_row, csv_rows, progress_label="Deactivating devices:"
+    )
+    click.echo("Run complete.")
```
kiran-chaudhary, line 367: Is this needed, intentionally printing the statement?

tox.ini
```diff
@@ -1,6 +1,6 @@
 [tox]
 envlist =
-    py{38,37,36,35,27}
```
kiran-chaudhary, line 3: Do we not intend to support the removed versions?

src/code42cli/file_readers.py
```diff
@@ -20,19 +20,28 @@ def read_csv(file, headers=None):
     """Helper to read a csv file object into dict rows, automatically removing header row
     if it exists, and errors if column count doesn't match header list length.
     """
-    reader = csv.DictReader(file, fieldnames=headers)
+    reader = csv.DictReader(file)
     first_row = next(reader)
-    if None in first_row or None in first_row.values():
```
kiran-chaudhary, line 25: Would like to understand what has changed here, does it addresses a special case? if so, a unit test would suffice to explain the same.

src/code42cli/cmds/devices.py
```diff
@@ -0,0 +1,368 @@
...
+from code42cli.file_readers import read_csv_arg
+from code42cli.options import format_option
+from code42cli.options import sdk_options
+from code42cli.output_formats import OutputFormatter
+
+
+@click.group(cls=OrderedGroup)
+@sdk_options(hidden=True)
+def devices(state):
+    """For managing computers within Code42."""
```
kiran-chaudhary, line 23: Is `within Code42` more apt or `within an organisation`? Similarly in all places elsewhere.

ceciliastevens: The commands don't require an organization UID and will default to showing all devices within your Code42 environment (i.e. all that are visible to your user). For that reason I didn't specify "organization" within the docstrings.

I could update it to say "organization", or if we want to be exactingly correct we could say "within a Code42 tenant" or something like that.

src/code42cli/file_readers.py
ceciliastevens: Previously, the DictReader required that the CSV have a precise number of columns, and that the columns be in the correct order, but it did not require column headers. This change allows for files formatted in that way, but also allows for CSVs with an arbitrary number and order of columns as long as the correct headers are included.

I'll add some tests to clarify.

tox.ini
ceciliastevens: This PR is dependent on the end of support for python 3.5 and 2.7, as neither of those versions are supported by Pandas, which we require here. For that reason, we can't merge this PR until issue #166 is resolved.

src/code42cli/cmds/devices.py
ceciliastevens: Nope, removed.

src/code42cli/file_readers.py
ceciliastevens: I've added test_file_readers.py and two tests to cover the CSV cases this change supports.

src/code42cli/cmds/devices.py
unparalleled-js: I'd still say `For managing devices within Code42` because no other part of our docs in py42 or the CLI refer to them as `computers`.

src/code42cli/cmds/devices.py
```diff
@@ -0,0 +1,399 @@
...
+    help="The date on which the archive should be purged from cold storage in yyyy-MM-dd format. If not provided, the date will be set according to the appropriate org settings.",
+)
+
+
+@devices.command()
+@device_id_option
+@change_device_name_option
+@purge_date_option
+@sdk_options()
+def deactivate(state, device_id, change_device_name, purge_date):
```
unparalleled-js, line 56: `device_id` should probably be an `click.Argument` rather than a `click.option` since it is always required and there are no other required args, that is the pattern we have been doing anyway.

src/code42cli/cmds/devices.py
```diff
@@ -0,0 +1,399 @@
...
+
+
+@click.group(cls=OrderedGroup)
+@sdk_options(hidden=True)
+def devices(state):
+    """For managing computers within Code42."""
+    pass
+
+
+device_id_option = click.option(
```
unparalleled-js, line 27: This needs to call out that is a legacy ID..

It is unfortunate though that we have to use legacy IDs.

I am wondering what thoughts are for making these be the normal device GUIDs and just getting the legacy ID behind the scenes.

src/code42cli/cmds/devices.py
```diff
@@ -0,0 +1,399 @@
...
+        sdk.devices.deactivate(device_id)
+    except exceptions.Py42BadRequestError:
+        raise Code42CLIError("The device {} is in legal hold.".format(device_id))
+    except exceptions.Py42NotFoundError:
+        raise Code42CLIError("The device {} was not found.".format(device_id))
+    except exceptions.Py42ForbiddenError:
+        raise Code42CLIError("Unable to deactivate {}.".format(device_id))
+    if purge_date or change_device_name:
+        device = sdk.devices.get_by_id(device_id)
+        guid = device.data["guid"]
```
unparalleled-js, line 72: It looks like we conditionally make a call to get the guid..

I am convinced that we should take `guid` as the argument and get the `legacy id` behind the scenes and not expose it to the end users anymore

tests/cmds/test_devices.py
```diff
@@ -0,0 +1,478 @@
...
+            TEST_PURGE_DATE,
+        ],
+        obj=cli_state,
+    )
+    cli_state.sdk.archive.update_cold_storage_purge_date.assert_called_once_with(
+        TEST_ARCHIVE_GUID, TEST_PURGE_DATE
+    )
+
+
+def test_deactivate_changes_device_name(
```
unparalleled-js, line 318: This test makes it sound like that is expected every time.

Would be better to call it `test_deactivate_when_given_flag_changes_device_name()`

unparalleled-js, line 318: There should be a test that also makes sure that it _doesn't_ change the name when the name is already prefixed with `deactivated_` because I noticed that conditional in the logic. (sorry if there is one and I just am not seeing it)

CHANGELOG.md
```diff
@@ -8,14 +8,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
-## Unreleased
-
 ### Added
 
 - `code42 audit-logs` commands:
     - `search` to search for audit-logs.
     - `send-to` to send audit-logs to server.
 
+## Unreleased
```
unparalleled-js, line 17: unreleased hoist

src/code42cli/cmds/devices.py
```diff
@@ -0,0 +1,399 @@
...
+        sdk.archive.update_cold_storage_purge_date(archive_guid, purge_date)
+
+
+def _change_device_name(sdk, guid, name):
+    device_settings = sdk.devices.get_settings(guid)
+    device_settings.name = name
+    sdk.devices.update_settings(device_settings)
+
+
+_DEVICE_INFO_KEYS_MAP = OrderedDict()
```
unparalleled-js, line 99: If this is only used in a single function, it would make sense to localize there, either by moving this until the command logic or wrapping it a helper function and calling it within the command

src/code42cli/cmds/devices.py
```diff
@@ -0,0 +1,399 @@
...
+
+@devices.command()
+@device_id_option
+@format_option
+@sdk_options()
+def show(state, device_id, format=None):
+    """Print device info."""
+    formatter = OutputFormatter(format, _DEVICE_INFO_KEYS_MAP)
+    device_info = _get_device_info(state.sdk, device_id)
+    if device_info:
```
unparalleled-js, line 122: What is the situation when device_info is falsy?

src/code42cli/cmds/devices.py
```diff
@@ -0,0 +1,399 @@
...
+        if len(device["backupUsage"]) > 0
+        else None
+    )
+    return device
+
+
+@devices.group(cls=OrderedGroup)
+@sdk_options(hidden=True)
+def bulk(state):
+    """Tools for managing devices in bulk"""
```
unparalleled-js, line 164: Missing period

src/code42cli/cmds/devices.py
```diff
@@ -0,0 +1,399 @@
...
+    pass
+
+
+@bulk.command(name="list", help="Get information about many devices")
```
unparalleled-js, line 168: Missing period in the help text

src/code42cli/cmds/devices.py
```diff
@@ -0,0 +1,399 @@
...
+@click.option(
+    "--active",
+    required=False,
+    type=bool,
+    is_flag=True,
+    default=None,
+    help="Include to return only active devices.",
```
unparalleled-js, line 175: This should elaborate slightly more like `Include to return only active or in active devices` and should mention the default in the help text as well (see `code42 alerts search -h` as an example.

So in full, like `Include to return only active or in active devices. Defaults to getting all the devices.`

src/code42cli/cmds/devices.py
```diff
@@ -0,0 +1,399 @@
...
+    required=False,
+    type=str,
+    default=None,
+    help="Optionally provide to limit devices to only the ones in the org you specify. Note that child orgs will be included.",
+)
+@click.option(
+    "--drop-most-recent",
+    required=False,
+    type=int,
+    help="Will drop the X most recently connected devices for each user from the result list where X is the number you provide as this argument. Can be used to avoid passing the most recently connected device for a user to the deactivate command",
```
unparalleled-js, line 194: I know our formatters might not be bothered by these longer help texts, but if you manually hit `Enter` for it to be multiline, our formatter with automatically adjust it but still keep it multi-line, if that makes sense. It would be nice!  This feedback applies to all new help texts for click options in this PR

unparalleled-js: also missing period on this one

src/code42cli/cmds/devices.py
```diff
@@ -0,0 +1,399 @@
...
+)
+@click.option(
+    "--include-backup-usage",
+    required=False,
+    type=bool,
+    default=False,
+    is_flag=True,
+    help="Include to return backup usage information for each device (may significantly lengthen the size of the return)",
```
unparalleled-js, line 202: missing period

unparalleled-js: I'd audit the rest of these

src/code42cli/cmds/devices.py
ceciliastevens: The underlying method in py42 uses legacy ID, as does the API it calls. I agree that it would be best if we could use GUID, but is it best to have that translation live here? Or should it be done in py42?

src/code42cli/cmds/devices.py
```diff
@@ -0,0 +1,399 @@
...
+@click.option(
+    "--include-settings",
+    required=False,
+    type=bool,
+    default=False,
+    is_flag=True,
+    help="Include to include device settings in output",
+)
+@sdk_options()
+def bulk_list(
```
unparalleled-js, line 221: Should be this under the `bulk` command group? I realize it uses the bulk processor, but would users know they need to search under bulk to find the list command?

It does not like this like command takes bulk data but rather it is just async, so it should perhaps just be `code42 devices list`

src/code42cli/cmds/devices.py
```diff
@@ -0,0 +1,399 @@
...
+        devices_dataframe = _drop_devices_which_have_not_connected_in_some_number_of_days(
+            devices_dataframe, days_since_last_connected
+        )
+    if include_settings:
+        devices_dataframe = _add_settings_to_dataframe(state.sdk, devices_dataframe)
+    if include_usernames:
+        devices_dataframe = _add_usernames_to_device_dataframe(
+            state.sdk, devices_dataframe
+        )
+    click.echo(devices_dataframe.to_csv())
```
unparalleled-js, line 249: When output is large (greater than 10), we have been using `click.echo_via_pager()`. Would that make sense here?

tests/cmds/test_devices.py
```diff
@@ -0,0 +1,478 @@
...
+                "storePointId": 1000,
+                "selectedBytes": 1658317953,
+                "selectedFiles": 596,
+                "todoBytes": 0,
+                "format": "ARCHIVE_V1",
+            }
+        ]
+    }
+]
+TEST_DEVICE_RESPONSE = """{"data":{"computerId":139527,"name":"testname","osHostname":"testhostname","guid":"954143368874689941","type":"COMPUTER","status":"Active","active":true,"blocked":false,"alertState":0,"alertStates":["OK"],"userId":203988,"userUid":"938960273869958201","orgId":3099,"orgUid":"915323705751579872","computerExtRef":null,"notes":null,"parentComputerId":null,"parentComputerGuid":null,"lastConnected":"2020-10-12T16:55:40.632Z","osName":"win","osVersion":"10.0.18362","osArch":"amd64","address":"172.16.208.140:4242","remoteAddress":"72.50.201.186","javaVersion":"11.0.4","modelInfo":null,"timeZone":"America/Chicago","version":1525200006822,"productVersion":"8.2.2","buildVersion":26,"creationDate":"2020-05-14T13:03:20.302Z","modificationDate":"2020-10-12T16:55:40.632Z","loginDate":"2020-10-12T12:54:45.132Z","service":"CrashPlan"}}"""
```
unparalleled-js, line 49: Like a similar comment elsewhere, if you manually make this super long lines multiline, our formatter make it look the way it wants and still keep it multiline... Some of these are quite long!

tests/cmds/test_devices.py
```diff
@@ -0,0 +1,478 @@
...
+
+
+def test_deactivate_deactivates_device(runner, cli_state, deactivate_device_success):
+    runner.invoke(
+        cli, ["devices", "deactivate", "--device-id", TEST_DEVICE_ID], obj=cli_state
+    )
+    cli_state.sdk.devices.deactivate.assert_called_once_with(TEST_DEVICE_ID)
+
+
+def test_deactivate_updates_purge_date(
```
unparalleled-js, line 294: Like the test below, should have the state as part of the signature, like `test_deactivate_when_given_purge_flag_purges_date()` or else it sounds like a default behavior test

tests/cmds/test_devices.py
unparalleled-js: also something worth auditing the rest of the tests for

src/code42cli/cmds/devices.py
unparalleled-js: Is this because the bulk stats are helpful?  Does this command take a while?  Even if that is the case, it still might make sense to not this not be in the bulk bucket... I am curious what others think as well

tox.ini
unparalleled-js: There is a classifier in `setup.py` also need to be removed, line 56

unparalleled-js: Also line 17 in `.github/workflows/build.yml`

setup.py
```diff
@@ -36,6 +36,7 @@
         "c42eventextractor==0.4.0",
         "keyring==18.0.1",
         "keyrings.alt==3.2.0",
+        "pandas>=1.1.3",
         "py42>=1.9",
     ],
     extras_require={
```
unparalleled-js, line 42: python 3.5 classifier at bottom of this file should be removed

src/code42cli/cmds/devices.py
ceciliastevens: I've updated this text to use 'devices' and to say "within your Code42 environment", which is more specific than "within Code42" while still accurately reflecting that it is not limited to a single organization

src/code42cli/cmds/devices.py
ceciliastevens: I've updated this to an argument.

CHANGELOG.md
ceciliastevens: I think this got a bit weird because at some point I merged in additional changes from the main branch. How should I organize things here? Should the new changes be in "added" or "unreleased"?

src/code42cli/cmds/devices.py
ceciliastevens: Moved it into the function.

src/code42cli/cmds/devices.py
ceciliastevens: I was thinking it might be if we didn't get device info, but we should still have the response even if it's empty. Removed this conditional.

src/code42cli/cmds/devices.py
ceciliastevens: Yes, the output could be quite long. Updated.

src/code42cli/cmds/devices.py
ceciliastevens: I put it under the bulk command group because it deals with devices in bulk. My thought was that we'd have an organization of commands that deal with an individual device (show, deactivate, etc), and then a bulk command for operating on many devices at once (bulk list, bulk deactivate, etc.).

I could move the list command out of the bulk group, but I worry that it might be confusing as it's meant to be used in combination with the other bulk command.

ceciliastevens: The use of run_bulk_process is only for adding device settings, which require an API call per device (i.e. they cannot be retrieved from sdk.devices.get_all()).

tox.ini
ceciliastevens: Updated setup.py and build.yml

src/code42cli/cmds/devices.py
unparalleled-js: It would be a breaking change in py42, so maybe it is a discussion for py42 2.0, unless the methods support both. I think my favorite option is do this in the CLI for now.

src/code42cli/cmds/devices.py
unparalleled-js: I see where you are coming from. Listing the devices and showing a device are both top-level commands in the context of this application though, like these others:

`code42 legal-hold list`
`code42 departing-employee list`
`code42 profile list`

and 

`code42 legal-hold show`
`code42 profile show`

etc.

And bulk operations specifically refer to bulk processing a data set, like any WRITE command on a CSV.

I also see listing devices as having other use cases outside of being used with the other bulk command in this PR.

CHANGELOG.md
unparalleled-js: At the moment, you would just need to delete the unreleased and put back the one that was deleted at the top.

However, we plan on cutting a release of audit logs before merging anything else, soon, so you could probably skip the CHANGELOG until the end

src/code42cli/cmds/devices.py
```diff
@@ -0,0 +1,398 @@
...
+
+
+device_id_argument = click.argument("device-id", type=str)
+
+change_device_name_option = click.option(
+    "--change-device-name",
+    required=False,
+    is_flag=True,
+    default=False,
+    help="""Include to prepend "deactivated_" and today\'s date to the name of any
```
unparalleled-js, line 34: You don't need the `\'` anymore, I don't think it will escape when using the """

src/code42cli/cmds/devices.py
ceciliastevens: Makes sense. I'll move `list` out of the `bulk` group.

src/code42cli/cmds/devices.py
```diff
@@ -0,0 +1,399 @@
...
+        max(
+            [
+                backupDestination["lastCompletedBackup"]
+                for backupDestination in device["backupUsage"]
+            ]
+        )
+        if len(device["backupUsage"]) > 0
+        else None
+    )
+    return device
```
timabrmsn, line 153: A couple things here: 

1. As-is this is pretty hard to read and understand what's happening. Because we're checking `if len(device["backupUsage"]) > 0` in all 3 of these values, it would clarify things to break that out separately to make understanding how we're doing the calculations for values that do exist much easier:

```
    if not device["backupUsage"]:
        device["archiveBytes"] = 0
        device["lastBackup"] = None
        device["lastCompletedBackup"] = None
```
 
2. This currently fails if `backupDestination["lastCompletedBackup"]` is null/None as `max()` can't compare it. We'll need to filter out Nones, and then also handle an empty list.
3. Are we sure we want to reduce all these to single values across separate destinations? Imagine a device that has a really old 1TB archive on destination A that it hasn't backed up to in a year (for whatever reason), there is also destination B that was recently added, but the device has only backed up 1GB to it (maybe it's a different backup set with smaller selection), the device report would show this device as 1TB archive bytes with a very recent backup completed date.

ceciliastevens: I'm updating it to handle null and empty cases more gracefully.

Regarding point 3, my goal with these processed values is to show a summary, the full detail (including the full backupUsage object) is available when using CSV or JSON formatting. 

When attempting to show information for all destinations, we run into trouble with the CSV format because the default output for a list is `[item1,item2,item3]`, which will behave oddly if they are all supposed to be in a single column (which they are, in this case). This could also present a problem for the CSV produced by `devices list` as well.

I think we'll need a helper method to format lists in a CSV-friendly manner (i.e. "item1 | item2 | item3" or something)

src/code42cli/cmds/devices.py
unparalleled-js: I agree that the bulk command should continue to use the legacy Id. We make sure to should call that out.

ceciliastevens: Updated `devices show` and `devices deactivate` to use GUID. Clarified the docstring in `devices bulk deactivate` that it does not use GUID, and instead uses legacy deviceId.

src/code42cli/cmds/devices.py
```diff
@@ -0,0 +1,396 @@
...
+        devices_dataframe = _drop_devices_which_have_not_connected_in_some_number_of_days(
+            devices_dataframe, days_since_last_connected
+        )
+    if include_settings:
+        devices_dataframe = _add_settings_to_dataframe(state.sdk, devices_dataframe)
+    if include_usernames:
+        devices_dataframe = _add_usernames_to_device_dataframe(
+            state.sdk, devices_dataframe
+        )
+    click.echo_via_pager(devices_dataframe.to_csv())
```
timabrmsn, line 231: Our other `list` commands have a convention of multiple format options (with the default being a human-readable table). 

But.. This has me thinking. DataFrames have a really convenient `.to_string()` method that basically does the equivalent to our hand-rolled table output formatter. 

We could adopt the use of dataframes for all of our commands that need to output tabular results and simplify our codebase a ton (in addition to enabling other features across the board).

src/code42cli/cmds/devices.py
timabrmsn: Since the `show` command only handles one device at a time, I think it would make sense to show as much detail for it as possible. Could we just have each destination be it's own row so the user can get the full picture?

src/code42cli/cmds/devices.py
```diff
@@ -0,0 +1,399 @@
...
+    required=False,
+    type=int,
+    help="Return only devices that have not connected in the number of days specified.",
+)
+@click.option(
+    "--org-uid",
+    required=False,
+    type=str,
+    default=None,
+    help="Optionally provide to limit devices to only the ones in the org you specify. Note that child orgs will be included.",
```
timabrmsn: In all these help texts, we can skip words like "optionally provide" or "include to" and just state what the result of using the option is. e.g.:

`Limit devices to only the ones in the specified org-uid. Note that child orgs will be included.`

src/code42cli/cmds/devices.py
unparalleled-js: That sounds great! Should we get this PR through first and then refactor? We can do it as a quick follow to this.

ceciliastevens: I've added a formatter for dataframes, and changed the `list` method to use it. There isn't much there but it can use the same format option as the other methods that accept different formats.

src/code42cli/cmds/devices.py
ceciliastevens: I've updated show to output a second table with backup set information. This allows us to pull out this code entirely, since we no longer need to do any processing of the backup set information, and replace it with an output formatter.

src/code42cli/cmds/devices.py
```diff
@@ -0,0 +1,383 @@
...
+            )
+            row["deactivated"] = "True"
+        except Exception as e:
+            row["deactivated"] = "False: {}".format(e)
+        return row
+
+    result_rows = run_bulk_process(
+        handle_row, csv_rows, progress_label="Deactivating devices:"
+    )
+    formatter.echo_formatted_list(result_rows)
```
unparalleled-js, line 383: Cool, I think this is the first bulk command that outputs data after the process completes.. Is there any other intended purpose for this besides giving the user feedback? I am curious how this looks next to the progress bar and stats

unparalleled-js: We could potentially add an option to silence the progress bar so the data is pipeable, if it needs to be?

src/code42cli/output_formats.py
```diff
@@ -76,6 +77,30 @@ def _requires_list_output(self):
         return self.output_format in (OutputFormat.TABLE, OutputFormat.CSV)
 
 
+class DataFrameOutputFormatter:
+    def __init__(self, output_format):
```
unparalleled-js, line 81: Wow! Thanks for doing this

unparalleled-js: I am currently facing a bug with CSV windows output formatter and now I am wondering if this fixes it...https://stackoverflow.com/questions/3348460/csv-file-written-with-python-has-blank-lines-between-each-row

src/code42cli/cmds/devices.py
ceciliastevens: I think it already plays nice with piping as-is. I just tested with piping into a file, and the file only had the end results, not the progress bar.

unparalleled-js:
Is this ready? I notice it is still a draft PR. @timabrmsn ?

ceciliastevens:
> Is this ready? I notice it is still a draft PR. Are you also good, @timabrmsn ? Edit: besides the conflits

per the discussion with Alan back in October, I've left this as a draft until we're ready to remove 3.5 support (i.e. we're waiting on #166. 

If y'all are ready to go ahead, I can mark it as ready and resolve the conflicts.

timabrmsn:
Feel free to resolve the conflicts and move the PR out of draft, but I'm not quite done with reviewing yet.

ceciliastevens:
Ok, I've resolved the conflicts and marked as ready.

src/code42cli/cmds/devices.py
```diff
@@ -0,0 +1,383 @@
...
+
+def _update_cold_storage_purge_date(sdk, guid, purge_date):
+    archives_response = sdk.archive.get_all_by_device_guid(guid)
+    archive_guid_list = [
+        archive["archiveGuid"]
+        for page in archives_response
+        for archive in page["archives"]
+    ]
+    for archive_guid in archive_guid_list:
+        sdk.archive.update_cold_storage_purge_date(archive_guid, purge_date)
```
timabrmsn, line 91: This fails on v2 (security plan archives) since if the user is still active v2 archives will stay active and not have a purge date. We should filter those where `'format': 'ARCHIVE_V2'`.

src/code42cli/cmds/devices.py
```diff
@@ -0,0 +1,383 @@
...
+        sdk.devices.deactivate(device.data["computerId"])
+    except exceptions.Py42BadRequestError:
+        raise Code42CLIError("The device {} is in legal hold.".format(device_guid))
+    except exceptions.Py42NotFoundError:
+        raise Code42CLIError("The device {} was not found.".format(device_guid))
+    except exceptions.Py42ForbiddenError:
+        raise Code42CLIError("Unable to deactivate {}.".format(device_guid))
+    if purge_date or change_device_name:
+        guid = device_guid
+        name = device.data["name"]
```
timabrmsn, line 72: This seems unnecessary, at least the `if purge_date: guid = device_guid` portion.

src/code42cli/cmds/devices.py
```diff
@@ -0,0 +1,383 @@
...
+    _DEVICE_INFO_KEYS_MAP["productVersion"] = "Code42 Version"
+    _DEVICE_INFO_KEYS_MAP["osName"] = "Operating System"
+    _DEVICE_INFO_KEYS_MAP["osVersion"] = "Operating System Version"
+
+    _BACKUP_SET_KEYS_MAP = OrderedDict()
+    _BACKUP_SET_KEYS_MAP["targetComputerName"] = "Destination"
+    _BACKUP_SET_KEYS_MAP["lastBackup"] = "Last Backup Activity"
+    _BACKUP_SET_KEYS_MAP["lastCompleted"] = "Last Completed Backup"
+    _BACKUP_SET_KEYS_MAP["archiveBytes"] = "Archive Size in Bytes"
+    _BACKUP_SET_KEYS_MAP["archiveGuid"] = "Archive GUID"
```
timabrmsn, line 121: Since we're dropping 3.5, we don't need to use OrderedDict here, since dicts are now ordered by default in 3.6+

And these should maybe be declared outside the function so they can be used elsewhere (tests, etc?) if needed in the future.

src/code42cli/output_formats.py
```diff
@@ -76,6 +77,30 @@ def _requires_list_output(self):
...
+        output_format = output_format.upper() if output_format else OutputFormat.TABLE
+        self.output_format = output_format
+        self._format_func = DataFrame.to_string
+        self._output_args = {}
```
timabrmsn, line 85: We should add `index=False` to all these output args so we're not printing/adding the dataframe index values.

src/code42cli/cmds/devices.py
```diff
@@ -0,0 +1,383 @@
...
+        current_result_dict = {
+            "guid": current_device_settings.guid,
+            "included_files": [
+                backup_set.included_files
+                for backup_set in current_device_settings.backup_sets
+            ],
+            "excluded_files": [
+                backup_set.excluded_files
+                for backup_set in current_device_settings.backup_sets
+            ],
```
timabrmsn, line 283: Since `.included_files` and `.excluded_files` return lists, these comprehensions end up looking like `[[:user]]` for a single backup set or `[[:user],[:user]]` for multiple backup sets. We could flatten to a single list and remove duplicates by using a set:

```
included_files = set()
for bs in current_device_settings.backup_sets:
    for path in bs.included_files:
        included_files.add(path)

current_result_dict = {
             "guid": current_device_settings.guid,
              "included_files": list(included_files)
...
```

timabrmsn (changes requested):

src/code42cli/cmds/devices.py
ceciliastevens: I originally had them outside the function, and moved them in at @unparalleled-js 's suggestion, since they're not currently used anywhere else, so we don't need to create them unless this particular method is called.

I could move them back out now, but it'd be pretty simple to move them later when/if they end up being re-used. What do you think?

src/code42cli/cmds/devices.py
ceciliastevens: This is what I had here originally, but I changed it to a list of lists to better represent the case where there are multiple backup sets. For example, consider the case (of one of our customers) where there is one backup set that includes the entire user folder but excludes certain specific folders, and then a second set that includes specifically only those folders excluded by the first set. So you end up with a structure like this:

```
set 1:
    included:
         :user
    excluded:
         :user/examplefolder

set 2:
     included:
         :user/examplefolder
     excluded:
         :user/examplefolder/examplefolder2
```

If we use a set, we end up with:

```
included: :user, :user/examplefolder
excluded: :user/examplefolder, :user/examplefolder/examplefolder2
```

whereas with a list of lists, the structure is more clear:

```
included: [:user], [:user/examplefolder]
excluded: [:user/examplefolder], [:user/examplefolder/examplefolder2]
```

So in order to handle this case, I think that a list of lists is the clearest option (it is also what the current standalone script does).

Another option would be to have one row per backup set instead of one row per device, but that can also be confusing and make it more difficult to get accurate counts/filtering on the result spreadsheet (which I believe is why the standalone script also opts for a single line per device, and a list-of-lists representation for multiple backup sets).

src/code42cli/cmds/devices.py
unparalleled-js: I also suggested that we could put them in a helper method.. That way, they can still be used in tests if needed AND that makes them lazily loaded which is better imo

src/code42cli/cmds/devices.py
```diff
@@ -103,7 +100,7 @@ def _change_device_name(sdk, guid, name):
 @sdk_options()
 def show(state, device_guid, format=None):
     """Print device info. Requires device GUID."""
-    _DEVICE_INFO_KEYS_MAP = OrderedDict()
+    _DEVICE_INFO_KEYS_MAP = dict()
```
unparalleled-js, line 109: We also don't need this syntax, we could just initialize it like `{"name": "Name"...  }`

src/code42cli/cmds/devices.py
```diff
@@ -103,7 +100,7 @@ def _change_device_name(sdk, guid, name):
 @sdk_options()
 def show(state, device_guid, format=None):
     """Print device info. Requires device GUID."""
-    _DEVICE_INFO_KEYS_MAP = OrderedDict()
```
unparalleled-js: If we keep this in a method (like a helper method), the variable name should not be all-caps

src/code42cli/cmds/devices.py
ceciliastevens: moved them to helper methods.

src/code42cli/cmds/devices.py
timabrmsn: I'm wondering if we should split these up into two sub-commands, something like `list` and `list-backups`, where `list` just shows a single-row per device with high level status, and `list-backups` breaks it down to a row per backup set, and has _all_ the detail of each set. 

That way we don't have to worry about cramming just the right amount of information into a single row in a way that makes sense, and can leave it up to the end-user to use which output format they need depending on the use-case.

ceciliastevens: Currently there's so little that we're doing in --include-settings that is not related to backup sets (only the full disk access status), I feel like it's fine to just have --include-settings have one line per backup set, rather than splitting everything out. I've updated it to do one line per backup set.

timabrmsn: I think as it is there's already a ton of information in this output, and if we're duplicating guid rows for each backup set, and also including the "regular" settings in each row, it makes it quite messy (if one device has a huge include/exclude list it can make the table _really_ wide). 

I'd prefer to separate them and have two tighter, easier-to-read outputs, which enables us to put more data in the `list-backup-config` output if we want (we're missing backup set names right now, which I'd also want to put in).

ceciliastevens: That makes sense. I've refactored it to have a `devices list` and `devices list-backup-sets` as separate commands. Is this more what you had in mind?

timabrmsn: Yeah, I think that's a lot better. Since `list --include-settings` now only adds the full-disk-access state, should we separate that out into a separate function since we don't need to call `.get_settings()` on every guid in `list`?

ceciliastevens: I will definitely remove that extraneous get_settings call. It is possible that more settings will be added at a later date, such as archive security key type or USMT migration status, at that point we can re-add the code to get the specific settings we need.

Also I seem to have introduced a bug in bulk processing that is causing multiple bulk processes to try and save their results to the same results array. I'm trying to track that down now, but it only occurs when I run the full test suite so that's a bit odd.

setup.py
```diff
@@ -36,6 +36,7 @@
         "c42eventextractor==0.4.0",
```
unparalleled-js, line 36: I believe we also need to add `!=3.5.*` to `python_requires=`

src/code42cli/cmds/devices.py
ceciliastevens: Ok, `devices list --include-settings` now does not include the extraneous `.get_settings()` call, and only adds full disk access status to Mac OS devices. As more settings are added in the future, we may need to re-add that call, but for now this is all that we're reporting on.

setup.py
ceciliastevens: Updated.

src/code42cli/cmds/devices.py
```diff
@@ -0,0 +1,449 @@
...
+    "--include-settings",
+    required=False,
+    type=bool,
+    default=False,
+    is_flag=True,
+    help="""Include device settings in output.""",
+)
+@format_option
+@sdk_options()
+def list(
```
timabrmsn, line 215: This should be renamed to `_list` so we don't shadow the built-in `list()`.

src/code42cli/cmds/devices.py
```diff
@@ -0,0 +1,449 @@
...
+        "archiveBytes": "Archive Size in Bytes",
+        "archiveGuid": "Archive GUID",
+    }
+
+
+def _get_device_info(sdk, device_guid):
+    return sdk.devices.get_by_guid(device_guid, include_backup_usage=True).data
+
+
+def _get_key_from_list_of_dicts(key, list_of_dicts):
```
timabrmsn, line 146: Looks like this is no longer used anywhere.

ceciliastevens: Removed.

src/code42cli/cmds/devices.py
ceciliastevens: Renamed it to `list_devices`, to match the other `list_backup_sets` method.

src/code42cli/cmds/devices.py
```diff
@@ -0,0 +1,445 @@
...
+        return device_dataframe.merge(result_list, how="left", on="guid")
+    except KeyError:
+        return device_dataframe
+
+
+def _drop_devices_which_have_not_connected_in_some_number_of_days(
+    devices_dataframe, days_since_last_connected
+):
+    return devices_dataframe.loc[
+        to_datetime(datetime.now(), utc=True)
```
timabrmsn, line 304: This doesn't actually convert the time to UTC time, it just sets the timezone info to UTC. 

```
>>> from datetime import datetime
>>> from pandas import to_datetime
>>> 
>>> now = datetime.now()
>>> now
datetime.datetime(2020, 12, 29, 14, 48, 43, 395819)
>>> to_datetime(now, utc=True)
Timestamp('2020-12-29 14:48:43.395819+0000', tz='UTC')
```

We'll want to use `datetime.utcnow()`. 

Also, can we turn these datetime conversions into variables to make the `.loc[]` logic more readable?

```
    now = to_datetime(datetime.utcnow(), utc=True)
    last_connected = to_datetime(devices_dataframe["lastConnected"], utc=True)
    delta = to_timedelta(days_since_last_connected, unit="days")
    return devices_dataframe.loc[now - last_connected > delta]
```

ceciliastevens: done! I removed the `utc=True` from the last_connected line as well, since to_timedelta does retrieve the time zone info from the timestamps our console provides. This way we will correctly raise an error if console starts giving us timestamps in non-UTC timezones.

timabrmsn: We need to keep that on the last_connected conversion, otherwise the subtraction won't work:

```
>>> last_connected = '2019-12-16T17:31:31.161-06:00'
>>> to_datetime(datetime.utcnow(), utc=True) - to_datetime(last_connected)
Traceback (most recent call last):
  File "<input>", line 1, in <module>
    to_datetime(datetime.utcnow(), utc=True) - to_datetime(last_connected)
  File "pandas/_libs/tslibs/timestamps.pyx", line 364, in pandas._libs.tslibs.timestamps._Timestamp.__sub__
TypeError: Timestamp subtraction must have the same timezones or no timezones
```

ceciliastevens: Oh, I see, it does do the conversion there. Ok, I've added it back.

src/code42cli/cmds/devices.py
```diff
@@ -0,0 +1,447 @@
...
+def bulk(state):
+    """Tools for managing devices in bulk."""
+    pass
+
+
+@bulk.command(
+    name="deactivate",
+    help="""Deactivate all devices on the given list.
+            Takes as input a CSV with a deviceId column.
+            The deviceId column must have the legacy deviceId value, not GUID.""",
```
timabrmsn, line 421: This is no longer true now that `_deactivate_device()` takes a guid.

tests/cmds/test_devices.py
```diff
@@ -0,0 +1,555 @@
...
+
+
+@pytest.fixture
+def devices_list_generator(mocker):
+    return [TEST_COMPUTER_PAGE]
+
+
+@pytest.fixture
+def users_list_generator(mocker):
+    return [TEST_USERS_LIST_PAGE]
```
timabrmsn, line 282: These don't need to be fixtures, they can just be regular generator functions:

```
def devices_list_generator():
     yield TEST_COMPUTER_PAGE
```

Then just call it in the test:

```
cli_state.sdk.devices.get_all.return_value = devices_list_generator()
```

Otherwise the mock return value ends up just being a regular list, and if the real code does something like call `next()` on the generator these tests would break unnecessarily.

tests/cmds/test_devices.py
```diff
@@ -0,0 +1,555 @@
...
+
+
+def test_deactivate_fails_if_device_is_on_legal_hold(
+    runner, cli_state, deactivate_device_in_legal_hold_failure
+):
+    result = runner.invoke(
+        cli, ["devices", "deactivate", TEST_DEVICE_ID], obj=cli_state
+    )
+    assert result.exit_code == 1
+    assert "The device {} is in legal hold.".format(TEST_DEVICE_ID)
```
timabrmsn, line 428: all of these that are checking the output text need to assert that the text is `in result.ouput`, currently, they're just asserting the string, which will always be true since it's not empty.

timabrmsn: I guess there's just the 3 missing the `in result.output`

test_deactivate_fails_if_device_does_not_exist
test_deactivate_fails_if_device_is_on_legal_hold
test_deactivate_fails_if_device_deactivation_forbidden

timabrmsn (changes requested):
A few final things I found when looking at the tests, then I think this is ready to merge!

tests/cmds/test_devices.py
ceciliastevens: updated

tests/cmds/test_devices.py
ceciliastevens: updated

timabrmsn (approved):

PR #168, opened 2020-12-02
Feature/List Detection Employees w/ Filtering

unparalleled-js:
`list` commands for `departing-employee` and `high-risk-employee`, really useful for checking the occupants of these lists w/o having to open the UI.

Test:

`code42 departing-employee list`
`code42 departing-employee list -f RAW-JSON`
`code42 departing-employee list -f CSV`

`code42 high-risk-employee list`
`code42 high-risk-employee list -f RAW-JSON`
`code42 high-risk--employee list -f CSV`

Questions:

Do we want to echo something like "nobody on list" when 0 results?

### DIFF
--- a/src/code42cli/cmds/departing_employee.py
+++ b/src/code42cli/cmds/departing_employee.py
@@ -31,7 +31,7 @@ def departing_employee(state):
 @sdk_options()
 def _list(state, format):
     """Lists the employees on the Departing Employee list."""
-    
+
     employee_generator = _get_departing_employees(state.sdk)
     list_employees(
         employee_generator,
--- a/src/code42cli/cmds/departing_employee.py
+++ b/src/code42cli/cmds/departing_employee.py
@@ -4,7 +4,7 @@ from py42.exceptions import Py42NotFoundError
 from code42cli.bulk import generate_template_cmd_factory
 from code42cli.bulk import run_bulk_process
 from code42cli.click_ext.groups import OrderedGroup
-from code42cli.cmds.detectionlists import HEADER_KEYS_MAP, list_employees
+from code42cli.cmds.detectionlists import list_employees
 from code42cli.cmds.detectionlists import update_user
 from code42cli.cmds.detectionlists.options import cloud_alias_option
 from code42cli.cmds.detectionlists.options import notes_option
--- a/src/code42cli/cmds/departing_employee.py
+++ b/src/code42cli/cmds/departing_employee.py
@@ -30,8 +30,15 @@ def departing_employee(state):
 @format_option
 @sdk_options()
 def _list(state, format):
+    """Lists the employees on the Departing Employee list."""
+    
     employee_generator = _get_departing_employees(state.sdk)
-    list_employees(employee_generator, format, "departing employee")
+    list_employees(
+        employee_generator,
+        format,
+        "departing employee",
+        {"departureDate": "Departure Date"},
+    )
 
 
 @departing_employee.command()
--- a/src/code42cli/cmds/departing_employee.py
+++ b/src/code42cli/cmds/departing_employee.py
@@ -4,7 +4,7 @@ from py42.exceptions import Py42NotFoundError
 from code42cli.bulk import generate_template_cmd_factory
 from code42cli.bulk import run_bulk_process
 from code42cli.click_ext.groups import OrderedGroup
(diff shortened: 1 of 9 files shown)


alanag13:
> Do we want to echo something like "nobody on list" when 0 results?

Probably, since this is an established pattern in other places where a list yields no results. I would say something like `There are currently no users on the <listname> list.`

src/code42cli/cmds/departing_employee.py
```diff
@@ -25,6 +27,18 @@ def departing_employee(state):
     pass
 
 
+@departing_employee.command("list")
+@format_option
+@sdk_options()
+def _list(state, format):
+    employee_generator = _get_departing_employees(state.sdk)
+    employee_list = [
+        employee for employees in employee_generator for employee in employees["items"]
```
alanag13, line 36: I would prefer if we avoid putting multiple instances of `for` on the same line.

src/code42cli/cmds/departing_employee.py
```diff
@@ -25,6 +27,18 @@ def departing_employee(state):
...
+    ]
+    formatter = OutputFormatter(format, HEADER_KEYS_MAP)
```
alanag13, line 38: should departing employees also include the departure date in the output?

kiran-chaudhary (changes requested):
Newline character in table format makes the output look weird a bit.
```
Username                             Notes                          Departure Date   
juliya.smith+partners@code42.com     test                           None             
spatel@code42.com                    Test workato test test         2020-10-20       
test-1603479510@test.com             overwrite notes
update notes   2020-11-02       
test-1603479374@test.com             overwrite notes
update notes   2020-11-02       
test-1603479222@test.com             overwrite notes
update notes   2020-11-02       
test-1603478993@test.com             overwrite notes

```

kiran-chaudhary:
An option to filter results by departing date too can be useful. Use case: A user is interested in knowing employees departing in next 1 month or between given dates.

unparalleled-js:
@kiran-chaudhary Thank you for your feedback! You found some good stuff, I addressed it all I believe except I made a compromise.  Instead of implementing some custom filtering via Departure Date, I added the API-given filter clauses as options.  They include `EXFILTRATION_24_HOURS` and `EXFILTRATION_30_DAYS` and departing employee even has one for `LEAVING_TODAY`... So these mirror the same functionality in the UI!  Thanks Kiran, let me know what you think

kiran-chaudhary:
> @kiran-chaudhary Thank you for your feedback! You found some good stuff, I addressed it all I believe except I made a compromise. Instead of implementing some custom filtering via Departure Date, I added the API-given filter clauses as options. They include `EXFILTRATION_24_HOURS` and `EXFILTRATION_30_DAYS` and departing employee even has one for `LEAVING_TODAY`... So these mirror the same functionality in the UI! Thanks Kiran, let me know what you think

This looks good now. 

Below points are not on our team's plate but a general suggestion to be taken up with concerned team:

- Analysing the data it seems there should be an option to filter out results of past dates i.e right now "departing employee" displays data even of "departed employees", if not added then `OPEN` filter will display current data even in 2022.

- As we renamed our product from "Insider threat" to "Insider Risk", on that same thought process, we should recommend renaming "EXFILTRATION_24_DAYS" to a milder word like "DEPARTING_WITHIN_24_DAYS".

src/code42cli/cmds/departing_employee.py
```diff
@@ -12,10 +14,17 @@
...
 from code42cli.file_readers import read_csv_arg
 from code42cli.file_readers import read_flat_file_arg
+from code42cli.options import format_option
 from code42cli.options import sdk_options
 
 
 DATE_FORMAT = "%Y-%m-%d"
+filter_option = click.option(
+    "--filter",
+    help="Departing employee filter options.",
```
kiran-chaudhary, line 24: Help message should specify the default option.

kiran-chaudhary: `ALL` seems more relevant than 'OPEN', right now this is mirroring API's naming convention so its fine, just a thought if we could create a map of custom names against the choices.

kiran-chaudhary (approved):
LGTM, Just one concern to address the default option in help text.

unparalleled-js:

PR #169, opened 2020-12-04
Bugfix/audit-logs checkpoints

timabrmsn:
Adds checkpointing to audit-logs commands. 

I had to rework the logic a bit since the audit logs results come down in reverse-chronological order (most recent first), the send-to command was sending the events "backwards", which also made checkpointing the last processed event problematic.

I also noticed some opportunities for refactoring around the click options logic (--begin and --checkpoint, etc), but I think that doesn't make sense until we remove the extractor and can refactor a few other things at once.


src/code42cli/cmds/auditlogs.py
```diff
@@ -0,0 +1,255 @@
...
+AUDIT_LOG_TIMESTAMP_FORMAT = "%Y-%m-%dT%H:%M:%S.%f"
+
+AUDIT_LOGS_DEFAULT_HEADER = OrderedDict()
+AUDIT_LOGS_DEFAULT_HEADER["timestamp"] = "Timestamp"
+AUDIT_LOGS_DEFAULT_HEADER["type$"] = "Type"
+AUDIT_LOGS_DEFAULT_HEADER["actorName"] = "ActorName"
+AUDIT_LOGS_DEFAULT_HEADER["actorIpAddress"] = "ActorIpAddress"
+AUDIT_LOGS_DEFAULT_HEADER["userName"] = "AffectedUser"
+AUDIT_LOGS_DEFAULT_HEADER["userId"] = "AffectedUserUID"
+# AUDIT_LOGS_DEFAULT_HEADER["success"] = "Success"
```
unparalleled-js, line 35: Could you remove the commented out sections?

src/code42cli/cmds/auditlogs.py
```diff
@@ -0,0 +1,255 @@
...
+# AUDIT_LOGS_DEFAULT_HEADER["resultCount"] = "ResultCount"
+
+filter_option_usernames = click.option(
+    "--username", required=False, help="Filter results by usernames.", multiple=True,
+)
+filter_option_user_ids = click.option(
+    "--user-id", required=False, help="Filter results by user ids.", multiple=True,
```
unparalleled-js, line 42: Might be an unpopular opinion, but I'd prefer if we capitalized acronyms, like `IDs`, when it shows up in user-facing documentation text.

src/code42cli/cmds/auditlogs.py
```diff
@@ -0,0 +1,255 @@
...
+AUDIT_LOGS_DEFAULT_HEADER["timestamp"] = "Timestamp"
+AUDIT_LOGS_DEFAULT_HEADER["type$"] = "Type"
+AUDIT_LOGS_DEFAULT_HEADER["actorName"] = "ActorName"
+AUDIT_LOGS_DEFAULT_HEADER["actorIpAddress"] = "ActorIpAddress"
+AUDIT_LOGS_DEFAULT_HEADER["userName"] = "AffectedUser"
+AUDIT_LOGS_DEFAULT_HEADER["userId"] = "AffectedUserUID"
+# AUDIT_LOGS_DEFAULT_HEADER["success"] = "Success"
+# AUDIT_LOGS_DEFAULT_HEADER["resultCount"] = "ResultCount"
+
+filter_option_usernames = click.option(
```
unparalleled-js, line 38: Is this a c42-username?  Might be worth sharing some of these filter options. File Events has a --c42-username and a I think it could potentially make use of a user-id filter (though it doesn't currently).

src/code42cli/cmds/auditlogs.py
```diff
@@ -0,0 +1,255 @@
...
+    help="Filter results by affected user ids.",
+    multiple=True,
+)
+filter_option_affected_usernames = click.option(
+    "--affected-username",
+    required=False,
+    help="Filter results by affected usernames.",
+    multiple=True,
+)
+filter_option_event_types = click.option(
```
unparalleled-js, line 63: It is just a tad odd that we can't even search file events by event type.

src/code42cli/cmds/auditlogs.py
```diff
@@ -0,0 +1,255 @@
...
+    "--event-type",
```
unparalleled-js, line 64: When searching file events, `-t` represent `Exposure Type`... Which, now in retrospect is confusing.  Users may not know what they are searching for, especially when you limit the logs by `Event Type` rather. Does this add even more to that uncertainty?

If there is a 2.0 list for the CLI, I propose we rename `-t` for file events searches.

src/code42cli/cmds/auditlogs.py
```diff
@@ -0,0 +1,255 @@
...
+    )
+
+
+def _search(sdk, format, save_checkpoint, **filter_args):
+    formatter = OutputFormatter(format, AUDIT_LOGS_DEFAULT_HEADER)
+    response_gen = sdk.auditlogs.get_all(**filter_args)
+    events = _get_all_audit_log_events(response_gen)
+    event_count = len(events)
+    if not event_count:
+        click.echo("No results found.")
```
unparalleled-js, line 205: `click.echo("No results found.")` is used like 4 times throughout the codebase, might be worth extracting into a shared method or a constant.

src/code42cli/cmds/auditlogs.py
```diff
@@ -0,0 +1,255 @@
...
+            if save_checkpoint:
+                ts = _parse_audit_log_timestamp_string_to_timestamp(event["timestamp"])
+                save_checkpoint(ts)
+
+
+def _get_all_audit_log_events(response_gen):
+    events = []
+    try:
+        for response in response_gen:
+            response_dict = json.loads(response.text)
```
unparalleled-js, line 235: Is this a `Py42Response`? You can just do `response_dict = response.data`

src/code42cli/cmds/auditlogs.py
```diff
@@ -0,0 +1,255 @@
...
+            if EVENT_KEY in response_dict:
+                events.extend(response_dict.get(EVENT_KEY))
+    except KeyError:
```
unparalleled-js, line 238: It looks this will never happen because of lines 236-237.  We already checking to see if the key is in the dict

src/code42cli/cmds/auditlogs.py
```diff
@@ -0,0 +1,255 @@
...
+        # API endpoint (get_page) returns a response without events key when no records are found
+        # e.g {"paginationRangeStartIndex": 10000, "paginationRangeEndIndex": 10000, "totalResultCount": 1593}
+        pass
+    return events
+    
+
+def _parse_audit_log_timestamp_string_to_timestamp(ts):
```
unparalleled-js, line 245: Won't this potentially cause us to miss events if we bump ms but already more events exist on the backend so now we list the end of the last ms of events?

alanag13:
@timabrmsn
> the audit logs results come down in reverse-chronological order (most recent first)
if the api doesn't allow for us to specify the sort direction, I think we should sort it here to account for that behavior potentially changing in the future.

src/code42cli/cmds/auditlogs.py
alanag13: the way the extractor handles this is to exclude the highest timestamp values from being retrieved and add 1ms to the _second_ highest timestamp.

unparalleled-js: We are going to have to port that logic over eventually anyway, maybe now is the time...

timabrmsn:
Ok, got the PR to only show my recent changes instead of everything Kiran already did.

src/code42cli/cmds/auditlogs.py
timabrmsn: I asked the same thing in the initial review, the exception would actually be raised from py42, but we haven't fixed that yet there, so we have to catch it here: https://github.com/code42/code42cli/pull/160#discussion_r527737186

src/code42cli/cmds/auditlogs.py
timabrmsn: So does any given extractor run always leave at least one event out? 

Since we're going to be pulling the checkpointing logic fully into the CLI when we remove the extractor, I wonder if we should update the cursor to also store a(n) event identifier(s) alongside the timestamp, that way we can always process all the events at the end of each run, but also dedupe events at the start of each run.

timabrmsn:
@alanag13 

> if the api doesn't allow for us to specify the sort direction, I think we should sort it here to account for that behavior potentially changing in the future

I can add a sorting.

src/code42cli/cmds/auditlogs.py
```diff
@@ -155,11 +201,39 @@ def send_to(
...
+    event_count = len(events)
+    if not event_count:
+        click.echo("No results found.")
+        return
+    elif event_count > 10:
+        click.echo_via_pager(formatter.get_formatted_output(events))
+    else:
+        formatter.echo_formatted_list(events)
+    if save_checkpoint:
+        ts = _parse_audit_log_timestamp_string_to_timestamp(events[0]["timestamp"])
```
unparalleled-js, line 217: So we are still getting the latest timestamp?  Is this a problem? I was expecting to see logic like this imported here: https://github.com/code42/c42eventextractor/blob/master/src/c42eventextractor/extractors.py#L169

src/code42cli/cmds/auditlogs.py
kiran-chaudhary: Wondering why don't we use the API filter option to implement checkpoint behaviour. The audit-log [API](https://support.code42.com/Administrator/Cloud/Monitoring_and_managing/Code42_API_resources/Search_Audit_Log_events_with_the_Code42_API) supports filter by date-range. We can pass the checkpoint value in the start-time field in the api request.

timabrmsn: That is what we're doing, we replace the `begin` argument with the checkpoint value if `--use-checkpoint` is given and a timestamp exists in the cursor: 

https://github.com/code42/code42cli/pull/169/files#diff-ac4db69e66795de05d5b7b5d999d10611f3cefe5f220cb6ba1442b9abe751714R132-R138

timabrmsn: I've implemented audit logs checkpointing logic to store hashes of the last events of any given run that share the same timestamp, then the next run's events are hashed and compared and pruned if they exist in the checkpoint.

timabrmsn:
This is ready for review again now, with updated cursor and event deduplication.

src/code42cli/cmds/auditlogs.py
```diff
@@ -153,43 +200,62 @@ def send_to(
...
-            pass
+
+def _dedupe_checkpointed_events_and_store_updated_checkpoint(
+    cursor, checkpoint_name, events
+):
+    checkpoint_events = cursor.get_events(checkpoint_name)
+    new_timestamp = None
+    new_events = []
+    for event in events:
+        event_hash = hash_event(event)
```
kiran-chaudhary, line 241: since the checkpoint that gets stored finally is the last event, why not just set the timestamp of the last event rather than do it one by one for all events. Calling `list` over this function will cause the generator to execute completely.

timabrmsn: I made `_dedupe_checkpointed_events_and_store_updated_checkpoint` a generator because `send-to` sends events one at a time to the server, and in the event of a network disconnect or other exception during sending, the checkpoint will remain accurate up to the last successful event processed, since we'd error out mid-generator. 

For `search` where we're just echoing to stdout (and need to know the length of the event list to determine _how_ we're echoing), we have to call list() on it first.

And it is possible that the last two events have the same timestamp, which if we want to accurately deduplicate across runs when events have the same timestamp, we need to store both of those last event hashes in the cursor, so we need to process every event in the list to be sure.

timabrmsn: I'll add a docstring to the function to make that intention clear for future readers.

unparalleled-js:
There should be a change to `test_profile.py` for testing that delete deletes the audit logs cursor checkpoint:

https://github.com/code42/code42cli/blob/master/tests/test_profile.py#L227

tests/cmds/test_auditlogs.py
```diff
@@ -92,15 +199,206 @@ def test_search_audit_logs_with_all_filter_parameters(runner, cli_state, date_st
...
+
+def test_send_to_without_existing_checkpoint_writes_both_event_hashes_with_same_timestamp(
+    cli_state,
+    runner,
+    test_audit_log_response_with_only_same_timestamps,
+    audit_log_cursor_with_checkpoint,
+    send_to_logger,
+):
+    cli_state.sdk.auditlogs.get_all.return_value = [
+        test_audit_log_response_with_only_same_timestamps
```
unparalleled-js, line 385: Two things about these tests make me nervous:

1.) The mock get_all responses should actually be generator instead of lists

2.)  I don't see any tests that make use of more than 1 page.

src/code42cli/cmds/auditlogs.py
```diff
@@ -103,11 +124,18 @@ def search(
     affected_user_id,
     affected_username,
     format,
+    use_checkpoint,
```
unparalleled-js, line 127: This is not a good name for this... This describes a boolean.

I know we probably can't change this now to maintain consistency with the other checkpoints, but we could at least do something like this:

```
if use_checkpoint:
  checkpoint_name = use_checkpoint
```

Seems redundant but would improve readability, as I just wasted like 2 minutes wondering why we have to pass a boolean into the `_dedup...()` method

src/code42cli/cmds/search/cursor_store.py
```diff
@@ -75,13 +76,31 @@ def __init__(self, profile_name):
         super().__init__(dir_path)
 
 
-def get_file_event_cursor_store(profile_name):
-    return FileEventCursorStore(profile_name)
+class AuditLogCursorStore(BaseCursorStore):
```
unparalleled-js, line 79: There should be test class `tests.cmds.search.test_cursor_store.TestAuditLogCursorStore` but I am not seeing it

src/code42cli/cmds/search/cursor_store.py
```diff
@@ -75,13 +76,31 @@ def __init__(self, profile_name):
...
+                    return []
+        except FileNotFoundError:
+            return []
 
-def get_alert_cursor_store(profile_name):
-    return AlertCursorStore(profile_name)
+    def replace_events(self, cursor_name, new_events):
+        location = path.join(self._dir_path, cursor_name) + "_events"
+        with open(location, "w") as checkpoint:
+            return checkpoint.write(json.dumps(new_events))
```
unparalleled-js, line 98: It does not seem like this return value gets used anywhere

src/code42cli/util.py
```diff
@@ -167,3 +169,9 @@ def _get_default_header(header_items):
             if key not in header and isinstance(key, str):
                 header[key] = key
     return header
+
+
+def hash_event(event):
```
unparalleled-js, line 174: `hash_json()` might be a more accurate name

unparalleled-js:
There is a whole lot of these logs, huh?  Running `code42 audit-logs search -c "myc" -b 2020-11-05`for the first time, started a few minutes ago.. Currently on the 200th page.....  

Should we warn about that?

Should we lower the begin date?  It is set to 90 days right now.

Oh wait it just errored...

```
ValueError: time data '2020-11-19T16:29:26' does not match format '%Y-%m-%dT%H:%M:%S.%f'
```

So looks like that timestamp bug

src/code42cli/cmds/auditlogs.py
```diff
@@ -81,16 +85,33 @@ def filter_options(f):
...
-    pass
+    # store cursor getter on the group state so shared --begin option can use it in validation
+    state.cursor_getter = _get_audit_log_cursor_store
+
+
+@audit_logs.command()
+@click.argument("checkpoint-name")
+@sdk_options()
+def clear_checkpoint(state, checkpoint_name):
+    """Remove the saved file event checkpoint from `--use-checkpoint/-c` mode."""
```
kiran-chaudhary, line 107: Remove the saved audit-log checkpoint

alanag13:
@unparalleled-js @timabrmsn 
>Oh wait it just errored...
ValueError: time data '2020-11-19T16:29:26' does not match format '%Y-%m-%dT%H:%M:%S.%f'
So looks like that timestamp bug

Are we using the extractor for this? If so, that bug is fixed in the extractor in master, it just hasn't been released.

unparalleled-js:
> @unparalleled-js @timabrmsn
> 
> > Oh wait it just errored...
> > ValueError: time data '2020-11-19T16:29:26' does not match format '%Y-%m-%dT%H:%M:%S.%f'
> > So looks like that timestamp bug
> 
> Are we using the extractor for this? If so, that bug is fixed in the extractor in master, it just hasn't been released.

We are not.. Tim also has come up with another way of handling the edges of the timestamp checking, which I think might be both less confusing and error-prone than how we are doing it in the extractor.

unparalleled-js:
For planning purpose, do we like this more? Should we make file events and alert checkpointing occur this way too?

src/code42cli/cmds/search/cursor_store.py
timabrmsn: it was like that on the parent class `replace()` method too, just copy/pasted. I've removed return from both.

src/code42cli/cmds/search/cursor_store.py
timabrmsn: Added these.

src/code42cli/cmds/auditlogs.py
timabrmsn: You're right, I've updated to make it clear.

tests/cmds/test_auditlogs.py
timabrmsn: Updated to use generator and to have tests that use more than one page.

src/code42cli/util.py
timabrmsn: Since this is aimed at deterministically hashing events that don't have a unique ID in them, this should take a dict of the event data so it ensures the json is sorted before hashing. 

Maybe I should just call it `hash_dict` and raise an exception if it's passed a string?

timabrmsn:
> > Are we using the extractor for this? If so, that bug is fixed in the extractor in master, it just hasn't been released.
> 
> We are not.. Tim also has come up with another way of handling the edges of the timestamp checking, which I think might be both less confusing and error-prone than how we are doing it in the extractor.

This error wasn't directly related to the new checkpointing logic at all, it was that our assumed timestamp format couldn't handle when it doesn't have a ms value. 

I've updated the parsing function to fix that now.

timabrmsn:
> For planning purpose, do we like this more? Should we make file events and alert checkpointing occur this way too?

I do like think this new way is better since it's easier to understand and also doesn't leave off the most recent events of any given run.

I think there's a bit more room for improvement to make this more flexible and re-usable, but I didn't want to spend more time holding this PR up. I think we can look at improving/abstracting checkpointing a bit more once we start extracting the extractor.

kiran-chaudhary (commented):
Observed performance degradation while testing. 
- One of the reason I can think of is, sorting takes place before filtering. Considering current implementation where irrespective of the checkpoint `audit-logs` API returns all results and sorting so many results before filtering them makes it slow.
- It will be slow for all results anyways due to sorting, filtering would help only when checkpoint is used. So if sorting is the culprit then, need to really ask whether sorting is needed or not. As in a use case, a user might be more interested in the logs of an event that has occurred recently over something that happened in past, given that a user is interested in logs over an alert that was triggered recently. 

These are general suggestions and can be thought of for future enhancements, if it makes sense

- Since we display events/logs/data in ascending order, it would be good to print the timestamp recorded by `--use-checkpoint` at the start or end of execution. When there are lots of results, in terminal it is tough to reach to the last result and figure out the timestamp recorded. 

- An option like `list-timestamp` that would display the name of the checkpoint and timestamp recorded against it, would be helpful to know beforehand what results a user might get using a checkpoint.

src/code42cli/cmds/auditlogs.py
```diff
@@ -153,43 +204,81 @@ def send_to(
...
+def _get_all_audit_log_events(sdk, **filter_args):
     response_gen = sdk.auditlogs.get_all(**filter_args)
-
     events = []
     try:
-        for response in response_gen:
-            response_dict = json.loads(response.text)
-            if EVENT_KEY in response_dict:
-                events.extend(response_dict.get(EVENT_KEY))
+        responses = list(response_gen)
```
alanag13, line 226: @timabrmsn This is giving me pause and making me reconsider my earlier recommendation that we sort here to protect against future changes in sorting behavior.

If I understand this correctly, if someone were to set up a recurring job with the start date being say, 90 days, ago and checking every 5 minutes, this would load the entire result set into memory and then sort it each time. If that date is still set to the same number a year from now, that in memory-list could get really huge and things are going to get significantly slower.

I realize now that the approach taken here is what is necessary in order to guarantee the sort order, but I failed to recognize that at the time I gave this recommendation. Sorry for the back-and-forth on this, but I think it would be better now if we assumed the default order and spoke to the audit log team to make sure about making sure not to change the default / keeping us in the loop if they do.

timabrmsn: But if they're running a recurring job, they'd be using a checkpoint, no? And that would cause the begin timestamp to be 5 mins ago for each run after the first one. 

So it's only if the initial run has a ton of events that we'd need to worry about memory/cpu time.

And I don't think we're going to hit any real-world performance problems here. I ran a `auditlogs.get_all()` on my test org and got ~550 events, then made a ton of copies of those to see what memory usage looked like. At around 10million events python was using ~4.5GB, and sorting those results took 3-4 seconds.

timabrmsn:
I think the performance degradation is likely not so much due to the sorting, but to the fact that we have to retrieve all pages before we can output anything now. 

Before we would process each page as they came in (which when writing to the pager meant you'd get a page right away, but to see the "next" page you'd have to exit the pager and it would load a new page in a new pager). Now you might wait a few seconds before any result show up, but they'll all be in the same pager so you can see all events at once. 

There's really no way around sorting results ourselves until the upstream team adds it to their API, since we really _need_ to send events with `send-to` in ascending order, and we _should_ print events in ascending order (imagine appending multiple `audit-logs search --use-checkpoint` events to a file, that file would would be all out of order).

Getting _all_ results and sorting also vastly simplifies the checkpoint logic. Which in my opinion is worth a little bit of a perceived performance hit.

src/code42cli/cmds/auditlogs.py
alanag13: Oh okay, I totally misunderstood the implementation then. I thought that the checkpoint was just determining which events should be printed, not which shouldn't be fetched.

I haven't touched the CLI in a while so I forgot that the checkpoint time would of course override what was input as the begin time in terminal. Carry on.

alanag13 (approved):

PR #173, opened 2020-12-09
More integration tests

kiran-chaudhary:
Added trivial integration test cases, those that do not need any pre-test setup or cleanup.


integration/test_departing_employee.py
```diff
@@ -0,0 +1,21 @@
...
+@pytest.mark.parametrize(
+    "command, error_msg",
+    [
+        (f"{DEPARTING_EMPLOYEE_COMMAND} add", "Missing argument 'USERNAME'."),
+        (f"{DEPARTING_EMPLOYEE_COMMAND} remove", "Missing argument 'USERNAME'."),
+        (f"{DEPARTING_EMPLOYEE_COMMAND} bulk add", "Missing argument 'CSV_FILE'."),
+        (f"{DEPARTING_EMPLOYEE_COMMAND} bulk remove", "Missing argument 'FILE'."),
+    ],
+)
+def test_departing_employee_command_returns_error_exit_status_when_missing_required_parameters(
```
kiran-chaudhary, line 16: These negative tests should fall under unit tests.

integration/test_alert_rules.py
```diff
@@ -0,0 +1,50 @@
+import pytest
+from integration import run_command
+
+ALERT_RULES_COMMAND = "code42 alert-rules"
+
+
+@pytest.mark.parametrize(
+    "command",
+    [
+        f"{ALERT_RULES_COMMAND} list",
```
unparalleled-js, line 10: `string.format` is more consistent with how we format strings in py42

kiran-chaudhary: changed.

integration/test_departing_employee.py
unparalleled-js: Yes I agree. Should we move them to their respective unit test modules? There might some tests of this similar nature dispersed already, but I know there is not a ton

unparalleled-js:
These look good! I have the same thought as you - that the negatives tests are more like integration tests when they verify args are present.  I think they are good tests to have. Just recently, we had a big where deleting profile name arg was no required, these tests would have caught that earlier.

unparalleled-js (approved):
looks good for now, but some of the same improvements for py42 integration tests apply here (make able to run via pytest)

kiran-chaudhary:
> looks good for now, but some of the same improvements for py42 integration tests apply here (make able to run via pytest)

Changed.

timabrmsn (approved):
LGTM. Were you still going to remove/move those negative tests that should be covered by unit tests?

kiran-chaudhary:
> LGTM. Were you still going to remove/move those negative tests that should be covered by unit tests?

We will keep those tests and I have added the `integration` marker to all positive cases here. So they both will stay. Now I plan to move all of them to `tests` directory after merging them.

PR #174, opened 2020-12-09
Add pre-test setup and post test cleanup for `send-to` command

kiran-chaudhary:
**Description**
Added integration tests for `send-to` commands.

Creates a server using `ncat` tool at port 5140 on localhost.
Executes actual command to test.
Closes the server.

### DIFF
--- a/tests/cmds/test_auditlogs.py
+++ b/tests/cmds/test_auditlogs.py
@@ -5,6 +5,9 @@ from logging import Logger
 
 import pytest
 from py42.response import Py42Response
+from tests.integration.util import DockerDaemon
+from tests.integration.util import SyslogServer
+from tests.integration import run_command
 from requests import Response
 
 from code42cli.click_ext.types import MagicDate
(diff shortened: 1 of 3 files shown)


alanag13:
@kiran-chaudhary `docker-compose up -d` will docker containers as daemons and would probably be a cleaner approach here that works on both mac and Unix. Likewise, `docker-compose down` will stop the daemon. I'd also prefer to see those called in a shell script instead of within python.

kiran-chaudhary:
> @kiran-chaudhary `docker-compose up -d` will launch docker containers as daemons and would probably be a cleaner approach here that works on both mac and Unix. Likewise, `docker-compose down` will stop the daemon. I'd also prefer to see those called in a shell script instead of within python.

Just to make sure we all are on same page, referring to the same script, added the shell script below, with few modification explained in **bold**
So `docker-compose up -d` is part of the script. We cannot have `docker-compose down` in the shell script until the test passes, so in the __exit__ method I will call `docker-compose down`. 

```
docker build --rm -t centos/rsyslog .;
docker-compose up -d; 
# you're not allowed to use privileged commands in a Dockerfile but since we've started
# the container with the privileged flag in docker-compose, we run this now right after starting.
#docker exec test-server_centossyslog_1 service rsyslog restart; ** This fails for me**
docker start test-server_centossyslog_1. **Added**
```

integration/test_auditlogs.py
```diff
@@ -3,13 +3,38 @@
...
+            yield run_command
+
+
 @pytest.mark.parametrize("command", [("{} '{}'".format(BASE_COMMAND, begin_date_str))])
 def test_auditlogs_search(command):
     return_code, response = run_command(command)
     assert return_code == 0
+
+
+@pytest.mark.parametrize(
```
unparalleled-js, line 28: Why is this parameterized? 
in the case where we want to add more commands, we don't to include `code42 audit-logs sent-to` on every parameter since it does not change

kiran-chaudhary: At that time I had thought of 2 commands to be tested one with `TCP` protocol option and another with `UDP`. Now that we have `TCP-TLS` having parameterize makes more sense. 

 Moreover, using parameterize even for a single test case is useful while debugging errors. With parameterize option we can see the actual command executed in this particular case... Below is a sample output of `pytest -v` 
 
 `tests/integration/test_auditlogs.py::test_auditlogs_send_to[code42 audit-logs send-to localhost -p TCP -b '2021-01-19 13:32:46'] ERROR                                                               [  4%]
tests/integration/test_auditlogs.py::test_auditlogs_search_command_returns_success_return_code[code42 audit-logs search -b '2021-01-19 13:32:46'] FAILED                                             [  9%]
tests/integration/test_auditlogs.py::test_auditlogs_search_command_returns_success_return_code[code42 audit-logs search -b '2021-01-19 13:32:46' -e '2021-01-21 13:32:46']`

PR #175, opened 2020-12-09
Bugfix: make profile name required for delete

unparalleled-js:
* Adds a missing test

* Safer delete profile method

### DIFF
--- a/src/code42cli/profile.py
+++ b/src/code42cli/profile.py
@@ -108,6 +108,7 @@ def create_profile(name, server, username, ignore_ssl_errors):
 
 def delete_profile(profile_name):
     profile = _get_profile(profile_name)
+    profile_name = profile.name
     if password.get_stored_password(profile) is not None:
         password.delete_password(profile)
     cursor_stores = get_all_cursor_stores_for_profile(profile_name)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
- # Changelog
+# Changelog
 
 All notable changes to this project will be documented in this file.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 ### Fixed
 
-- Issue where `code42 profile delete` was allowed even deleting the default profile is not allowed.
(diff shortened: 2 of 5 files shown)


src/code42cli/profile.py
```diff
@@ -108,6 +108,7 @@ def create_profile(name, server, username, ignore_ssl_errors):
 
 def delete_profile(profile_name):
     profile = _get_profile(profile_name)
+    profile_name = profile.name
```
timabrmsn, line 111: How would these get out of sync?

unparalleled-js: If you call the method with None it uses the default profile, and as a hacker, I actually use this individual module in other non-CLI related scripts (creating SDKs from CLI profiles for py42 related stuff).

I don't see anything bad that would happen because of this change in the CLI

unparalleled-js: It makes sense from the command perspective to not be allowed to delete the default profile, but from this module's perspective, if you are grabbing the profile from the name, you should use the name of the profile object because that is the real profile name you are working with

unparalleled-js: If you look at other methods, that is what happens too, so this is consistent with that

timabrmsn: Ok cool, I was just wondering if you had uncovered a bug caused by this, to get a sense of what the change did. 

I do see that if you just call `code42 profile delete` without a profile name, it errors out. This change allows it to delete the default, but then it prints the name wrong since we're just printing the argument passed: 

```
 # code42 profile delete

Deleting this profile will also delete any stored passwords and checkpoints. Are you sure? (y/n): y
Profile 'None' has been deleted.
```

It also doesn't print the "<profile_name> is currently the default profile!" message in the warning.. 

A simple fix would be to just make the profile_name arg required on just the delete command.. Thoughts?

unparalleled-js: Oh ok, yes I had thought that it _was_ required. If you are not allowed to delete the default profile, it really should be required. Are you sure that it is this change causing this error? What happens on master branch? There might actually be a bug related to this. Awesome that this was uncovered.

unparalleled-js: Yes I verified that this bug is indeed already on Master, so I will fix it as an adittion to this PR

timabrmsn (approved):

PR #178, opened 2020-12-14
Chore/rm dlrm errs

unparalleled-js:
Switches to using py42 Remove detection list employee exceptions rather than CLI's
I have been waiting for the py42 release to open this PR.

It deleted unneeded code for error handling; the only change the user see is that the exceptions now reference the employees by ID rather than username.


kiran-chaudhary (approved):

kiran-chaudhary:
Note the style check failing before merging.

PR #179, opened 2020-12-15
Feature/list detection lists

unparalleled-js:
Edit: I have to reopen this same PR because the CLA assistant is failing and I have spent too long trying to figure out why.

list commands for departing-employee and high-risk-employee, really useful for checking the occupants of these lists w/o having to open the UI.

Test:

code42 departing-employee list
code42 departing-employee list -f RAW-JSON
code42 departing-employee list -f CSV

code42 high-risk-employee list
code42 high-risk-employee list -f RAW-JSON
code42 high-risk--employee list -f CSV

Questions:

Do we want to echo something like "nobody on list" when 0 results?


src/code42cli/cmds/detectionlists/__init__.py
```diff
@@ -1,4 +1,30 @@
...
+
+def list_employees(employee_generator, output_format, additional_header_items=None):
+    additional_header_items = additional_header_items or {}
+    header = {"userName": "Username", "notes": "Notes", **additional_header_items}
+    employee_list = []
+    for employees in employee_generator:
+        for employee in employees["items"]:
+            if employee["notes"] and output_format == OutputFormat.TABLE:
+                employee["notes"] = (
+                    employee["notes"].replace("\n", ". ").replace("\t", ". ")
```
timabrmsn, line 16: I don't know if we should assume that every newline/tab doesn't already have a period in the note. 

If someone added a note with a few sentences with periods separated by newlines, we'd be duplicating the periods. 

And this translation should also happen for csv output, otherwise it'll write the newlines and mess up the csv format too. 

Maybe we should just translate any newlines to explicit `\n` strings so the intent stays the same but it keeps it all on a single line for printing?

src/code42cli/cmds/high_risk_employee.py
```diff
@@ -16,8 +18,18 @@
 from code42cli.errors import Code42CLIError
 from code42cli.file_readers import read_csv_arg
 from code42cli.file_readers import read_flat_file_arg
+from code42cli.options import format_option
 from code42cli.options import sdk_options
 
+
+filter_option = click.option(
+    "--filter",
+    help="High risk employee filter options. Defaults to OPEN.",
```
timabrmsn, line 27: Naming the default `OPEN` here is a bit confusing, as I was kind of assuming you'd be able to alternately filter by `CLOSED`, which doesn't seem to be an option. 

Could we just expose `EXFILTRATION_24_HOURS|EXFILTRATION_30_DAYS|LEAVING_TODAY` as available filter options and say the default is "no filter"?

tests/cmds/conftest.py
```diff
@@ -93,3 +108,16 @@ def f(*args):
...
+
+def get_generator_for_get_all(mocker, mock_return_items):
+    mock_return_items = mock_return_items or ""
+
+    def gen(*args, **kwargs):
+        response = mocker.MagicMock(spec=Request)
+        response.text = """{{"items": [{0}]}}""".format(mock_return_items)
+        pages = [response]
+        for page in pages:
+            yield Py42Response(page)
```
timabrmsn, line 121: You don't have to wrap a single item in a list here, you can just:

```
yield Py42Response(response)
```
and the generator will only have one item.

src/code42cli/cmds/high_risk_employee.py
unparalleled-js: How about `ALL` ?

src/code42cli/cmds/detectionlists/__init__.py
unparalleled-js: Screen Shot 2020-12-15 at 21 59 57


Newlines don't matter in CSVs

Your other points are valid however

unparalleled-js: I did first try replace newlines with explicit `\n` but the table outputter never liked it

timabrmsn (approved):

PR #183, opened 2020-12-18
Change/user to actor al

unparalleled-js:
* In the request data, these properties are referred to as `actor` and not `username` etc.
* In py42, at the low service level, we have the args as such. However, at the client level, we seem have to renamed and that has caused confusions that has trickled into the CLI because the response already does not distinguish the `affected` user properties like the request params do, and therefore there no way of really knowing what goes where. By using the correct `actor` params, it is much easier to orientate


src/code42cli/cmds/auditlogs.py
```diff
@@ -34,14 +34,17 @@
 
 
 filter_option_usernames = click.option(
-    "--username", required=False, help="Filter results by usernames.", multiple=True,
+    "--actor", required=False, help="Filter results by actors.", multiple=True,
```
timabrmsn, line 37: should this maybe be `--actor-username` to match the `--affected-username/--affected-user-id` pattern and make explicit what value it expects?

timabrmsn (approved):

PR #190, opened 2020-12-22
Feature/ssl sendto

unparalleled-js:
- New choice `TLS-TCP` for `--protocol` option used by `send-to` commands:
    - `code42 security-data send-to`
    - `code42 alerts send-to`
    - `code42 audit-logs send-to`
    for more securely transporting data.
      
- `--certs` option for `send-to` commands when using `--protocol TLS-TCP`.

Also includes a bit of a refactor, and adds missing tests by making it super easy to test both `search` and `send-to` at the same time.

I have an unpushed branch, on the extractor project, that includes dev certs, rsyslog config updates, and Dockerfile updates that makes this easier.  It is ok to push a branch with dev certs in it?

Here is how I tested (though I also made a branch on the extractor automatically move the certs into their correct locations):

```
Followed this guide: https://www.thegeekdiary.com/how-to-configure-rsyslog-server-to-accept-logs-via-ssl-tls/
I didn't put any data for any of the fields excepting choosing `y` for all yes/no questions

From /etc/pki/tls/certs directory:

certtool --generate-privkey --outfile ca-key.pem
certtool --generate-self-signed --load-privkey ca-key.pem --outfile ca.pem
certtool --generate-privkey --outfile rslclient-key.pem --bits 2048
certtool --generate-request --load-privkey rslclient-key.pem --outfile request.pem
certtool --generate-certificate --load-request request.pem --outfile rslclient-cert.pem --load-ca-certificate ca.pem --load-ca-privkey ca-key.pem
certtool --generate-privkey --outfile rslserver-key.pem --bits 2048
certtool --generate-request --load-privkey rslserver-key.pem --outfile request.pem
certtool --generate-certificate --load-request request.pem --outfile rslserver-cert.pem --load-ca-certificate ca.pem --load-ca-privkey ca-key.pem
cp rslserver* ../private/
cp ca.pem ../private/

service rsyslog restart

docker cp docker_centossyslog_1:/etc/pki/tls/certs .#
# Then remove or ignore irrelevant certs
```


src/code42cli/cmds/search/options.py
```diff
@@ -174,10 +175,37 @@ def create_search_options(search_term):
...
-
         f = search_interval_options(f, search_term)
         f = checkpoint_option(f)
         f = advanced_query_option(f)
         return f
 
     return search_options
+
+
+def server_options(f):
```
unparalleled-js, line 186: moved from a different file

src/code42cli/cmds/securitydata.py
```diff
@@ -194,24 +197,6 @@ def clear_checkpoint(state, checkpoint_name):
     _get_file_event_cursor_store(state.profile.name).delete(checkpoint_name)
 
 
-def _call_extractor(
```
unparalleled-js, line 197: moved to bottom of file and called in another helper method

src/code42cli/logger/__init__.py
```diff
@@ -5,13 +5,13 @@
...
 from threading import Lock
 
-from c42eventextractor.logging.formatters import FileEventDictToCEFFormatter
-from c42eventextractor.logging.formatters import FileEventDictToJSONFormatter
-from c42eventextractor.logging.formatters import FileEventDictToRawJSONFormatter
-from c42eventextractor.logging.handlers import NoPrioritySysLogHandlerWrapper
 from click.exceptions import ClickException
 
 from code42cli.cmds.search.enums import FileEventsOutputFormat
+from code42cli.logger.formatters import FileEventDictToCEFFormatter
```
unparalleled-js, line 11: just moved these form the extractor to here so now the only thing that remains in the extractor is the extractor itself, so one step closer to being able to delete that project

src/code42cli/logger/formatters.py
```diff
@@ -0,0 +1,139 @@
...
+from code42cli.logger.maps import FILE_EVENT_TO_SIGNATURE_ID_MAP
+from code42cli.logger.maps import JSON_TO_CEF_MAP
+
+CEF_TEMPLATE = (
+    "CEF:0|Code42|{productName}|1|{signatureID}|{eventName}|{severity}|{extension}"
+)
+CEF_TIMESTAMP_FIELDS = ["end", "fileCreateTime", "fileModificationTime", "rt"]
+
+
+class FileEventDictToCEFFormatter(Formatter):
```
unparalleled-js, line 15: moved from the extractor (yes I moved the tests too)

src/code42cli/logger/handlers.py
```diff
@@ -0,0 +1,122 @@
+import logging
+import socket
+import ssl
+from logging.handlers import SysLogHandler
+
+from code42cli.cmds.search.enums import ServerProtocol
+
+
+class NoPrioritySysLogHandler(SysLogHandler):
```
unparalleled-js, line 9: THIS! This is the best syslog server logger handler in python on the internet haha. Well, that is the intent anyway.

It combines the current state of the syslog handler code with the socket wrapping logic I learned from https://github.com/lhl/python-syslogssl

PLUS it still strips out the Priority related logic that we were not using

src/code42cli/logger/handlers.py
```diff
@@ -0,0 +1,122 @@
...
+        self._hostname = hostname
+        self._port = port
+        self._certs = certs
+        self.address = (hostname, port)
+        logging.Handler.__init__(self)
+        self._use_insecure = protocol != ServerProtocol.TLS
+        self.socktype = _try_get_socket_type_from_protocol(protocol)
+        self._socket = None
+
+    @property
```
unparalleled-js, line 34: I also was able to remove the `wrapper` class by making `socket` lazily connect

src/code42cli/options.py
```diff
@@ -136,26 +135,3 @@ def decorator(f):
         return f
 
     return decorator
-
-
-def server_options(f):
```
unparalleled-js, line 141: moved to the shared search options since this is search related

tests/cmds/conftest.py
```diff
@@ -119,3 +117,9 @@ def gen(*args, **kwargs):
         yield Py42Response(response)
 
     return gen
+
+
+def get_mark_for_search_and_send_to(command_group):
```
unparalleled-js, line 122: this method created parametized  tests specifically for both `search` and `send-to` and the testing diff looks complex because mostly started using this instead of the disparate other strategies that were employed

unparalleled-js:
Will try and find a way to have the imports work

timabrmsn:
We seem to just hang indefinitely if the TLS handshake doesn't get a response or there's another cert error. Can we add a timeout and clear failure message printed when the connection doesn't work?

unparalleled-js:
timeout does not appear to be working yet

src/code42cli/cmds/search/options.py
```diff
@@ -174,10 +175,37 @@ def create_search_options(search_term):
...
+    hostname_arg = click.argument("hostname")
+    protocol_option = click.option(
+        "-p",
+        "--protocol",
+        type=click.Choice(ServerProtocol(), case_sensitive=False),
+        default=ServerProtocol.UDP,
+        help="Protocol used to send logs to server. "
+        "Use TLS for additional security. Defaults to UDP.",
```
kiran-chaudhary, line 194: what about `--secure` option which accepts the `certs` file, i.e `--secure <cert file>`.  
Let the protocol's options remain as TCP or UDP as earlier. 

With current options, we have to cover cases like what if user passes `-p  TCP --certs <cert file>` or `-p UDP --certs <cert file>`? Do we make secure data transmission? 
Having `-p TCP --secure <cert file>` makes intent more clear where one is trying to say "I wan to transfer the data more securely" and `--secure` is not passed they are ok with insecure data transmission.

We can explain that we use TLS encryption in `--secure` help message.

src/code42cli/cmds/search/options.py
```diff
@@ -174,10 +175,37 @@ def create_search_options(search_term):
...
+
+
+def server_options(f):
+    hostname_arg = click.argument("hostname")
+    protocol_option = click.option(
+        "-p",
+        "--protocol",
+        type=click.Choice(ServerProtocol(), case_sensitive=False),
+        default=ServerProtocol.UDP,
+        help="Protocol used to send logs to server. "
```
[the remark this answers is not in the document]
unparalleled-js: No there is not. It is a multiline. Look at that line under it.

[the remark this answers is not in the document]
unparalleled-js: When you remove the trailing space of the first line, the help text looks like this: 
`Protocol used to send logs to server.Use TLS for additional security. Defaults to UDP.`

src/code42cli/cmds/search/options.py
unparalleled-js: Yeay maybe. But there are caveats there similarly.

Like protocol has to always be `TCP` when using secure. There is no secure `UDP`.

unparalleled-js: Plus `--secure` would have to both work as a flag and as an option because you can use NoCert option for TLS

kiran-chaudhary: Wondering if there will be a use case when a user would prefer insecure TCP communication over secure.

alanag13: unlike http/s, udp/tcp log transmission is still very commonly done insecurely since it typically is all done from within the same network. This is why we haven't even added this feature until now.

insecure should remain the default because it is still more common and also because changing this would be a breaking change.

unparalleled-js: I honestly can go either way on the design of this, `--protocol TCP --secure <optional-certs-files>` or `--protocol TLS --certs <certs-files>` or another way.

It does seem odd to have to explicitly say I want `--protocol TCP` if you want to use TLS-TCP.

But having the certs option only apply to one protocol is also a drawback

kiran-chaudhary: Looking at all the different scenarios and given that UDP still remains insecure, what about 

-  Defaults to UDP when protocol is not specified, as it is now.
- --protocol UDP     # Same as above just that the user is being explicit 
-  --protocol TCP     # Default insecure when TCP is specified
- --protocol TCP --TLS     # Secure TCP, --TLS should be allowed only when TCP is specified
- --protocol TCP --CA <CERT> # secure TCP, --CA should be allowed only when TCP is specified

alanag13: I would say that `--certs` could default to using TCP, and error if you specify a protocol known to be not compatible (in this case UDP).

I would prefer `--certs` over `--secure` since passing certs locations  to a param named `secure` makes a little less sense contextually, imo.

unparalleled-js: OK, but take note to all of the complexitie that arise from doing it this way ( in the next commit)

unparalleled-js: For example, this is what the new help text for protocol will be like: 

```
        help="Protocol used to send logs to server. "
        "Defaults to UDP unless providing certs, than it defaults to TCP.",
```

unparalleled-js: Also, does this mean we are not going to support `ssl.CERT_NONE` ?

unparalleled-js: If we do want to support no certs, how would one do that?

`send-to --certs None`

That command does really seem like it is using TLS, yet it can be....

unparalleled-js: Reasons like this (And more!) are why I thought it would make more sense to have TLS be part of the protocol... So, here is another middle-ground:


`code42 security-data send-to --protocol TLS-TCP`

Add certs optionally.

alanag13: You make a valid point about the default protocol being oddly conditional, which I agree should be avoided.
`CERT_NONE` is effectively just a way to bypass certificate validation, which is probably okay to support here for those that _really_ want it since this feature is mostly targeted at on-prem services that could have self-signed certs.

tests/cmds/test_devices.py
```diff
@@ -200,8 +200,8 @@
             "userId": 1014,
             "userUid": "836473273124890369",
             "status": "Active",
-            "username": "qatest@code42.com",
-            "email": "qatest@code42.com",
+            "username": "qatest@example.com",
+            "email": "qatest@example.com",
             "firstName": "Chad",
```
alanag13, line 205: Anonymize names

src/code42cli/logger/handlers.py
```diff
@@ -0,0 +1,134 @@
...
+    def socket(self):
+        if self._socket is None:
+            self.init_socket()
+        return self._socket
+
+    def init_socket(self):
```
alanag13, line 40: should this method be private, or does it override something on the base class?

src/code42cli/cmds/alerts.py
```diff
@@ -14,20 +12,22 @@
...
 
 
-SEARCH_DEFAULT_HEADER = OrderedDict()
-SEARCH_DEFAULT_HEADER["name"] = "RuleName"
-SEARCH_DEFAULT_HEADER["actor"] = "Username"
-SEARCH_DEFAULT_HEADER["createdAt"] = "ObservedDate"
-SEARCH_DEFAULT_HEADER["state"] = "Status"
-SEARCH_DEFAULT_HEADER["severity"] = "Severity"
-SEARCH_DEFAULT_HEADER["description"] = "Description"
+def _get_search_default_header():
```
kiran-chaudhary, line 22: Don't we need this `dict` to be an `OrderedDict`?

kiran-chaudhary: Similarly in other places?

unparalleled-js: No, since we dropped support for py3.5, there is no point in using `OrderedDict` anymore

src/code42cli/logger/handlers.py
```diff
@@ -0,0 +1,130 @@
...
+        self._hostname = hostname
+        self._port = port
+        self._certs = certs
+        self.address = (hostname, port)
+        logging.Handler.__init__(self)
+        self._use_insecure = protocol != ServerProtocol.TLS_TCP
+        self.socktype = _try_get_socket_type_from_protocol(protocol)
+        self.socket = None
+
+    def connect_socket(self):
```
unparalleled-js, line 34: So I couldn't really make this method internal as @alanag13  suggested because I want this to get called in the `get_logger()` method so that we can test the connection.. I don't want it initialize in the init method because it makes some of the testing require less mocks (And also principally it is better to not  make a network request in an initializer). I hope these changes make it more obvious.. See the `get_logger_for_server` method as well.

unparalleled-js:
I don't know if it is a problem with this code or my test server, but my syslog raises this error: `unexpected GnuTLS error -110 in nsd_gtls.c:551: The TLS connection was non-properly terminated.`

I can confirm that the socket's close method is indeed getting called...

unparalleled-js:
I could use help testing because I don't think my test server cares about the certs I tried to set up

unparalleled-js:
I should add that the logs do get transmitted correctly with SSL, so I do believe things are working as intended.

alanag13:
> I don't know if it is a problem with this code or my test server, but my syslog raises this error: `unexpected GnuTLS error -110 in nsd_gtls.c:551: The TLS connection was non-properly terminated.`

This seems to suggest that that error has to do with how the server implements TLS connections and can probably be safely ignored:
https://stackoverflow.com/questions/32113065/gnutls-error-110-the-tls-connection-was-non-properly-terminated

kiran-chaudhary:
During testing observed error `Exception: Unable to connect to 127.0.0.1. Failed with error: _ssl.c:835: The handshake operation timed out.`.  Even though observed encrypted data transfer.

unparalleled-js:
> During testing observed error `Exception: Unable to connect to 127.0.0.1. Failed with error: _ssl.c:835: The handshake operation timed out.`. Even though observed encrypted data transfer.
> 
> Command: `code42 security-data send-to -p TLS-TCP 127.0.0.1 -b 2021-01-01`

Talked offline, needs Server config

kiran-chaudhary:
> > During testing observed error `Exception: Unable to connect to 127.0.0.1. Failed with error: _ssl.c:835: The handshake operation timed out.`. Even though observed encrypted data transfer.
> > Command: `code42 security-data send-to -p TLS-TCP 127.0.0.1 -b 2021-01-01`
> 
> Talked offline, needs Server config

I do not have administrator privileges to test this on my local machine. `certtool` command fails for me, I need to figure out with IT team here to figure this out.

kiran-chaudhary (approved):
As discussed [here](https://github.com/code42/code42cli/pull/190#issuecomment-760916300),  Could not test and verify but can't keep this blocked until then. 
Approving as LGTM.

timabrmsn:
The error you were seeing on the server-side on close is because just calling `socket.close()` closes the TCP connection without tearing down the TLS layer first. Calling `.unwrap()` on the socket (if TLS) fixes it:

```
    def close(self):
        if not self._use_insecure:
            self.socket.unwrap()
        self.socket.close()
        logging.Handler.close(self)
```

tests/logger/test_handlers.py
```diff
@@ -0,0 +1,204 @@
...
+        handler.setFormatter(formatter)
+        handler.emit(mock_file_event_log_record)
+        expected_message = (formatter.format(mock_file_event_log_record) + "\n").encode(
+            "utf-8"
+        )
+        handler.socket.sendto.assert_called_once_with(
+            expected_message, (_TEST_HOST, _TEST_PORT)
+        )
+
+    def test_close_when_using_tls_unwraps_socket(self):
```
unparalleled-js, line 182: Added these `close` tests and also verified that unwrapping the socket does indeed make the server happy

src/code42cli/logger/handlers.py
```diff
@@ -0,0 +1,130 @@
...
+    if not use_insecure:
+        sock = _wrap_socket_for_ssl(sock, certs)
+    if sock_type == socket.SOCK_STREAM:
+        sock = _connect_socket(sock, sa)
+    return sock
+
+
+def _wrap_socket_for_ssl(sock, certs):
+    certs = certs or None
+    cert_reqs = ssl.CERT_REQUIRED if certs else ssl.CERT_NONE
```
timabrmsn, line 102: This is not what I would expect the default behavior to be. If I don't provide a cert file, it doesn't do any validation? If an end-user's server has a valid CA-signed cert, then they would expect to not need to provide a cert file, and would expect it to validate automatically.

I think this should be: If I try to connect to a server with an invalid/self-signed cert without using `--cert` option, it should always warn and fail. I can use `--cert` to provide that cert if I know it's valid but just not CA-signed. And we should maybe have a way to explicitly skip validation (maybe `--certs ignore`?), for those cases where it's a self-signed cert on the server, but they don't have access to that cert file. 

We can do this by something like:
```
context = ssl.create_default_context(cafile=certs, server_hostname=host)
sock = context.wrap_socket(sock)
```

If `cafile` is None, it defaults to using OS cert chains for validation. 

https://docs.python.org/3/library/ssl.html#socket-creation

unparalleled-js: Ah, that makes sense. I did not consider automatic cert validation.

Is an inaccessible self-signed cert the only case where skipping validation would make sense? Seems like the user should just use TCP alone then.

unparalleled-js: I want to make sure of this behavior, because the ssl.wrap_context method defaults to CERT_NONE if you don't pass any certs

unparalleled-js: Have you confirmed that automatica cert validation is not currently happening?

timabrmsn: That's probably the main case, but there could be others. Like if the server had a valid CA-signed cert, but dns lookup wasn't working on the client machine so they had to use just the host IP.. 

But plain TCP wouldn't work in the cases where the server is only listening for TLS connections, as the connection would get rejected server-side.

unparalleled-js: Ah! Looks like you are right. This top level method is actually deprecated for reasons similar to this.

unparalleled-js: Is the phrase `CERT_NONE` common enough that we could have `--certs NONE` or `--certs none` work for ignoring them?

unparalleled-js: CURL uses `--insecure` in this case, so we could have an extra flag that only applies to `TLS`, but I can see that getting confusing

timabrmsn: I'd rather have the meaning explicit in what the user has to type, so `--certs none` doesn't quite do it. An additional `--insecure` option is fine IMO.

alanag13: `--insecure` only applies to certain types of request in curl too 😛 . I agree with @timabrmsn that being insecure is something we would want to make sure someone doesn't opt into unknowingly, we should be as clear as possible here.

alanag13: @timabrmsn good find in considering certs signed by a valid CA. That wasn't something that I hadn't considered since this functionality is typically used to point at an on-prem syslog server, but it's certainly a valid case.

unparalleled-js: The reason I thought it would confusing was because of the UDP/TCP - those could also be considered insecure.

Perhaps `--ignore-cert-validation` is best. Explicit.

unparalleled-js:
When making the latest changes, I noticed that Alerts send_to defaults to JSON while the securitydata and auditlogs default to RAW-JSON.

src/code42cli/cmds/alerts.py
```diff
@@ -272,12 +272,15 @@ def send_to(
...
     or_query,
     certs,
+    ignore_cert_validation,
     **kwargs
 ):
     """Send alerts to the given server address.
 
     HOSTNAME format: address:port where port is optional and defaults to 514.
     """
+    if ignore_cert_validation:
```
unparalleled-js, line 281: I tried putting this in a callback, having it set the other param automatically, but couldn't get it to work right

unparalleled-js: I can create custom command type to handle it though

src/code42cli/logger/handlers.py
unparalleled-js: Similarly to before, I can't ever successfully use certs, maybe because they are self-signed?
Other than that, I think the behavior of using the `--ignore-cert-validation` works now the same as it did before when excluding the `--certs` flag.  Perhaps there are some implementation improvements yet though

src/code42cli/cmds/alerts.py
unparalleled-js: Or include it in the CLIState

src/code42cli/logger/handlers.py
timabrmsn: I got it to work by generating a key + cert with localhost and 127.0.0.1 as subject alternate name (SAN) params (https://medium.com/@antelle/how-to-generate-a-self-signed-ssl-certificate-for-an-ip-address-f0dd8dddf754)

Create a config file:
```
[req]
default_bits  = 2048
distinguished_name = req_distinguished_name
req_extensions = req_ext
x509_extensions = v3_req
prompt = no
[req_distinguished_name]
countryName = XX
stateOrProvinceName = N/A
localityName = N/A
organizationName = Self-signed certificate
commonName = 120.0.0.1: Self-signed certificate
[req_ext]
subjectAltName = @alt_names
[v3_req]
subjectAltName = @alt_names
[alt_names]
IP.1 = 127.0.0.1
DNS.1 = localhost
```

Generate the cert+key from the config: 

```
openssl req -x509 -nodes -days 730 -newkey rsa:2048 -keyout key.pem -out cert.pem -config san.cnf
```

Now tell ncat to use that cert when listening with ssl: 

```
ncat -kl -p 9999 --ssl --ssl-cert cert.pem --ssl-key key.pem
```

Running the CLI without --certs gives us the self-signed error:

```
code42 audit-logs send-to localhost:9999 -b 1d -p TLS-TCP 
Error: Unable to connect to localhost:9999. Failed with error: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self signed certificate (_ssl.c:1122).
```

Providing the same cert ncat is using works: 

```
code42 audit-logs send-to localhost:9999 -b 1d -p TLS-TCP --certs cert.pem
```

unparalleled-js: heck yeah, thanks for these instructions

alanag13:
> When making the latest changes, I noticed that Alerts send_to defaults to JSON while the securitydata and auditlogs default to RAW-JSON.

should we update Alerts to use `RAW-JSON` by default? This wouldn't be a breaking change, and I think fields being null on alerts is pretty rare.

timabrmsn:
It looks like we don't error if you try to use `-p TCP` when sending to a TLS server, the command appears to run ok, but nothing comes through. I wonder if we can catch that and raise a CLIError somehow?

unparalleled-js:
> It looks like we don't error if you try to use `-p TCP` when sending to a TLS server, the command appears to run ok, but nothing comes through. I wonder if we can catch that and raise a CLIError somehow?

Huh, that is interesting point. Can servers be equipped to handle both? Will there always be an error?

Maybe there is a way to force an error

unparalleled-js:
> It looks like we don't error if you try to use `-p TCP` when sending to a TLS server, the command appears to run ok, but nothing comes through. I wonder if we can catch that and raise a CLIError somehow?

It also permits UDP

unparalleled-js:
> It looks like we don't error if you try to use `-p TCP` when sending to a TLS server, the command appears to run ok, but nothing comes through. I wonder if we can catch that and raise a CLIError somehow?

From https://docs.python.org/3/library/socket.html#socket.socket.sendall

`On error, an exception is raised, and there is no way to determine how much data, if any, was successfully sent.`

timabrmsn:
Since TLS is over TCP the initialization of the TCP connection works, but a server listening with TLS is going to respond initially with a TLS handshake message that the client would negotiate before actually sending the payload data. We would probably need to check for any response message from the server and if it's a handshake, raise an error. 

UDP as a protocol is designed for firing and forgetting, not getting any acknowledgement if it was successful or not, so there's nothing we're going to be able to do there.

timabrmsn:
Actually, scratch that. The server doesn't return a handshake response unless the client sends one. It should close the TCP connection if the data the client sends _isn't_ a TLS handshake initiation... Poking at some things to see if there's an easy way to check/handle that.

timabrmsn:
Ah, it actually _is_ erroring now, we're just eating the error output in the logger here: https://github.com/code42/code42cli/blob/feature/ssl-sendto/src/code42cli/logger/__init__.py#L19

If you flip that to True and try to send `-P TCP` to a TLS server it spews a ton of stack traces to stderr. 

I did define this to correctly raise a click exception on a network connection break: https://github.com/code42/code42cli/blob/feature/ssl-sendto/src/code42cli/logger/__init__.py#L40-L47

But it looks like I might not have ever actually put that in place on the logger itself. If we use that logic here it should work: https://github.com/code42/code42cli/blob/feature/ssl-sendto/src/code42cli/logger/handlers.py#L72

src/code42cli/logger/handlers.py
```diff
@@ -0,0 +1,145 @@
...
+            self.handleError(record)
+
+    def handleError(self, record):
+        """Override logger's `handleError` method to exit if an exception is raised while trying to
+        log, otherwise it would continue to gather and process events if the connection breaks but send
+        them nowhere.
+        """
+        t, _, _ = sys.exc_info()
+        if t == BrokenPipeError:
+            raise SyslogServerNetworkConnectionError()
```
unparalleled-js, line 89: To get `Code42CLIError` to work, we raise a custom exception here, and that gets caught in `click_ext.groups.py` line 61, which then gets converted into a `Code42CLIError`. The reason for this is because we are unable to import `Code42CLIError` here because of a circular dependency.  Before, it was raising a `ClickException`, but that text is not red.

timabrmsn (approved):

PR #193, opened 2020-12-30
fix nightly test runs

timabrmsn:


kiran-chaudhary (approved):

tox.ini
```diff
@@ -43,9 +43,12 @@ deps =
     pytest == 4.6.11
     pytest-mock == 2.0.0
     pytest-cov == 2.10.0
+    pexpect == 4.8.0
     git+https://github.com/code42/py42.git@master#egg=py42
     git+ssh://git@github.com/code42/c42eventextractor.git@master#egg=c42eventextractor
 
+commands = pytest --cov=code42cli --cov-report xml -v -rsxX -l --tb=short --strict -m "not integration"
```
kiran-chaudhary, line 50: this line already exists at line 15.

kiran-chaudhary (changes requested):
Need to remove the duplicate line.

kiran-chaudhary (approved):

kiran-chaudhary:
I think the stop testing 3.5 hasn't worked quite expected. As such that was already introduced by my changes so `expect` module changes seem fine. However it would be better if we remove it from automation pipeline.

timabrmsn:
Yeah, I figured we should make that it's own separate PR, since we have to remove it from the github config too.

PR #196, opened 2021-01-04
Faster test run

unparalleled-js:
Make the tox unit tests command faster
Adds section in contributing for how to run integration tests:

```pytest -m "integration"```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -124,6 +124,12 @@ This will also test that the documentation build passes and run the style checks
 $ tox -e py
 ```
 
+If you want to run the integration tests in your current python environment, you can do:
+
+```bash
+pytest -m "integration"
+```
+
 ### Writing tests
 
 Put actual before expected values in assert statements. Pytest assumes this order.
--- a/tox.ini
+++ b/tox.ini
@@ -19,7 +19,7 @@ commands =
     # -l: show locals in tracebacks
     # --tb=short: short traceback print mode
     # --strict: marks not registered in configuration file raise errors
-    pytest --cov=code42cli --cov-report xml -v -rsxX -l --tb=short --strict -m --ignore=integration
+    pytest --cov=code42cli --cov-report xml -v -rsxX -l --tb=short --strict --ignore=integration
 
 [testenv:docs]
 deps =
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: CONTRIBUTING.md [hunk 1]
```diff
@@ -124,6 +124,12 @@ This will also test that the documentation build passes and run the style checks
 $ tox -e py
 ```
 
+If you want to run the integration tests in your current python environment, you can do:
+
+```bash
+pytest -m "integration"
```
unparalleled-js, line 130: This is the benefit of marking the tests as integration

tox.ini
```diff
@@ -19,7 +19,7 @@ commands =
     # -l: show locals in tracebacks
     # --tb=short: short traceback print mode
     # --strict: marks not registered in configuration file raise errors
-    pytest --cov=code42cli --cov-report xml -v -rsxX -l --tb=short --strict -m "not integration"
+    pytest --cov=code42cli --cov-report xml -v -rsxX -l --tb=short --strict --ignore=integration
```
unparalleled-js, line 22: This is the benefit of _still_ keeping the integration tests in their own directory... Ignoring a whole directory is significantly faster than ignoring individual tests with a `pytest mark`.

CONTRIBUTING.md
kiran-chaudhary: We might want to add that to run integration tests "Set environment variables C42_USERNAME and C42_PW with CCA credentials".

tox.ini
kiran-chaudhary: This is nice, as I marked the individual test class as `integration`, some of the negative tests which were kind of unit tests still exists in `integration` directory, we might want to move them back to appropriate unit test files.

I can create a pull request against this branch with the changes and then we can merge this to master to make sure all unit tests are executed.

kiran-chaudhary (commented):
We need some more refactoring to get this work as desired.

CONTRIBUTING.md
kiran-chaudhary: I added this information in `py42` repo but somehow missed it here.

unparalleled-js:
@kiran-chaudhary Yeah, unit tests should probably not be in the integration test folder!  Thanks, you could also just open that PR against Master.

kiran-chaudhary (approved):
LGTM, as discussed I will pull in another PR to move tests to their appropriate locations.

PR #198, opened 2021-01-07
Refactor magic date parsing & some search options

timabrmsn:
This PR moves our "magic" date parsing we use for --begin/--end options out from callback functions which are very specific to existing commands (specific rounding, returns a timestamp instead of datetime, limiting date range, etc.) and makes more flexible to be used elsewhere. 

Now a `click.option` can take `code42cli.click_ext.types.MagicDate` as the `type` param (with an optional rounding function in the ctor) and get a datetime object back. If additional manipulation needs to be done on the result (like converting to timestamp), a callback function can still be used. 

It also pulls the "search" options back out into individual functions to allow for easier re-use and flexibility.


src/code42cli/click_ext/types.py
```diff
@@ -16,3 +20,98 @@ def convert(self, value, param, ctx):
...
+
+    timestamp formats:
+        yyyy-MM-dd
+        yyyy-MM-dd HH
+        yyyy-MM-dd HH:MM
+        yyyy-MM-dd HH:MM:SS
+
+    short-string (day, hour, min) formats:
+        30d
+        24h
```
unparalleled-js, line 39: Do we also want to support 24H? (case-insensitve)

timabrmsn: yeah, that's simple enough to include

src/code42cli/cmds/securitydata.py
```diff
@@ -160,6 +156,26 @@ def _get_saved_search_query(ctx, param, arg):
     cls=incompatible_with("advanced_query"),
 )
 
+begin_option = opt.begin_option(
+    SECURITY_DATA_KEYWORD,
+    callback=lambda ctx, param, arg: convert_datetime_to_timestamp(
+        limit_date_range(arg, max_days_back=90)
```
unparalleled-js, line 162: `90` should probably be a constant defined somewhere so it's easy to change

src/code42cli/click_ext/types.py
```diff
@@ -16,3 +20,98 @@ def convert(self, value, param, ctx):
...
+    """
+
+    TIMESTAMP_REGEX = re.compile(r"(\d{4}-\d{2}-\d{2})\s*(.*)?")
+    MAGIC_TIME_REGEX = re.compile(r"(\d+)([dhm])$")
+    HELP_TEXT = (
+        "Accepts a date/time in yyyy-MM-dd (UTC) or yyyy-MM-dd HH:MM:SS "
+        "(UTC+24-hr time) format where the 'time' portion of the string "
+        "can be partial (e.g. '2020-01-01 12' or '2020-01-01 01:15') or "
+        "a 'short time' value representing days (30d), hours (24h) or "
+        "minutes (15m) from current time."
```
unparalleled-js, line 52: from __the__ current time

src/code42cli/click_ext/types.py
unparalleled-js: I think you need to include it in the regex somehow for it to work

src/code42cli/click_ext/types.py
```diff
@@ -16,3 +20,99 @@ def convert(self, value, param, ctx):
...
+    name = "magicdate"
+
+    def __init__(self, rounding_func=None):
+        self.round = rounding_func
+
+    def get_metavar(self, param):
+        return "[DATE|TIMESTAMP|SHORT_TIME]"
+
+    def __repr__(self):
+        return "MagicDate"
```
unparalleled-js, line 64: Does this have a CLI use?  It seems to not offer much context, less than not defining a `__repr__` even because it is might easy to miscontrue with the repr of the str `MagicDate`

src/code42cli/cmds/auditlogs.py
```diff
@@ -31,7 +29,14 @@
 AUDIT_LOGS_DEFAULT_HEADER["userName"] = "AffectedUser"
 AUDIT_LOGS_DEFAULT_HEADER["userId"] = "AffectedUserUID"
 
-
+begin_option = opt.begin_option(
```
unparalleled-js, line 32: Seems like a possibility of re-using code, some parameterized method for creating these two options

src/code42cli/cmds/devices.py
```diff
@@ -297,6 +297,10 @@ def handle_row(guid):
         return device_dataframe
 
 
+def _filter_dataframe_by_date_column(df, column, start, end):
```
unparalleled-js, line 300: This method does not filter anything

src/code42cli/date_helper.py
```diff
@@ -15,90 +14,36 @@
...
-def verify_timestamp_order(min_timestamp, max_timestamp):
-    if min_timestamp is None or max_timestamp is None:
+def convert_datetime_to_timestamp(dt):
+    if dt is None:
         return
-    if min_timestamp >= max_timestamp:
-        raise click.BadParameter(
-            param_hint=["-b", "--begin"], message="cannot be after --end date."
-        )
+    return (dt - datetime.utcfromtimestamp(0)).total_seconds()
```
unparalleled-js, line 20: This is like the python2 compat way of doing this, not that it is invalid.

We can make use of the python3 feature and just do `dt.timestamp()`, I think

unparalleled-js:
```
>>> n.timestamp()
1610037626.854844
>>> (n - datetime.utcfromtimestamp(0)).total_seconds()
1610037626.854844
>>> 
```

unparalleled-js:
It would see nice to see a CHANGELOG entry under Changed - `Improved verbiage around date arguments`, I think those kind of updates look good in a changelog because it makes it evident we care about UX

tests/test_magic_date_type.py
```diff
@@ -0,0 +1,113 @@
...
+one_ms = timedelta(milliseconds=1)
+
+
+class TestMagicDateNoRounding:
+    md = MagicDate()
+
+    def convert(self, val):
+        return self.md.convert(val, ctx=None, param=None)
+
+    def test_when_given_date_str_parses_successfully(self):
```
unparalleled-js, line 22: glaring lack of negative tests in this whole file

src/code42cli/cmds/auditlogs.py
timabrmsn: That's kind of how it was before, but it was only re-usable in a specific way. I'd rather have there be a tiny bit of duplication in each command that uses these options and have them be flexible enough to not have to rewrite everything if a new command needs a slightly different configuration.

src/code42cli/cmds/devices.py
timabrmsn: yeah, that was an inadvertent commit from another in-progress branch.

src/code42cli/date_helper.py
```diff
@@ -18,7 +18,7 @@
 def convert_datetime_to_timestamp(dt):
     if dt is None:
         return
-    utc_epoch = datetime.utcfromtimestamp(0).astimezone(timezone.utc)
+    utc_epoch = datetime.utcfromtimestamp(0).replace(tzinfo=timezone.utc)
```
unparalleled-js, line 21: I am surprised the tzinfo when doing utcfromtimestamp is not already set to UTC

timabrmsn: Yeah, it returns a naive dt object just with what the UTC value would be from epoch 0. The value doesn't ultimately change, but forcing the tzinfo on the object makes sure we don't make a wrong assumption later when interacting with it.

src/code42cli/click_ext/types.py
alanag13: repr determines some stuff that shows up in debuggers when hovering over stuff while you've hit breakpoint, which can be helpful. In this case though since what's returned is just the name of the class, I'm not sure it's needed.

src/code42cli/cmds/auditlogs.py
alanag13: agree with tim, code should only be reused if its beleived that the functionality will _always_ stay the same, not just because it is _currently_ the same.

tests/test_magic_date_type.py
alanag13: if we add case insensitivity, we should have tests for that too.
There should also be at least one test for each rounding strategy that verifies that this fails under conditions where it's supposed to

src/code42cli/click_ext/types.py
unparalleled-js: Ah, I see. If it makes debugging look cleaner, than it makes sense I suppose... The default repr just includes a mem address which may seem noisy

timabrmsn: I just copied the pattern from an existing click.ParamType subclass, I'm not exactly sure when __repr__ will be useful, but it can't hurt.

src/code42cli/cmds/securitydata.py
timabrmsn: I don't know if adding another variable makes it any easier to change, the first place I'd go to look for the config for this option would be at the option definition, making it a constant somewhere else just means I have to make one more hop to see what it is/change it.

tests/test_magic_date_type.py
timabrmsn: Good call, as adding the negative tests did catch some cases we weren't handling correctly. I've tweaked the regex to be a bit tighter around what we're grabbing, and also made it so we accept `2020-01-01T12:00:00` as a valid format.

src/code42cli/date_helper.py
timabrmsn: updated.

alanag13 (approved):

PR #199, opened 2021-01-08
Feature cases

kiran-chaudhary:
**Description**
Added `cases` command  to CLI.

Note: This will remain draft PR until `cases` is released in `py42`.

**Issues Resolved**

- closes INTEG-1192

Added following commands:
- `code42 cases`  with subcommands
    - `create` , `update`,  `export` ,  `list` , `show` 

- `code42 cases file-events` commands with subcommands
    - `add` ,   `remove` ,   `list` ,  `show` 
    
 **PR Checklist**

[ Y] Add unit tests to verify this change
[ Y] Add an entry to CHANGELOG.md describing this change
[ Y] Add docstrings.


src/code42cli/cmds/cases.py
```diff
@@ -0,0 +1,246 @@
...
+from py42.exceptions import Py42BadRequestError
+from py42.exceptions import Py42NotFoundError
+
+from code42cli.click_ext.groups import OrderedGroup
+from code42cli.options import format_option
+from code42cli.options import sdk_options
+from code42cli.output_formats import OutputFormatter
+
+
+_HEADER_KEYS_MAP = OrderedDict()
```
unparalleled-js, line 13: `OrderedDict` is no longer needed in the CLI, from the removal of py3.5 support.

unparalleled-js: I've also starting favoring a helper method to create the dict rather than defining it at root level, but that is not a big deal

src/code42cli/cmds/cases.py
```diff
@@ -0,0 +1,246 @@
...
+_HEADER_KEYS_MAP["name"] = "Name"
+_HEADER_KEYS_MAP["assignee"] = "Assignee"
+_HEADER_KEYS_MAP["status"] = "Status"
+_HEADER_KEYS_MAP["createdAt"] = "Creation Time"
+_HEADER_KEYS_MAP["findings"] = "Notes"
+
+name_option = click.option(
+    "--name", required=False, help="Name of the case.", default=None
```
unparalleled-js, line 21: `required=False` is not needed, as click options default to being optional.

unparalleled-js: I believe `default=None` is also not needed

src/code42cli/cmds/cases.py
```diff
@@ -0,0 +1,246 @@
...
+from code42cli.options import sdk_options
+from code42cli.output_formats import OutputFormatter
+
+
+_HEADER_KEYS_MAP = OrderedDict()
+_HEADER_KEYS_MAP["name"] = "Name"
+_HEADER_KEYS_MAP["assignee"] = "Assignee"
+_HEADER_KEYS_MAP["status"] = "Status"
+_HEADER_KEYS_MAP["createdAt"] = "Creation Time"
+_HEADER_KEYS_MAP["findings"] = "Notes"
```
unparalleled-js, line 18: The `show` command required a case number.
For that reason, we should include the case number in the `list` output, so that I can find a case when calling list and then copy the case number for using `show`.

src/code42cli/cmds/cases.py
```diff
@@ -0,0 +1,246 @@
...
+description_option = click.option(
+    "--description", required=False, help="Description of the case.", default=None
+)
+notes_option = click.option(
+    "--notes", required=False, help="Notes on the case.", default=None
+)
+subject_option = click.option(
+    "--subject", required=False, help="User UID of a subject of the case.", default=None
+)
+status_option = click.option(
```
unparalleled-js, line 35: The status option should probably include `type=click.Choice(CaseStatus.choices()),` (after importing the enum)

src/code42cli/cmds/cases.py
```diff
@@ -0,0 +1,246 @@
...
+
+@click.group(cls=OrderedGroup)
+@sdk_options(hidden=True)
+def cases(state):
+    """For managing cases and events associated with cases."""
+    pass
+
+
+@cases.command()
+@click.argument("name", type=str)
```
unparalleled-js, line 56: `type=str` is not needed as that is the default

src/code42cli/cmds/cases.py
```diff
@@ -0,0 +1,246 @@
...
+def create(state, name, subject, assignee, description, notes):
+    """Create a new case."""
+    response = state.sdk.cases.create(
+        name,
+        subject=subject,
+        assignee=assignee,
+        description=description,
+        findings=notes,
+    )
+    return response
```
unparalleled-js, line 71: I don't think returning the response is necessary

src/code42cli/cmds/cases.py
```diff
@@ -0,0 +1,246 @@
...
+    return cases
+
+
+@cases.command("list")
+@click.option(
+    "--name",
+    required=False,
+    help="Filter by name of a case, supports partial name matches.",
+)
+@click.option("--assignee", required=False, help="Filter by user UID of assignee.")
```
unparalleled-js, line 109: Curious why this command does not use the options defined at the top of the file.

The only difference I can tell is the different help text.

Perhaps it would be better to have helper methods that accept a `help` argument and creates the options accordingly.

src/code42cli/cmds/cases.py
```diff
@@ -0,0 +1,246 @@
...
+        assignee=assignee,
+        description=description,
+        findings=notes,
+        status=status,
+    )
+
+
+def _parse_cases(pages):
+    cases = [case for page in pages for case in page["cases"]]
+    if not cases:
```
unparalleled-js, line 98: echoing `No cases found.` does not seem like a responsibility of a method named (`_parse_cases()`).

I was confused when looking at the command because I could not find where the `click.echo("No cases found.")` was happening

src/code42cli/cmds/cases.py
```diff
@@ -0,0 +1,246 @@
...
+    required=False,
+    help="Fetch cases created after given date time.",
+)
+@click.option(
+    "--max-create-time",
+    required=False,
+    help="Fetch cases created before given date time.",
+)
+@click.option(
+    "--min-update-time",
```
unparalleled-js, line 125: `min` and `max` times should be `--begin` (or `-b`)  and `--end` (`-e`) and use the familiar options, to be like the other options in the Code42 CLI.

src/code42cli/cmds/cases.py
```diff
@@ -0,0 +1,246 @@
...
+@cases.command()
+@case_number_arg
+@click.option(
+    "--path", type=str, help="File path. Defaults to current directory", default="."
+)
+@sdk_options()
+def export(state, case_number, path):
+    """Download case detail summary in a pdf file at the given path with name <case_number>_case_summary.pdf."""
+    response = state.sdk.cases.export_summary(case_number)
+    with open("{}/{}_case_summary.pdf".format(path, case_number), "wb") as f:
```
unparalleled-js, line 183: When running the tests, this actually does file IO and creates a PDF. File IO should be avoided in unit tests.

What you can do is move this to helper method somewhere, like in util, perhaps something like `save_file_as()` that takes the destination and the content. THEN, you can mock that util method using `mocker.patch` in the unit tests.

This will both prevent the file from being created as well as allow you to do mock verifications on the util method.

src/code42cli/cmds/cases.py
```diff
@@ -0,0 +1,246 @@
...
+
+@file_events.command()
+@case_number_arg
+@file_event_id_option
+@sdk_options()
+def add(state, case_number, event_id):
+    """Associate an event id to the case."""
+    try:
+        state.sdk.cases.file_events.add_event(case_number, event_id)
+    except Py42BadRequestError as err:
```
unparalleled-js, line 231: This error handling logic can be shared code. It is used multiple places.

src/code42cli/cmds/cases.py
```diff
@@ -0,0 +1,246 @@
...
+    if cases:
+        formatter.echo_formatted_list(cases)
+
+
+@cases.command()
+@case_number_arg
+@sdk_options()
+def show(state, case_number):
+    """Show case details."""
+    click.echo(state.sdk.cases.get_case(case_number))
```
unparalleled-js, line 171: The method `get_case()` was renamed to just `get()`

src/code42cli/cmds/cases.py
```diff
@@ -0,0 +1,246 @@
...
+    formatter = OutputFormatter(format, _HEADER_KEYS_MAP)
+    cases = _parse_cases(pages)
+    if cases:
+        formatter.echo_formatted_list(cases)
+
+
+@cases.command()
+@case_number_arg
+@sdk_options()
+def show(state, case_number):
```
unparalleled-js, line 169: `show` commands are supposed to be more table like, not just echoing JSON.
See other show commands in the CLI as an example

src/code42cli/cmds/cases.py
```diff
@@ -0,0 +1,246 @@
...
+
+
+@file_events.command("show")
+@case_number_arg
+@file_event_id_option
+@sdk_options()
+def _show(state, case_number, event_id):
+    """Show event details for the given event id associated with the case."""
+    try:
+        click.echo(state.sdk.cases.file_events.get_event(case_number, event_id))
```
unparalleled-js, line 217: Show command are supposed to be more table like, and not just echoing JSON. See other show commands in the CLI.

unparalleled-js, line 217: `get_event()` was renamed to just `get()`

tests/cmds/test_cases.py
```diff
@@ -0,0 +1,74 @@
+from code42cli.main import cli
+
+
+def test_create_case_calls_create_with_expected_params(runner, cli_state):
+    runner.invoke(
```
unparalleled-js, line 5: For the `create_` and the `update_` tests, can you use all the options so that we can have the tests verify that each of the options gets passed into py42?

tests/cmds/test_cases.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+def test_show_calls_get_case_with_expected_params(runner, cli_state):
+    runner.invoke(
+        cli, ["cases", "show", "1"], obj=cli_state,
+    )
+    cli_state.sdk.cases.get_case.assert_called_once_with(1)
+
+
+def test_export_calls_export_summary_with_expected_params(runner, cli_state):
```
unparalleled-js, line 42: As mentioned in another comment, this test creates a file and that should be mocked

tests/cmds/test_cases.py
```diff
@@ -0,0 +1,74 @@
...
+        name="TEST_CASE2",
+        status=None,
+        assignee=None,
+        description=None,
+        findings=None,
+        subject=None,
+    )
+
+
+def test_list_calls_get_all_with_expected_params(runner, cli_state):
```
unparalleled-js, line 28: None of these test are verifying output.  We should have some tests that verify some data is present in the show and list commands via `res.output` (if you set `res =` the runner invocation.

See other tests in the CLI as an example.

src/code42cli/cmds/cases.py
```diff
@@ -0,0 +1,246 @@
...
+        click.echo("Invalid case-number or event-id.")
+        raise err
+
+
+@file_events.command()
+@case_number_arg
+@file_event_id_option
+@sdk_options()
+def add(state, case_number, event_id):
+    """Associate an event id to the case."""
```
unparalleled-js, line 228: to `a` case instead of `the`

src/code42cli/cmds/cases.py
kiran-chaudhary: Yes, its due to different help message. The one defined at top is used in multiple places whereas this isn't.

src/code42cli/cmds/cases.py
kiran-chaudhary: changed.

src/code42cli/cmds/cases.py
kiran-chaudhary: The errors are different, and no two methods are exactly identical even if we ignore the `state.sdk....` statement.

tests/cmds/test_cases.py
kiran-chaudhary: Done.

src/code42cli/cmds/cases.py
kiran-chaudhary: Changed.

src/code42cli/cmds/cases.py
kiran-chaudhary: added.

src/code42cli/cmds/cases.py
kiran-chaudhary: the tests are failing after making this change, locally executing `pytest` passes the tests whereas `tox` command fails.

unparalleled-js: Why does it fail?

kiran-chaudhary: it cannot load module `py42.clients.cases` from the statement `from py42.clients.cases import CaseStatus`

src/code42cli/cmds/cases.py
```diff
@@ -0,0 +1,241 @@
...
+
+case_number_arg = click.argument("case-number", type=int)
+name_option = click.option("--name", help="Name of the case.",)
+assignee_option = click.option("--assignee", help="User UID of the assignee.")
+description_option = click.option("--description", help="Description of the case.")
+notes_option = click.option("--notes", help="Notes on the case.")
+subject_option = click.option("--subject", help="User UID of a subject of the case.")
+status_option = click.option(
+    "--status",
+    help="Status of the case. `OPEN` or `CLOSED`",
```
unparalleled-js, line 20: Missing period

src/code42cli/cmds/cases.py
unparalleled-js: It is still doing it.

src/code42cli/cmds/cases.py
```diff
@@ -0,0 +1,241 @@
...
+    type=click.Choice(CaseStatus.choices()),
+)
+file_event_id_option = click.option(
+    "--event-id", required=True, help="File event id associated to the case."
+)
+
+
+def _get_cases_header():
+    return {
+        "number": "number",
```
unparalleled-js, line 30: Can we add case number here?

I thought I made this comment already but can't find it.

src/code42cli/cmds/cases.py
```diff
@@ -0,0 +1,241 @@
...
+        formatter.echo_formatted_list(cases)
+    else:
+        click.echo("No cases found.")
+
+
+@cases.command()
+@case_number_arg
+@sdk_options()
+@format_option
+def show(state, case_number, format):
```
unparalleled-js, line 153: I don't see the point in this command, it does not have any additional information.

I do not think we need a show command -- all of the data is already on `list`.

unparalleled-js: Unlesssssss we have a `--include-events` flags and show the events associated with it too

unparalleled-js: Similar to `code42 legal-hold show <id> --include-policy`

The events data can be JSON, as the policy in the legal hold command is also JSON.

So the small table and then the a header and then the json if want to include the event

src/code42cli/cmds/cases.py
```diff
@@ -0,0 +1,241 @@
...
+        findings=notes,
+        status=status,
+    )
+
+
+@cases.command("list")
+@click.option(
+    "--name", help="Filter by name of a case, supports partial name matches.",
+)
+@click.option("--assignee", required=False, help="Filter by user UID of assignee.")
```
unparalleled-js, line 100: `required=False` not needed

tests/cmds/test_cases.py
```diff
@@ -0,0 +1,232 @@
...
+
+
+def test_file_events_show_calls_get_event_with_expected_params(runner, cli_state):
+    runner.invoke(
+        cli, ["cases", "file-events", "show", "1", "--event-id", "1"], obj=cli_state,
+    )
+    cli_state.sdk.cases.file_events.get.assert_called_once_with(1, "1")
+
+
+def test_file_events_show_returns_error_message_when_py42_raises_exception(
```
unparalleled-js, line 171: The `when` clauses should be in the middle

`def test_<method>_<condition>_<expectation>()`

tests/cmds/test_cases.py
```diff
@@ -0,0 +1,232 @@
...
+    py42_response.data = json.loads(EVENT_DETAILS)
+    cli_state.sdk.cases.file_events.get.return_value = py42_response
+    result = runner.invoke(cli, ["cases", "file-events", "show", "1", "--event-id", "e"], obj=cli_state,)
+    assert "0_1d71796f-af5b-4231-9d8e-df6434da4663_984418168383179707_986472527798692818_971" in result.output
+
+
+def test_events_list_returns_expected_data(runner, cli_state):
+    cli_state.sdk.cases.file_events.get_all.return_value = json.loads(ALL_EVENTS)
+    result = runner.invoke(cli, ["cases", "file-events", "list", "1"], obj=cli_state,)
+    assert "0_1d71796f-af5b-4231-9d8e-df6434da4663_984418168383179707_986472527798692818_971" in result.output
```
unparalleled-js, line 232: Missing newline at the end of this file.

tests/cmds/test_cases.py
kiran-chaudhary: Added new tests to cover that.

src/code42cli/cmds/cases.py
kiran-chaudhary: Added.

src/code42cli/cmds/cases.py
kiran-chaudhary: Number above refers to `case number`, do you want me to rename it?

src/code42cli/cmds/cases.py
kiran-chaudhary: The json response does has more data, but I was not sure which fields are guaranteed and which to display so for `table` format, it is similar to `lists` command output, but for other formats it will contain more data.

kiran-chaudhary: In list command, we can make `--file-events` option optional or rename it appropriately as suggested, but in either case we need to decide which fields are important and should be displayed in the table.

src/code42cli/cmds/cases.py
kiran-chaudhary: Removed.

tests/cmds/test_cases.py
kiran-chaudhary: Renamed.

tests/cmds/test_cases.py
kiran-chaudhary: Added.

src/code42cli/cmds/cases.py
kiran-chaudhary: As per my understanding, `tox` fetches released version of `py42` and hence cannot load the module, whereas since I have configured `py42` to point to local `py42` module, `pytest` is able to load the module.  So once `cases` is released in `py42` this might get fixed.

src/code42cli/cmds/cases.py
unparalleled-js: No I made this comment before doing re-checking out the branch after your last force-commit

src/code42cli/cmds/cases.py
```diff
@@ -0,0 +1,250 @@
...
+@cases.command()
+@case_number_arg
+@click.option(
+    "--path", type=str, help="File path. Defaults to current directory.", default="."
+)
+@sdk_options()
+def export(state, case_number, path):
+    """Download case detail summary in a pdf file at the given path with name <case_number>_case_summary.pdf."""
+    response = state.sdk.cases.export_summary(case_number)
+    with open("{}/{}_case_summary.pdf".format(path, case_number), "wb") as f:
```
unparalleled-js, line 181: On a `Windows` machine, using a path with the other slashes `\`, will this work?

kiran-chaudhary: also should we check whether the last character is a path separator or not, a user may or may not pass it!

src/code42cli/cmds/cases.py
unparalleled-js: Right, it will get fixed when we release py42

src/code42cli/cmds/cases.py
unparalleled-js: Perhaps the file events JSON is too extreme (too large and noisy). Nevermind that! However, looking at the Ticket item for this: 

`
option exists for returning the full event data for events in the case
`

we should include all the data.  It is just 1 case. So, either:

1.) Create a table with all the data and then
1a.) Ignore capitalizing the header because we don't all the keys ahead of time and we have said in that past that is it not super important. There are even benefits to having the legitimate keys as header items as well.
1b.) If an expected header item is not on the response, just assume it is `None` and echo `None` for the row prop.

or

2.) Echo the full JSON blob. I'd go with this option if the JSON is very big (which I don't think it is) and ends up looking really bad as an ascii table

src/code42cli/cmds/cases.py
unparalleled-js: I think if you do `os.path.join()`, it will automatically take care of that.

src/code42cli/cmds/cases.py
```diff
@@ -0,0 +1,240 @@
...
+    )
+
+
+@cases.command("list")
+@click.option(
+    "--name", help="Filter by name of a case, supports partial name matches.",
+)
+@click.option("--subject", help="Filter by user UID of the subject of a case.")
+@click.option("--assignee", help="Filter by user UID of assignee.")
+@click.option("--begin-create-time", **BEGIN_DATE_DICT)
```
kiran-chaudhary, line 112: Do we need to support --checkpoint feature here?

kiran-chaudhary: I refactored (not exactly) the "begin option" to be used without extractor!

tests/cmds/test_cases.py
```diff
@@ -0,0 +1,279 @@
...
+
+
+@pytest.mark.parametrize(
+    "command, error_msg",
+    [
+        ("{} create --description d".format(CASES_COMMAND), MISSING_NAME),
+        ("{} update --description d".format(CASES_COMMAND), MISSING_CASE_NUMBER),
+        ("{} show".format(CASES_COMMAND), MISSING_CASE_NUMBER),
+        ("{} export".format(CASES_COMMAND), MISSING_CASE_NUMBER),
+        # ("{} add ".format(CASES_FILE_EVENTS_COMMAND), MISSING_CASE_NUMBER),
```
kiran-chaudhary, line 262: This test and the 2 others commented below, fails with error `Error: Invalid value for 'CASE_NUMBER':  is not a valid integer\n` when executed through `pytest` but on command line the return expected `MISSING_CASE_NUMBER` error message. 

Not sure whether to change the test expectation or keep the test!!

src/code42cli/cmds/cases.py
unparalleled-js: I don't think we need a checkpoint here.

src/code42cli/cmds/cases.py
```diff
@@ -0,0 +1,240 @@
...
+)
+
+CASES_KEYWORD = "cases"
+BEGIN_DATE_DICT = set_begin_default_dict(CASES_KEYWORD)
+END_DATE_DICT = set_end_default_dict(CASES_KEYWORD)
+
+
+def _get_cases_header():
+    return {
+        "number": "number",
```
unparalleled-js, line 40: The rest of these headers are capitalized except for `number`

tests/cmds/test_cases.py
```diff
@@ -0,0 +1,279 @@
...
+
+
+def test_show_returns_expected_data(runner, cli_state, py42_response):
+    py42_response.data = json.loads(CASE_DETAILS)
+    cli_state.sdk.cases.get.return_value = py42_response
+    result = runner.invoke(cli, ["cases", "show", "1"], obj=cli_state,)
+    assert "test@test.test" in result.output
+
+
+def test_list_returns_expected_data(runner, cli_state, py42_response, mocker):
```
unparalleled-js, line 221: mocker is not needed here

tests/cmds/test_cases.py
unparalleled-js: The reason for this is because you are ending these commands with spaces.
When you call `split(" ")`, it has an extra space-arg at the end, which it thinks is the case numbers.  Obviously, empy space is not a number.

Do this: `"{} add".format(CASES_FILE_EVENTS_COMMAND), MISSING_CASE_NUMBER),`
instead of this `"{} add ".format(CASES_FILE_EVENTS_COMMAND), MISSING_CASE_NUMBER),`

tests/cmds/test_cases.py
```diff
@@ -0,0 +1,279 @@
...
+    [
+        ("{} create --description d".format(CASES_COMMAND), MISSING_NAME),
+        ("{} update --description d".format(CASES_COMMAND), MISSING_CASE_NUMBER),
+        ("{} show".format(CASES_COMMAND), MISSING_CASE_NUMBER),
+        ("{} export".format(CASES_COMMAND), MISSING_CASE_NUMBER),
+        # ("{} add ".format(CASES_FILE_EVENTS_COMMAND), MISSING_CASE_NUMBER),
+        ("{} add --event-id 3".format(CASES_FILE_EVENTS_COMMAND), MISSING_CASE_NUMBER),
+        ("{} add 3".format(CASES_FILE_EVENTS_COMMAND), MISSING_EVENT_ID),
+        ("{} remove 3".format(CASES_FILE_EVENTS_COMMAND), MISSING_EVENT_ID),
+        # ("{} remove ".format(CASES_FILE_EVENTS_COMMAND), MISSING_CASE_NUMBER),
```
unparalleled-js, line 266: `("{} remove "` -> `("{} remove"`

tests/cmds/test_cases.py
```diff
@@ -0,0 +1,279 @@
...
+        (
+            "{} remove --event-id 3".format(CASES_FILE_EVENTS_COMMAND),
+            MISSING_CASE_NUMBER,
+        ),
+        # ("{} list ".format(CASES_FILE_EVENTS_COMMAND), MISSING_CASE_NUMBER),
```
unparalleled-js, line 271: `"{} list " ` -> `"{} list"`

tests/cmds/test_cases.py
kiran-chaudhary: Eagle's eye. Thanks!

src/code42cli/cmds/cases.py
kiran-chaudhary: Done.

tests/cmds/test_cases.py
kiran-chaudhary: Removed.

tests/cmds/test_cases.py
kiran-chaudhary: Done.

tests/cmds/test_cases.py
kiran-chaudhary: Done.

src/code42cli/cmds/cases.py
```diff
@@ -0,0 +1,240 @@
...
+        formatter.echo_formatted_list(cases)
+    else:
+        click.echo("No cases found.")
+
+
+def _display_file_events(sdk, case_number):
+    click.echo("\nFile Events:\n")
+    response = sdk.cases.file_events.get_all(case_number)
+    if not response["events"]:
+        click.echo("No events found.")
```
unparalleled-js, line 154: need to return here or else it prints:

```
No events found.
{'events': []}
```

src/code42cli/cmds/cases.py
```diff
@@ -0,0 +1,240 @@
...
+    formatter = OutputFormatter(format, _get_cases_header())
+    cases = [case for page in pages for case in page["cases"]]
+    if cases:
+        formatter.echo_formatted_list(cases)
+    else:
+        click.echo("No cases found.")
+
+
+def _display_file_events(sdk, case_number):
+    click.echo("\nFile Events:\n")
```
unparalleled-js, line 151: I'd print this line after first checking if there are file events.

src/code42cli/cmds/cases.py
```diff
@@ -0,0 +1,240 @@
...
+@case_number_arg
+@file_event_id_option
+@sdk_options()
+def remove(state, case_number, event_id):
+    """Remove the associated event id from the case."""
+    try:
+        state.sdk.cases.file_events.delete(case_number, event_id)
+    except Py42NotFoundError as err:
+        click.echo("Invalid case-number or event-id.")
+        raise err
```
unparalleled-js, line 240: Convert these two lines to just :  `raise Code42CLIError("Invalid case-number or event-id.")`


Two things here:

1.) In general, don't do `raise err`, just do `raise`, it preservers the stack for debuggin is easier.

2.) Instead of printing something and then raising an error, we have a process for raising CLI error so that they show up in red text and get a message printed

src/code42cli/cmds/cases.py
```diff
@@ -0,0 +1,240 @@
...
+@file_events.command()
+@case_number_arg
+@file_event_id_option
+@sdk_options()
+def add(state, case_number, event_id):
+    """Associate an event id to a case."""
+    try:
+        state.sdk.cases.file_events.add(case_number, event_id)
+    except Py42BadRequestError as err:
+        click.echo("Invalid case-number or event-id.")
```
unparalleled-js, line 226: Same here -- changed these two lines to `raise Code42CLIError("Invalid case-number or event-id.")`

src/code42cli/cmds/cases.py
```diff
@@ -0,0 +1,240 @@
...
+@file_events.command("list")
+@case_number_arg
+@sdk_options()
+@format_option
+def file_events_list(state, case_number, format):
+    """List all the events associated with the case."""
+    formatter = OutputFormatter(format, _get_events_header())
+    try:
+        response = state.sdk.cases.file_events.get_all(case_number)
+    except Py42NotFoundError as err:
```
unparalleled-js, line 207: Change this to `"raise Code42CLIError("Invalid case-number.")`"

unparalleled-js: After the custom exceptions merge in py42, you can dome something even easier:  In `click_ext.groups`, line 54ish, there is a list of imported py42 custom exceptions that automatically get raised/ printed... All we need to do is add any cases-related custom exceptions there and you don't need to do any error handling in the command itself.

However, since that is not merged yet, simply raising a Code42CLIError is the same thing.

src/code42cli/cmds/cases.py
```diff
@@ -0,0 +1,240 @@
...
+        if include_file_events:
+            _display_file_events(state.sdk, case_number)
+    except Py42NotFoundError:
+        click.echo("Invalid case-number {}.".format(case_number))
+
+
+@cases.command()
+@case_number_arg
+@click.option(
+    "--path", type=str, help="File path. Defaults to current directory.", default="."
```
unparalleled-js, line 180: type=str is unneeded

src/code42cli/cmds/cases.py
```diff
@@ -0,0 +1,240 @@
...
+)
+@sdk_options()
+def export(state, case_number, path):
+    """Download case detail summary in a pdf file at the given path with name <case_number>_case_summary.pdf."""
```
unparalleled-js, line 184: Download _a_ case detail summary.

_as_ a PDF file

src/code42cli/cmds/cases.py
unparalleled-js, line 180: to _the_ current directory

src/code42cli/cmds/cases.py
```diff
@@ -169,9 +177,10 @@ def show(state, case_number, format, include_file_events):
         response = state.sdk.cases.get(case_number)
         formatter.echo_formatted_list([response.data])
         if include_file_events:
-            _display_file_events(state.sdk, case_number)
+            events = _get_file_events(state.sdk, case_number)
+            _display_file_events(events)
     except Py42NotFoundError:
-        click.echo("Invalid case-number {}.".format(case_number))
+        Code42CLIError("Invalid case-number {}.".format(case_number))
```
unparalleled-js, line 183: Need to `raise`

src/code42cli/cmds/cases.py
```diff
@@ -235,6 +243,5 @@ def remove(state, case_number, event_id):
     """Remove the associated event id from the case."""
     try:
         state.sdk.cases.file_events.delete(case_number, event_id)
-    except Py42NotFoundError as err:
-        click.echo("Invalid case-number or event-id.")
-        raise err
+    except Py42NotFoundError:
+        Code42CLIError("Invalid case-number or event-id.")
```
unparalleled-js, line 247: Missing `raise`

src/code42cli/cmds/cases.py
```diff
@@ -204,9 +213,9 @@ def file_events_list(state, case_number, format):
...
         response = state.sdk.cases.file_events.get_all(case_number)
-    except Py42NotFoundError as err:
-        click.echo("Invalid case-number.")
-        raise err
+    except Py42NotFoundError:
+        raise Code42CLIError("Invalid case-number.")
+
     if not response["events"]:
         click.echo("No events found.")
     for event in response["events"]:
```
unparalleled-js, line 221: You are looping through the events twice here.

You can get rid of the `for` loop
Else we are printing a lot of extra

unparalleled-js: Or perhaps this is supposed to loop through pages, if that is the case, then the for loop is incorrect, the first `event` never gets used..

Do you mean to do this:

```
for page in response:
          events = [event for event in page["events"]]
          formatter.echo_formatted_list(events)
```

unparalleled-js: I recommend IDEs like PyCharm, it points out mistakes like this whereas text editors do not typically

src/code42cli/cmds/cases.py
kiran-chaudhary: Done.

src/code42cli/cmds/cases.py
kiran-chaudhary: Done.

src/code42cli/cmds/cases.py
kiran-chaudhary: Done.

src/code42cli/cmds/cases.py
kiran-chaudhary: Done.

src/code42cli/cmds/cases.py
kiran-chaudhary: Done.

src/code42cli/cmds/cases.py
kiran-chaudhary: Done.

src/code42cli/cmds/cases.py
kiran-chaudhary: Done.

src/code42cli/cmds/cases.py
kiran-chaudhary: Done.

src/code42cli/cmds/cases.py
kiran-chaudhary: Done.

src/code42cli/cmds/cases.py
kiran-chaudhary: Done.

unparalleled-js (approved):

src/code42cli/cmds/cases.py
```diff
@@ -0,0 +1,247 @@
...
+        return None
+    return json.loads(response.text)
+
+
+def _display_file_events(events):
+    if events:
+        click.echo("\nFile Events:\n")
+        click.echo(pformat(events))
+    else:
+        click.echo("No events found.")
```
unparalleled-js, line 163: Could you include a newline here as well, like

`click.echo("\nNo events found.")`

kiran-chaudhary: wondering whether we should use `click.echo()` over specifying "\n", to make it os independent.

PR #203, opened 2021-01-13
Refactor move all integration tests into tests directory

kiran-chaudhary:


tox.ini
```diff
@@ -19,7 +19,7 @@ commands =
     # -l: show locals in tracebacks
     # --tb=short: short traceback print mode
     # --strict: marks not registered in configuration file raise errors
```
unparalleled-js, line 21: seems like we need the comment 

`# --ignore=tests/integration: exclude integration tests`

kiran-chaudhary: added.

unparalleled-js (approved):

PR #204, opened 2021-01-19
Misc `devices` command bugfixes

timabrmsn:
Fixes bugs I found when writing up testing criteria for `devices` command post-merge. 

Removes the `--days-since-last-connected` option and replaces them with some more flexible date filtering options: `--last-connected-after`/`--last-connected-before` and `--created-after`/`--created-before`. 

Replaces `--drop-most-recent` with `--exclude-most-recently-connected` to clarify what is intended.


src/code42cli/cmds/devices.py
```diff
@@ -103,17 +106,28 @@ def _change_device_name(sdk, guid, name):
 
 @devices.command()
 @device_guid_argument
-@format_option
+@click.option(
+    "-f",
+    "--format",
+    type=click.Choice(
+        [OutputFormat.TABLE, OutputFormat.JSON, OutputFormat.RAW], case_sensitive=False
```
timabrmsn, line 113: CSV format doesn't make sense for `devices show`, so needed to tweak choices available.

ceciliastevens:
Would it be possible to keep the functionality of `--drop-most-recent` in some fashion? It's not directly replicated by allowing filtering before/after certain connect dates, and it is a commonly used feature in the old standalone device deactivation script that we are hoping the devices command could replace.

src/code42cli/cmds/devices.py
```diff
@@ -145,13 +159,16 @@ def _get_device_info(sdk, device_guid):
...
     default=None,
-    help="Get only active or deactivated devices. Defaults to getting all devices.",
 )
-
+inactive_option = click.option(
+    "--inactive",
+    is_flag=True,
+    help="Limits results to only deactivated devices.",
+    cls=incompatible_with("active"),
+)
```
timabrmsn, line 171: Since the `active` param actually takes 3 states (None, True, False) the boolean option type didn't work.

src/code42cli/output_formats.py
```diff
@@ -88,10 +88,24 @@ def __init__(self, output_format):
             self._format_func = DataFrame.to_csv
         elif output_format == OutputFormat.RAW:
             self._format_func = DataFrame.to_json
-            self._output_args.update({"orient": "records", "lines": False})
+            self._output_args.update(
+                {
+                    "orient": "records",
+                    "lines": False,
+                    "index": True,
```
timabrmsn, line 95: `index: False` throws an error when trying to export a dataframe to json.

src/code42cli/output_formats.py
```diff
@@ -88,10 +88,24 @@ def __init__(self, output_format):
...
+                    "default_handler": str,
```
timabrmsn, line 96: Without this, we were getting an error when trying to convert the backup set info to json, I think because it was trying to traverse all the dynamic attributes of the `DeviceSettings` object in some weird way and hit the recursion limit.

src/code42cli/cmds/devices.py
unparalleled-js: Typically, for show commands, we were just using Table

src/code42cli/cmds/devices.py
```diff
@@ -206,51 +210,78 @@ def _get_device_info(sdk, device_guid):
     is_flag=True,
     help="""Include device settings in output.""",
 )
+@click.option(
+    "--last-connected-before",
+    type=MagicDate(rounding_func=round_datetime_to_day_start),
+    help=f"Include devices only when the 'lastConnected' field is after provided value. {MagicDate.HELP_TEXT}",
```
unparalleled-js, line 216: For these helps, should be "is after _the_ provided value" , missing the's

timabrmsn:
> Would it be possible to keep the functionality of `--drop-most-recent` in some fashion? It's not directly replicated by allowing filtering before/after certain connect dates, and it is a commonly used feature in the old standalone device deactivation script that we are hoping the devices command could replace.

I removed it because it felt _very_ specific to a single targeted use-case, and could potentially cause confusion. I tried to think of a way to have other more general and explicit options to accomplish the same thing, but haven't yet come up with something, but I'm open to suggestions. 

I think that functionality would probably be better suited to a standalone script that just takes a csv and filters out most recent N devices, so if you have a customer who needs to purge old devices (but keep at least 1), they could do something like: 

```
code42 devices list | ./drop_most_recent.py 1 | code42 devices bulk deactivate -
```

src/code42cli/cmds/devices.py
```diff
@@ -206,51 +210,78 @@ def _get_device_info(sdk, device_guid):
...
         "osHostname",
         "status",
         "lastConnected",
+        "creationDate",
         "productVersion",
         "osName",
         "osVersion",
         "userUid",
     ]
-    devices_dataframe = _get_device_dataframe(
```
unparalleled-js, line 235: This table is really large.

I notice that `name` and `osHostname` are often the same thing, one having unicode and one not..

Can we consolidate these?

unparalleled-js: Also, do we need `computerId`?

timabrmsn:
@ceciliastevens is it necessary that the drop-most-recent type filtering have a variable number? What if we did `--find-potential-duplicates` that only showed oldest connection devices for users who have more than 1 device?

src/code42cli/cmds/devices.py
```diff
@@ -414,17 +415,14 @@ def bulk(state):
     pass
 
 
-@bulk.command(
-    name="deactivate",
-    help="""Deactivate all devices on the given list.
-            Takes as input a CSV with a 'guid' column.""",
-)
+@bulk.command(name="deactivate")
 @read_csv_arg(headers=["guid"])
```
unparalleled-js, line 419: I had thought in discussion with Cecilia, we agreed that it is ok to use `computerId` for the bulk commands, since performance is the strategy.

Did we decide device GUID is fine?

src/code42cli/cmds/devices.py
```diff
@@ -414,17 +415,14 @@ def bulk(state):
...
 @change_device_name_option
 @purge_date_option
 @format_option
 @sdk_options()
 def bulk_deactivate(state, csv_rows, change_device_name, purge_date, format):
+    """Deactivate all devices from the provided CSV containing a 'guid' column."""
     sdk = state.sdk
```
unparalleled-js, line 426: when deactivating a guid that does not exist, the error still shows up in the CSV output.
Looks kind of funny:

```
guid    deactivated                                                                                                                     
asdfd   False: Failure in HTTP call 404 Client Error: Not Found for url: https://console.us.code42.com/api/Computer/asdfd?idType=guid   
```

unparalleled-js: I am guessing this is intentional, kind of neat, just a bit different than the other bulk commands

src/code42cli/cmds/devices.py
```diff
@@ -414,17 +415,14 @@ def bulk(state):
...
-            Takes as input a CSV with a 'guid' column.""",
-)
+@bulk.command(name="deactivate")
 @read_csv_arg(headers=["guid"])
 @change_device_name_option
 @purge_date_option
 @format_option
 @sdk_options()
 def bulk_deactivate(state, csv_rows, change_device_name, purge_date, format):
+    """Deactivate all devices from the provided CSV containing a 'guid' column."""
```
unparalleled-js, line 425: If given a CSV file no rows besides a head `guid`, you get a strange error.

src/code42cli/cmds/devices.py
timabrmsn: updated

src/code42cli/cmds/devices.py
timabrmsn: The bulk command just re-used `_deactivate_device()` from the `devices deactivate` command, which was changed to take a guid, so bulk had to be changed too. We could probably follow on with an update after release to allow either the individual or bulk to accept an `--id-type` option so the user can decide.

src/code42cli/cmds/devices.py
ceciliastevens: `name` and `osHostname` are often but not always the same. `name` is a user-editable field, whereas `osHostname` is taken straight from the os. If the user doesn't change the name, then the two are the same, but I think in practice it is important to have both available.

src/code42cli/cmds/devices.py
phirst: I'm not in favor of dropping `--drop-most-recent`.  The use case is that many customers want to be sure that a user always has at least one active device.  There is no way to know the "right" date to pick.  This parameter should stay.

ceciliastevens:
> @ceciliastevens is it necessary that the drop-most-recent type filtering have a variable number? What if we did `--find-potential-duplicates` that only showed oldest connection devices for users who have more than 1 device?

I'll defer to @phirst's comment on this one, I think that a find-duplicates would be neat but it's not exactly the same on drop-most-recent.

src/code42cli/cmds/devices.py
timabrmsn: Ah, true. That simplifies things.

phirst:
I am **NOT** in favor of dropping `--drop-most-recent`.  Customers often want to run the script and want to be sure all users have at least one active device.  A date range won't be useful here.  This is one of the most commonly used flags in the existing script and provides peace of mind to customers running it.  They can then go through and cull out the devices that do need to be deactivated and use a list to clean them up.

phirst:
> I am **NOT** in favor of dropping `--drop-most-recent` variable number. Customers want to be sure all users have at least one active device, but for some their user base may legitimately have 2 devices (desktop and laptop). They want to be sure they're not deactivating devices that legitimately should remain active.

I had to clean up what I wrote after reading it more carefully...

src/code42cli/cmds/devices.py
unparalleled-js: Good to know. I should clarify that I am only referring to the Table output.. For CSV or JSON, all the data should be present.. People should not be using Table for programmatic reasons, so it is good to make it more like a UI summary.

unparalleled-js: To achieve this, all we need is a Header will a predefined list of props to show in the Table to pass to the formatter.  The formatter already knows to include all the data for other formats, iicr

src/code42cli/cmds/devices.py
timabrmsn: Yeah, this was intentional to have the results be output, but I just added some better error handling in `_deactivate_device()` since we weren't catching the 404 correctly. Also validated the guid is an int before making the request since we know it's gonna fail anyway.

src/code42cli/cmds/devices.py
timabrmsn: I updated `read_csv()` to handle this better, in re-writing it I also discovered a few other bugs and it took me awhile to get it all right.

src/code42cli/cmds/devices.py
timabrmsn: Since we're now using the `DataFrameOutputFormatter` to print these results (for Table, JSON, and CSV), the column filtering isn't done based on formatting, but on the `column` list that is passed to `_get_device_dataframe()`. So they're all going to be the same for now.

I think as we refactor other commands to use DataFrames instead of our own formatters, we should come up with an better internal API to tweak which columns are displayed, as well as offering options to the user to pick which ones they want. Hopefully it can be general enough that it can just be added to each command that uses a DataFrame by just wrapping with a decorator.

timabrmsn:
@phirst @ceciliastevens I decided this morning to re-include the drop-most-recent option, but just named a bit more explicitly and with clearer help text.

src/code42cli/cmds/devices.py
```diff
@@ -33,8 +35,7 @@ def devices(state):
     required=False,
     is_flag=True,
     default=False,
-    help="""Prepend "deactivated_" and today's date to the name of any
-    deactivated devices.""",
+    help="""Prepend 'deactivated_<current_date>' to the name of any deactivated devices.""",
```
unparalleled-js, line 38: I don't think the triple quotes are needed

src/code42cli/cmds/devices.py
```diff
@@ -59,8 +60,12 @@ def deactivate(state, device_guid, change_device_name, purge_date):
 
 
 def _deactivate_device(sdk, device_guid, change_device_name, purge_date):
-    device = sdk.devices.get_by_guid(device_guid)
     try:
+        int(device_guid)
```
unparalleled-js, line 64: Could also do

```
if not str(device_guid).isnumeric():`
  raise Code42CLIError("Not a valid guid.")
```

unparalleled-js, line 64: Is there a test for this?

src/code42cli/cmds/devices.py
```diff
@@ -206,51 +200,94 @@ def _get_device_info(sdk, device_guid):
...
+)
+@click.option(
+    "--last-connected-after",
+    type=MagicDate(rounding_func=round_datetime_to_day_end),
+    help="Include devices only when 'lastConnected' field is after the provided value. Argument format options are the same as --last-connected-before.",
+)
+@click.option(
+    "--created-before",
+    type=MagicDate(rounding_func=round_datetime_to_day_start),
+    help="Include devices only when 'creationDate' field is less than the provided value. Argument format options are the same as --last-connected-before.",
```
unparalleled-js, line 224: some of these lines are long and could benefit from the triple quotes

src/code42cli/file_readers.py
```diff
@@ -17,31 +19,44 @@ def read_csv_arg(headers):
...
+            for key in list(row.keys()):
+                if key not in headers:
+                    row.pop(key)
+            csv_rows.append(row)
+        if not csv_rows:
+            raise Code42CLIError("CSV contains no data rows.")
+        return csv_rows
+
+    # handle when first row has no expected headers
+    if all(field not in first_line for field in headers):
```
unparalleled-js, line 46: `elif`

I get that there is a return, but it's more readable and can prevent future bugs if we forget to change it to `elif`

src/code42cli/file_readers.py
```diff
@@ -17,31 +19,44 @@ def read_csv_arg(headers):
 
 
 def read_csv(file, headers):
```
unparalleled-js, line 21: This method is getting massive and could be more readable through the use of helper methods

tests/test_file_readers.py
```diff
@@ -22,7 +22,7 @@ def test_read_csv_handles_headerless_columns_in_proper_number_and_order(runner):
         assert result_list[1]["header3"] == "col3_val2"
```
unparalleled-js, line 22: Shouldn't there be some more tests for reading a csv, particularly when it only has some of the expected column headers?

unparalleled-js (commented):
Looks pretty good, if it works.

I think there are a lot of testing gaps still

src/code42cli/file_readers.py
timabrmsn: yeah, good call. I'll split it out so it reads cleaner

tests/test_file_readers.py
timabrmsn: true

src/code42cli/cmds/devices.py
kiran-chaudhary: Why do we want to convert it into a int, the help docs in `py42` say `guid (str)`!

src/code42cli/file_readers.py
timabrmsn: I struggled to find anything to pull out into a function that really helped readability, but did shorten it overall a bit by using a comprehension for filtering extra rows.

tests/test_file_readers.py
timabrmsn: these are now added

src/code42cli/cmds/devices.py
timabrmsn: We're not actually converting it, just making sure it _can_ be converted. Device guids will always be numerical, so this allows us to fail early before making an http request if we know it's gonna fail (if a column accidentally has a username in the guid column or something). 

Also, `.isnumeric()` fails if there's whitespace on either side of the guid string, where `int()` handles it fine.

timabrmsn: there is now!

unparalleled-js (approved):

PR #206, opened 2021-01-21
Use temp profile context when executing integration tests

unparalleled-js:
* Include a context manager that sets up a temporary default profile to use for executing integration tests. 
It will remove the temp profile and re-set the original profile (if there was one)

I have tested to make sure the original profile gets set as well making sure it works when there is no profile set.
I also confirmed that the original behavior still works if you set the environment variables.


tests/integration/__init__.py
```diff
@@ -1,36 +1,68 @@
...
+TEST_PROFILE_NAME = "TEMP-INTEGRATION-TEST"
+_LINE_FEED = b"\r\n"
+_PASSWORD_PROMPT = b"Password: "
+_ENCODING_TYPE = "utf-8"
 
 
-def encode_response(line, encoding_type=ENCODING_TYPE):
-    return line.decode(encoding_type)
+@contextmanager
+def use_temp_profile():
```
kiran-chaudhary, line 19: Can we create this as a fixture? A profile in this application is a product feature and so it makes more sense to have it as a fixture.

Secondly, from design perspective, integration module(directory) is kind of root of the integration tests and it shouldn't  be aware of `code42cli.profile`. i.e. loading it should not invoke application dependencies! At least that is how I see separation of concern. So if not fixture, what if we move this and related methods into a separate file(`util`)?

kiran-chaudhary (commented):
Apart from the one concern in design context, LGTM.

tests/integration/__init__.py
unparalleled-js: Even if it were in a separate file `util`, it is still (indirectly) invoking the CLI application specific module.

I understand what you are saying, but integration tests already need to have the CLI dependency installed, so I don't see it as a problem for it to important modules there, especially since it is doing it to further advance its purpose in testing it.

alanag13: Yeah agreed, ideally we would create/delete/get/switch profile in an integration test by executing cli commands instead of calling python functions

unparalleled-js: There is also not a real intuitive way to make a context manager a pytest fixture.
The only thing I can think of there is to make the `command_runner` a pytest fixture

unparalleled-js: When testing the profile commands, it makes sense that they are executed via CLI commands... But the intent of this is just to be a context

kiran-chaudhary: So my idea of separation of concern was in the context that "load modules only when that feature is actually being tested" i.e if we do not want to test the profile feature in certain tests then loading those modules doesn't make sense.

However since profile is used in all features, invoking it here is okay then.

unparalleled-js: We are testing profile commands in this fixture. We are making a clean-slate available for testing the other commands.  The implementation of how that occurs seems irrelevants to me.  It is test setup, not part of the test.

unparalleled-js: If you want a test that tests switching profile, that should be done differently than this.... This is not a means to testing switching profiles.

alanag13 (approved):

PR #207, opened 2021-01-22
Added help text for hostname

kiran-chaudhary:
Improve help text for `send-to`  command

Added help text for `hostname` argument to highlight the default port being used for data transfer.

### DIFF
--- a/src/code42cli/cmds/alerts.py
+++ b/src/code42cli/cmds/alerts.py
@@ -274,7 +274,10 @@ def send_to(
     or_query,
     **kwargs
 ):
-    """Send alerts to the given server address."""
+    """Send alerts to the given server address.
+
+    HOSTNAME format: address:port where port is optional and defaults to 514.
+    """
     logger = get_logger_for_server(hostname, protocol, format)
     cursor = _get_alert_cursor_store(cli_state.profile.name) if use_checkpoint else None
     handlers = ext.create_send_to_handlers(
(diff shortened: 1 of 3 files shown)


src/code42cli/cmds/alerts.py
```diff
@@ -274,7 +274,10 @@ def send_to(
     or_query,
     **kwargs
 ):
-    """Send alerts to the given server address."""
+    """Send alerts to the given server address.
+
+    HOSTNAME format: address:port where port is optional and defaults to 514.
```
kiran-chaudhary, line 279: I could have used a constant here but that was not being rendered in docs so had to copy them in all places.

alanag13: Instead of putting this in code here, can he make the help text in `server_options` instead? That way it's documented both in code and in the terminal, and doesn't need to be copied to multiple places.

alanag13: nevermind, I now see that `click.argument` doesn't have a `help` param, and that this is the advised way to do this.

unparalleled-js (approved):

unparalleled-js:
If we wanted a gross way to share doc strings, we can set the magic prop __doc__ on methods programmatically.

alanag13:
> If we wanted a gross way to share doc strings, we can set the magic prop `__doc__` on methods programmatically.
> 
> `send_to.__doc__ += f"\n\n{Constants.HOST_HELP}"`

gross indeed haha. docstrings are generally not a good candidate for sharing because methods are very likely to become different from one another and thus would end up becoming un-shared anyway.

alanag13 (approved):

PR #209, opened 2021-01-22
Reactivate

unparalleled-js:
It is hard to test deactivate without reactivate. This adds support for a reactivate device command


unparalleled-js:
It wouldn't be too hard to add `bulk_reactivate` if we think we should

timabrmsn:
Yeah, actually it does make sense to have bulk reactivate. I'm glad you caught this, since it's a very useful thing to be able to do in bulk (especially if you just accidentally bulk deactivated a bunch of devices!).

src/code42cli/cmds/devices.py
```diff
@@ -86,6 +87,29 @@ def _deactivate_device(sdk, device_guid, change_device_name, purge_date):
...
+        elif cmd_str == "deactivate":
+            sdk.devices.deactivate(device_id)
+        return device
+    except exceptions.Py42NotFoundError:
+        raise Code42CLIError("The device {} was not found.".format(device_guid))
+    except exceptions.Py42ForbiddenError:
+        raise Code42CLIError("Unable to {} {}.".format(cmd_str, device_guid))
+
+
+def _verify_guid_type(device_guid):
```
timabrmsn, line 106: we could actually just make this the callback on the `device_guid_option` and it'll automatically apply wherever we re-use the option.

timabrmsn:
```
def _verify_guid_type(ctx, param, arg):
    if arg is None:
        return
    try:
        int(arg)
    except ValueError:
         raise Code42CLIError("Not a valid guid.")
```

src/code42cli/cmds/devices.py
```diff
@@ -435,23 +465,56 @@ def bulk(state):
...
+    )
+    formatter.echo_formatted_list(result_rows)
+
+
+@bulk.command(name="reactivate")
+@read_csv_arg(headers=["guid"])
+@format_option
+@sdk_options()
+def bulk_reactivate(state, csv_rows, format):
+    """Deactivate all devices from the provided CSV containing a 'guid' column."""
```
timabrmsn, line 498: *Reactivate

src/code42cli/cmds/devices.py
```diff
@@ -435,23 +465,56 @@ def bulk(state):
...
+        _deactivate_device(
+            sdk, row["guid"], row["change_device_name"], row["purge_date"]
+        )
+
+    _bulk_handle_device_activation_update(
+        sdk, format, csv_rows, _deactivate, "deactivate", "Deactivating"
+    )
+
+
+def _bulk_handle_device_activation_update(
```
timabrmsn, line 482: This feels like one of those places that extracting out some minor duplicate logic into a function just makes it harder to read/understand what's happening. I don't see any problem with just copying the bulk_deactivate command and tweaking the little bit that's necessary to make it work for reactivate, rather than try to share code between the two commands.

unparalleled-js: Done

timabrmsn:
Are the bulk commands doing anything for you? All that's happening for me is the output gets printed, but nothing changes for the devices, and debug mode shows no api calls being made beyond the first auth.

unparalleled-js:
> Are the bulk commands doing anything for you? All that's happening for me is the output gets printed, but nothing changes for the devices, and debug mode shows no api calls being made beyond the first auth.

They were not.  They are now. I put things back the way they were. I am not sure what was going on.

timabrmsn (approved):

PR #212, opened 2021-01-25
Integration tests for send-to commands.

kiran-chaudhary:
**Description**

Added integration tests for `send-to` commands in `alerts`, `security-data` and audit-logs.

Creates a TCP or UDP server, based on protocol to be tested on localhost at port 5140.
Executes command to be tested.
Closes the server.


tests/integration/test_alerts.py
```diff
@@ -2,6 +2,8 @@
 from datetime import timedelta
 
 import pytest
+from tests.integration import run_command
```
timabrmsn, line 5: `run_command` is no longer defined, it looks like it got move into the `command_runner` fixture?

timabrmsn: As looked closer at this, I noticed that command_runner set up and tore down the temp profile on each test, so did some experimenting with handling it a bit differently, and came up with this fixture that can be used in integration tests that both handles the temp profile problem (without having to swap out the default), as well as makes run_command/pyexpect unnecessary: 


```
@pytest.fixture(scope="session")
def integration_test_profile():
    """Creates a temporary profile to use for executing integration tests."""
    host = os.environ.get("C42_HOST") or "http://127.0.0.1:4200"
    username = os.environ.get("C42_USER") or "test_username@example.com"
    password = os.environ.get("C42_PW") or "test_password"
    try:
        create_profile(TEST_PROFILE_NAME, host, username, True)
    except:
        pass
    state = CLIState()
    state.profile = get_profile(TEST_PROFILE_NAME)
    state._sdk = py42.sdk.from_local_account(host, username, password)
    yield state
    delete_profile(TEST_PROFILE_NAME)
```

Then as an example, here's how we'd write a `send-to` integration test with your DataServer class:

```
def test_auditlogs_send_to(runner, integration_test_profile):
    with DataServer(protocol="TCP"):
        result = runner.invoke(
            cli,
            [
                "audit-logs",
                "send-to",
                "localhost:5140",
                "-b",
                begin_date_str,
                "-p",
                "TCP",
            ],
            obj=integration_test_profile,
        )
    assert result.exit_code == 0
```

Thoughts?

kiran-chaudhary: This is nice.  

Though I have made some changes in the above code.

- We shouldn't be calling `py42` explicitly, let the application handle that.

- `try catch` with `pass` wasn't useful.

- Fetched current profile and set it back in tear down, to leave the system back in its original state, same as Jules had done in  `use_temp_file`.

- Set password to profile explicitly so that it doesn't prompt for it.

tests/integration/test_alerts.py
```diff
@@ -1,23 +1,25 @@
 from datetime import datetime
 from datetime import timedelta
+from shlex import split as split_command
```
kiran-chaudhary, line 3: Used `shlex` to parse string `cmd a --b c`  into a `list: [ "cmd", "a", "--b", "c"].  
Specifying the command in a string is easier to read and write.

tests/integration/test_alerts.py
kiran-chaudhary: The only drawback using `runner` is while specifying commands we can't specify the root command `code42`. We do this in unit tests, so as long as we are fine with that this is good.

tests/integration/test_auditlogs.py
```diff
@@ -19,8 +47,6 @@
         ("{} '{}' -e '{}'".format(BASE_COMMAND, begin_date_str, end_date_str)),
         ("{} '{}' --end '{}'".format(BASE_COMMAND, begin_date_str, end_date_str)),
         ("{} '{}' --event-type '{}'".format(BASE_COMMAND, begin_date_str, "test")),
-        ("{} '{}' --username '{}'".format(BASE_COMMAND, begin_date_str, "test")),
```
kiran-chaudhary, line 22: Removed command options are no longer supported and are renamed.

tests/integration/conftest.py
```diff
@@ -33,6 +36,23 @@ def use_temp_profile():
         switch_default_profile(current_profile_name)
 
 
+@pytest.fixture(scope="session")
+def integration_test_profile():
```
unparalleled-js, line 40: What is this for? It looks like we are making two integration test profiles now?

tests/integration/test_alerts.py
unparalleled-js: Why not just do `.split(" ")` ? Why do we need a whole new depedency to do this?

tests/integration/test_alerts.py
```diff
@@ -47,6 +49,23 @@
...
-    return_code, response = command_runner(command)
-    assert return_code == 0
+def test_alert_command_returns_success_return_code(
+    runner, integration_test_profile, command
+):
+    assert_test(runner, integration_test_profile, command)
+
+
+@pytest.mark.integration
+@pytest.mark.parametrize(
```
unparalleled-js, line 59: You tend to use parameterized tests a lot in a way that is difficult to read or change (as I experienced on Monday.

Parameterized tests should only be used when there is a simple parameter.  Whole complex arg strings are not good candidates here. It would be better to have repetitive individual tests than to have these parameterized tests everywhere

tests/integration/test_auditlogs.py
```diff
@@ -1,16 +1,44 @@
...
+def test_auditlogs_send_to(runner, integration_test_profile, command, protocol):
+    with DataServer(protocol=protocol):
+        result = runner.invoke(
+            cli, split_command(command), obj=integration_test_profile
+        )
+    assert result.exit_code == 0
+
+
 @pytest.mark.integration
 @pytest.mark.parametrize(
```
unparalleled-js, line 43: We should split this up into individual tests

tests/integration/test_auditlogs.py
```diff
@@ -1,16 +1,44 @@
...
-begin_date = datetime.utcnow() - timedelta(days=-10)
+begin_date = datetime.utcnow() - timedelta(days=2)
 begin_date_str = begin_date.strftime("%Y-%m-%d %H:%M:%S")
-end_date = datetime.utcnow() - timedelta(days=10)
+end_date = datetime.utcnow() - timedelta(days=0)
 end_date_str = end_date.strftime("%Y-%m-%d %H:%M:%S")
 
 
+@pytest.mark.integration
+@pytest.mark.parametrize(
```
unparalleled-js, line 21: We should split this up into individual tests

tests/integration/test_auditlogs.py
```diff
@@ -41,6 +67,7 @@
         ("{} '{}' --debug".format(BASE_COMMAND, begin_date_str)),
     ],
 )
-def test_auditlogs_search_command_returns_success_return_code(command, command_runner):
-    return_code, response = command_runner(command)
-    assert return_code == 0
+def test_auditlogs_search_command_returns_success_return_code(
+    runner, integration_test_profile, command, command_runner
```
unparalleled-js, line 71: The command runner is already using a temp profile. Why do we need to take another profile fixture here?

tests/integration/test_securitydata.py
```diff
@@ -0,0 +1,34 @@
...
+
+from code42cli.main import cli
+
+
+begin_date = datetime.utcnow() - timedelta(days=20)
+begin_date_str = begin_date.strftime("%Y-%m-%d")
+
+
+@pytest.mark.integration
+@pytest.mark.parametrize(
```
unparalleled-js, line 16: We should split this up into individual tests

tests/integration/test_alerts.py
timabrmsn: The try block for create_profile was for if the test profile already exists from a previous run that didn't get cleaned up (i.e. you ctrl-c out of a pytest run that you see everything failing).

tests/integration/util.py
```diff
@@ -30,3 +34,26 @@ def wrapper():
...
+        self.process = None
+
+    def __enter__(self):
+        self.process = subprocess.Popen(self.command.split(" "))
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.process.kill()
+
+
+def assert_test(runner, integration_test_profile, command):
```
unparalleled-js, line 57: `run_test` or `assert_test_is_successfull()`

tests/integration/test_alerts.py
```diff
@@ -47,6 +49,23 @@
...
+    "command,protocol",
+    [
+        ("alerts send-to localhost:5140 -p TCP -b {}".format(begin_date_str), "TCP"),
+        ("alerts send-to localhost:5140 -p UDP -b {}".format(begin_date_str), "UDP"),
+    ],
+)
+def test_alerts_send_to(runner, integration_test_profile, command, protocol):
```
unparalleled-js, line 66: If we are going to use click's runner here, we might as well do it for all the other tests.

tests/integration/conftest.py
timabrmsn: command_runner shouldn't be needed anymore with this fixture in place. With `command_runner` we were setting up and tearing down the test profile every single test, this just does it once for the session. This also sets the password, which means we don't need to use pyexpect to handle the prompt.

kiran-chaudhary: This is same as the other method you created, the only difference is now its a fixture, I am planning to remove one of them, but there were subtle changes in the approach so I created a new method so that it doesn't affect existing tests. The scope here was suppose to be `send-to` only.

tests/integration/test_alerts.py
kiran-chaudhary: the command string is the simplest parameter, with each line one can see all the different possible command combinations instead of reading each method. I don't know if I can have an explanation why this is easier for me.

tests/integration/test_alerts.py
timabrmsn: If we're using `set_password()` in an automated environment, we'll either need to make sure the keyring backend works, or set `.assume_yes=True` on the CLIState object before calling `set_password()` to it can fail back to the flat file. That's why I ended up just creating and setting the sdk manually on the CLIState object initially.

kiran-chaudhary: Yea but without that it prompts for password, so then we are back to square one :)

timabrmsn: We _could_ just use the runner here to create the profile w/ password...

```
cmd = f"profile create -n {TEST_PROFILE_NAME} -u {username} -s {host} --password {password}"
runner.invoke(cli, cmd.split())
```

And if we wanted to be a bit safer about cleanup, instead of changing the default profile, we could just append f"--profile {TEST_PROFILE_NAME}" to each integration test cmd.

tests/integration/test_alerts.py
kiran-chaudhary: when inputs like "2020-01-02 00:00:00" or ' New case' it will parse incorrectly with split(" "),

tests/integration/test_alerts.py
kiran-chaudhary: To convince team to allow this, when ssl is ready all we need in integration is to add one single line with the command in parametrize block and that will be it. ( given that the `DataServer` class is sufficient )

tests/integration/test_alerts.py
timabrmsn: Actually, I think I like the idea of just using the runner to create the profile with password in one-shot, since if we ever refactor any of the profile code, we shouldn't have to touch the tests (unless we make a breaking change to the user interaction too).

timabrmsn: Ugh, nevermind. It looks like I didn't add the `-y` option to `create` to bypass the prompt if keyring isn't available. So we'd need to update that first to make that work anyway.

kiran-chaudhary: I think having integration tests around profile should be a task on its own and using profile in other features should be separate, so imo it does not matter how profile was/is created while testing other features.

tests/integration/test_alerts.py
unparalleled-js: Looks like it's a standard library anyway, so it's not "another dependency" as I had thought

tests/integration/conftest.py
unparalleled-js: The runner seems the way to go, hopefully it is pretty simple to get rid of the other method and swap it all out

tests/integration/test_alerts.py
unparalleled-js: If a test fails, there is no way to run just that test. You have to run them all. This is not how parameterized tests should be used. We have been doing this a lot in the integration tests. It also makes it difficult to know which tests are still needed and which ones we have. A command string is a string of args -- not a single arg.

unparalleled-js: So, if the SSL test fails, we now have to run all the other integration tests each time or comment them out. 

 Also, when looking at the Test Failure report, it is not immediately clear which of the parameterized tests was the one that failed.

tests/integration/test_alerts.py
timabrmsn: I just merged in adding -y to the `profile create` command, so the integration test profile fixture can be as simple as: 

```
@pytest.fixture(scope="session")
def integration_test_profile():
    """Creates a temporary profile to use for executing integration tests."""
    host = os.environ.get("C42_HOST") or "http://127.0.0.1:4200"
    username = os.environ.get("C42_USER") or "test_username@example.com"
    password = os.environ.get("C42_PW") or "test_password"
    runner.invoke(
        cli, 
        f"profile create -n {TEST_PROFILE_NAME} -u {username} -s {host} --password {password} -y".split()
    )
    yield
    runner.invoke(cli, f"profile delete {TEST_PROFILE_NAME} -y")
```

Then if we pass `--profile {TEST_PROFILE_NAME}` to each command we don't need to mess with switching the default.

tests/integration/test_alerts.py
kiran-chaudhary: I agree that testing individual test with parameterize is difficult, there is a plugin that can be used to execute only failed tests from previous execution, but overall I see the pain here.
However on the second concern, the report does give exactly which test failed. In fact with parameterize we can see the command in the log against the failure in the report. Which we can use directly without a need to open the test file and see what command was executed under that method.

tests/integration/test_alerts.py
kiran-chaudhary: Instead of appending the profile name to every command what if we use `profile use` command to use the temporary profile?

tests/integration/test_alerts.py
kiran-chaudhary: This is the log I see when test is executed. Observe how command being passed as in input from parameterize is mentioned in it, the `-TCP` is another parameter that is passed for other purpose.  
```
$ pytest -s -x -v tests -m integration -k send-to
collected 677 items / 671 deselected / 6 selected                                                                                                                                                          

tests/integration/test_alerts.py::test_alerts_send_to[alerts send-to localhost:5140 -p TCP -b 2021-01-09-TCP] PASSED
tests/integration/test_alerts.py::test_alerts_send_to[alerts send-to localhost:5140 -p UDP -b 2021-01-09-UDP] PASSED
tests/integration/test_auditlogs.py::test_auditlogs_send_to[audit-logs send-to localhost:5140 -p TCP -b '2021-01-27 13:52:56'-TCP]
```

kiran-chaudhary: Seems like you might have tried `pytest` without `-v` option and hence all the confusion!! In tox we do use `-v` option so this shouldn't be a concern.

tests/integration/test_alerts.py
kiran-chaudhary: Done.

tests/integration/test_alerts.py
timabrmsn: I'm just trying to avoid having the tests leave our profiles in an unexpected state. If we change the default and the tests fail to complete and clean up the fixture, then we need to manually switch it back. It's not a huge concern, but it's happened to me when I was trying these out and if we can avoid it it'd be nice.

tests/integration/test_alerts.py
unparalleled-js: When a test fails for me, I typically use Pycharm, and that is where the output seems confusing to me, especially when there are lots of different commands in the same parameter.

This is just simply an abuse of parameterized tests.  Technically, you can put every single test in the whole suite into one giant parameterized test. Don't do this please!

Also, this is just part of the team process that has been discussed in the past on other occasions, ask @alanag13 for confirmation there.

Using parameterized tests this way increases the complexity of running the tests, and in make-or-break moments (such as the Monday of the latest release), it is frustrating to try and sift through these.  I hope you understand.

I will approve for now, but going forward, please let's not do this.

unparalleled-js (approved):
@kiran-chaudhary  I am not convinced that the parameterized test for entire command string is a correct use of the parameterized test marker for reasons I replied in the comments.

There is a chance these parameterized tests are going to end up being broken apart by other devs in the future, so I'd prefer that it would just happen now.

However, if this is urgent, then I don't consider it a blocker.

tests/integration/test_alerts.py
unparalleled-js: There is also the point that is just hard to know what tests we have and which ones are missing, when looking at the code. We have to parse a python dictionary with our eyeballs

tests/integration/test_alert_rules.py
```diff
@@ -1,6 +1,8 @@
 import pytest
+from tests.integration.conftest import append_profile
```
unparalleled-js, line 2: Since this gets imported everywhere, could include it on the command_runner itself, or subclass the runner to be `command_runner_with_profile`

tests/integration/test_alerts.py
kiran-chaudhary: I would happily change this if everyone else feels it so. However, I would share my disagreement with above raised concerns, particularly that "using parameterize here is abuse of the feature" 

- As per the [doc](https://docs.pytest.org/en/stable/example/parametrize.html#generating-parameters-combinations-depending-on-command-line)  the usage of `parameterize` is to provide all the possible combinations in a single place. In the current implementation I have not created a single test case to cover everything,  `send-to` and `list` commands have separate test methods and then all combinations of each sub-command are parameterized.

- On "usage of `pycharm`", I feel there has to be a way to configure it to pass other options like  `-v`, `-x` which would be helpful during debugging.

- On "team process", I would like to say, the context has changed and so team needs to reconsider the process. In unit tests' context it might (not sure) make sense to parameterize but in integration tests where there are multiple combinations to test we have to reconsider it.

- On "missing tests", the number of combinations possible in `list` commands are exponentially high and cannot all be tested, I have just made sure I cover  each option at-least once. Whether writing a method for each separate combination will help to figure out  missing tests is subjective and depends on reader.

I am ready to rework and write a separate method for each command if someone else is going to rework on this again, its better I do it. imo, its wastage of resources time and under utilisation of tools. I feel `parameterize` is a good feature of the tool `pycharm` and we should be ok using it. 

@alanag13  @timabrmsn thoughts!!

tests/integration/test_auditlogs.py
kiran-chaudhary: resolving this conversation as it is addressed [here](https://github.com/code42/code42cli/pull/212#discussion_r567613905)

tests/integration/test_auditlogs.py
kiran-chaudhary: resolving this conversation as it is being discussed [here](https://github.com/code42/code42cli/pull/212#discussion_r567613905)

tests/integration/test_auditlogs.py
kiran-chaudhary: done.

tests/integration/test_securitydata.py
kiran-chaudhary: resolving this conversation as it is being discussed [here](https://github.com/code42/code42cli/pull/212#discussion_r567613905)

tests/integration/util.py
kiran-chaudhary: changed.

tests/integration/conftest.py
kiran-chaudhary: Done.

tests/integration/test_alerts.py
timabrmsn: I think we're just pushing too much shared info into the parameters, since we're passing the whole CLI cmd string instead of just the changing parameters for each run. 

We should turn this:

```
@pytest.mark.integration
@pytest.mark.parametrize(
    "command",
    [
        "{} list".format(ALERT_RULES_COMMAND),
        "{} list -f CSV".format(ALERT_RULES_COMMAND),
        "{} list -f TABLE".format(ALERT_RULES_COMMAND),
        "{} list -f RAW-JSON".format(ALERT_RULES_COMMAND),
        "{} list -f JSON".format(ALERT_RULES_COMMAND),
        "{} list --format CSV".format(ALERT_RULES_COMMAND),
        "{} list --format TABLE".format(ALERT_RULES_COMMAND),
        "{} list --format JSON".format(ALERT_RULES_COMMAND),
        "{} list --format RAW-JSON".format(ALERT_RULES_COMMAND),
    ],
)
def test_alert_rules_list_command_returns_success_return_code(
    runner, integration_test_profile, command
):

    assert_test_is_successful(runner, append_profile(command))
```

Into something like this:

```
@pytest.mark.integration
@pytest.mark.parametrize(
    "options",
    [
        [],
        ["-f", "CSV"],
        ["-f", "TABLE"],
        ["-f", "RAW-JSON"],
        ["-f", "JSON"],
        ["--format", "CSV"],
        ["--format", "TABLE"],
        ["--format", "JSON"],
        ["--format", "RAW-JSON"],
    ],
)
def test_alert_rules_list_command_returns_success_return_code(
    runner, integration_test_profile, options
):
    cmd = ["alert-rules", "list", "--profile", "TEMP-INTEGRATION-TEST", *options]
    assert_test_is_successful(runner, cmd)
```

It does the same exact work, it's just easier to isolate/update what's being parametrized vs what is static to this individual test.

unparalleled-js: @timabrmsn This is certainly an improvement, and for most of the parameterized tests, this would be an improvement.

kiran-chaudhary: Ok, I will go ahead and do the changes as requested, also is splitting the command into a list better, I prefer reading without quotes and commas but can cover that as well if that is important!!

timabrmsn: I'm ok with the cmd and options being strings that get combined then split when passed to the runner.

unparalleled-js: @alanag13 said he would add his two cents here

unparalleled-js (dismissed):
holding my approval until this is worked out.

tests/integration/test_alerts.py
alanag13: I have a couple thoughts here -- I see that I was @mentioned a couple times here
- first off, when the team disagrees on implementation details like this, it is probably best to simply note that further discussion is needed, schedule some time to agree on an approach, etc, but to not allow a disagreement like this to delay a release/feature merge.

That being said, If I could make a style rule that would emit a warning whenever a parameterized test was used, I probably would. Parameterized tests can be useful, but are often used for situations that can later cause difficulty. I find them most useful when the parameter list is very small, as a means to test different types of an input (will this work with a very small number? a very low number? a decimal number) that are all taking a similar code path. An example in this scenario might be to parameterize both `-f` and `--format` to make sure they both work as input. 

The biggest disadvantage is that parameterized tests cannot typically be run individually. This is not just an inconvenience -- I have seen this lead to nasty, difficult-to-diagnose bugs  and/or false positives in tests because a test was dependent on a previous parameter running first.

Additionally, especially as someone who works in multiple frameworks/languages, parameterized tests introduce _uncertainty_ to me:
- if one of the parameterized tests fails, will I know which one?
- if one of the parameterized tests fails, do the rest still get run, or or does the test stop?
- if one of the parameterized tests fails, will the problem be in the same general section of code as the other parameters?

I have worked with testing tools where some of the answers to these questions is _no_. Even if the answer to all of these is _yes_, the fact that I do not know right away just by looking at the code is off-putting. This uncertainty doesn't occur when using one function per test, which is why I typically avoid them.

Typically, if these were unit tests, I would argue that `CSV/TABLE/RAW-JSON/JSON` takes significantly different code paths depending on which you choose and should be separate tests, but in this case (since it is an integration test), we can be reasonably confident that the process that thing we're trying to test works as expected using only one of those parameters -- there may not even be a need to create a separate test for each one. Recall that the point of the integration test is essentially just to test that we can hit the network successfully. Anything that could potentially go wrong due to your choice of format should be covered by a unit test.

alanag13: To summarize -- In the end, whether or not to use parameterized tests can probably be categorized as a personal preference. I have had experiences that have given me a strong personal preference to avoid them which i'm happy to share to allow others to avoid the same pain -- but I don't want this project to be built solely around enforcing my personal preferences.

kiran-chaudhary: Made changes based on above discussion.

Committed the changes in two steps, i.e a commit each for the step as follows:

1. Made changes as suggested by Tim to pass single option in parameterize
2. Removed parameterize options all together based on the point that  _" the point of the integration test is essentially just to test that we can hit the network successfully in each common use-case. Anything that could potentially go wrong due to your choice of format should be covered by a unit test."_

kiran-chaudhary: If someone wants to have all the tests back, I can easily do it by getting rid of the last commit.

kiran-chaudhary: For `send-to` sub-command tests, parameterize still exists but now it accepts just the protocol field as option which has just two values.

kiran-chaudhary: To answer Alan's queries above, the answer is yes to all three of them.

alanag13 (approved):

PR #216, opened 2021-01-25
org and split up tests

unparalleled-js:
* makes case number an option rather than an arg for the commands `code42 cases file-events add` and `code42 cases file-events remove` per the rule where if we have multiple required args, we use options instead so that arg order cannot be confused.

When fixing the tests after this change, I found it really hard to know where all the tests were since they were not organized. Also, there was a large and complex parameterized test that is easier to work with if broken up into separate tests, so this pr also organized and splits up those tests

### DIFF
--- a/src/code42cli/cmds/cases.py
+++ b/src/code42cli/cmds/cases.py
@@ -17,6 +17,9 @@ from code42cli.output_formats import OutputFormatter
 
 
 case_number_arg = click.argument("case-number", type=int)
+case_number_option = click.option(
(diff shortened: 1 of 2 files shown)


src/code42cli/cmds/cases.py
```diff
@@ -17,6 +17,9 @@
 
 
 case_number_arg = click.argument("case-number", type=int)
```
alanag13, line 19: should this arg still be here?

alanag13 (approved):

PR #217, opened 2021-01-25
JSON output format and other fixes

unparalleled-js:
* Use `json.dumps()` rather than `pprint` to follow suit with other `show` commands' output format.
* Impove doc strings language
* Change default param from `"."` to `os.getwd()`

### DIFF
--- a/src/code42cli/cmds/cases.py
+++ b/src/code42cli/cmds/cases.py
@@ -20,10 +20,14 @@ case_number_option = click.option(
     "--case-number", type=int, help="The number assigned to the case.", required=True
 )
 name_option = click.option("--name", help="The name of the case.",)
-assignee_option = click.option("--assignee", help="The UID of the user assigned to the case.")
+assignee_option = click.option(
+    "--assignee", help="The UID of the user to assign to the case."
+)
 description_option = click.option("--description", help="The description of the case.")
 findings_option = click.option("--findings", help="Any findings for the case.")
-subject_option = click.option("--subject", help="The user UID of the subject of the case.")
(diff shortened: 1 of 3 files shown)


unparalleled-js:
I will put any remaining changes on this pr ... standby

src/code42cli/cmds/cases.py
```diff
@@ -20,20 +19,23 @@
...
     "--case-number", type=int, help="The number assigned to the case.", required=True
 )
-name_option = click.option("--name", help="Name of the case.",)
-assignee_option = click.option("--assignee", help="User UID of the assignee.")
-description_option = click.option("--description", help="Description of the case.")
-findings_option = click.option("--findings", help="Findings on the case.")
-subject_option = click.option("--subject", help="User UID of a subject of the case.")
+name_option = click.option("--name", help="The name of the case.",)
+assignee_option = click.option(
+    "--assignee", help="The UID of the user to assign to the case."
```
unparalleled-js, line 24: `to assign to the case` makes sense for both `update` and `create`, where this is used.

unparalleled-js:
OOOk I am done now

unparalleled-js:
My final concern is that we have a --change-device-name flag for deactivating devices that prepends deactivated_<todays-date> to the name of the device... However, there is no way to undo this with the CLI... We can reactivate the device but we can't rename the device or remove the deactivate prefix..... Not sure what to do here. The flag in general seems like a really specific use-case, maybe too specific for the CLI.

alanag13 (approved):

PR #222, opened 2021-02-03
Handle custom exceptions in cases

kiran-chaudhary:
**Description**

- Handle custom exceptions added in `Py42 cases`.

**Issues Resolved**
INTEG-1468

- [n/a] Add docstrings for any new public parameters / methods / classes

### DIFF
--- a/src/code42cli/cmds/cases.py
+++ b/src/code42cli/cmds/cases.py
@@ -4,7 +4,9 @@ import os
 import click
 from py42.clients.cases import CaseStatus
 from py42.exceptions import Py42BadRequestError
+from py42.exceptions import Py42CaseAlreadyHasEventError
 from py42.exceptions import Py42NotFoundError
+from py42.exceptions import Py42UpdateClosedCaseError
 
 from code42cli.click_ext.groups import OrderedGroup
 from code42cli.errors import Code42CLIError
@@ -241,6 +243,10 @@ def add(state, case_number, event_id):
     """Associate a file event to a case, by event ID."""
     try:
         state.sdk.cases.file_events.add(case_number, event_id)
+    except Py42UpdateClosedCaseError:
+        raise
+    except Py42CaseAlreadyHasEventError:
+        raise
     except Py42BadRequestError:
         raise Code42CLIError("Invalid case-number or event-id.")
 
--- a/tests/cmds/conftest.py
+++ b/tests/cmds/conftest.py
@@ -72,12 +72,17 @@ def cli_state_without_user(sdk_without_user, cli_state):
 
 
 @pytest.fixture
-def user_already_added_error(mocker):
+def custom_error(mocker):
     err = mocker.MagicMock(spec=HTTPError)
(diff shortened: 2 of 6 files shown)


src/code42cli/cmds/cases.py
```diff
@@ -241,6 +244,10 @@ def add(state, case_number, event_id):
     """Associate a file event to a case, by event ID."""
     try:
         state.sdk.cases.file_events.add(case_number, event_id)
+    except Py42UpdateClosedCaseError:
+        raise
+    except Py42CaseAlreadyHasEventError:
+        raise
     except Py42BadRequestError:
```
kiran-chaudhary, line 251: Since the above 2 custom exceptions are subclass of `Py42BadRequestError` had to catch and re-raise them to avoid displaying below error message.

tests/cmds/conftest.py
```diff
@@ -72,12 +72,17 @@ def cli_state_without_user(sdk_without_user, cli_state):
 
 
 @pytest.fixture
-def user_already_added_error(mocker):
```
kiran-chaudhary, line 75: This is just a refactor to reuse `custom_error`

src/code42cli/cmds/cases.py
unparalleled-js: What cases causes the regular BadRequest to show up?  Do we really need to handle that?  I would expect a BadRequest to the more of the fault of the CLI rather than the User

unparalleled-js: We should be able to remove the error handling enitirely here

src/code42cli/cmds/cases.py
```diff
@@ -241,6 +243,10 @@ def add(state, case_number, event_id):
...
         raise Code42CLIError("Invalid case-number or event-id.")
```
unparalleled-js, line 252: There are other spots in this file where we are doing error handling that I would expect could be removed now that we have added the custom exception to the global catcher

src/code42cli/cmds/cases.py
kiran-chaudhary: When the event-id or case-number are invalid it returns BadRequestError, if we add BadRequest in the groups list we would lose the custom messages we are showing to the user.

It will also affect `devices` and other feature where we are passing custom messages by catching BadRequest error

src/code42cli/cmds/cases.py
kiran-chaudhary: This is being discussed [here](https://github.com/code42/code42cli/pull/222#discussion_r569296895), resolving this comment to avoid multiple discussion on the same topic.

src/code42cli/cmds/cases.py
unparalleled-js: I am definitely not suggestion add BadRequest to the groups list! I just wanted to make sure it was actually needed here

unparalleled-js (approved):
Looks good, keeping the existing exception handling in place makes sense.

PR #223, opened 2021-02-03
Raise exception instead of echo in alert-rules add when rule-id is invalid

kiran-chaudhary:
**Description**

- Raise exception when rule-id is invalid, make error message adhere to format

**Issues Resolved**
INTEG-1470

- [n/a] Add docstrings for any new public parameters / methods / classes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,10 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
     - `cases file-events add` sets an event id that is already added to the case.
     - `cases file-events remove` is performed on an already closed case.
 
+### Fixes
+
+- Fixes issue where `code42 alert-rules bulk add` would show as successful when adding users to a non-existent alert rule.
+
 ### Added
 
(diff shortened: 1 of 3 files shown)


unparalleled-js:
Seems like we should add a changelog under a `Fixes` header like

`Fixes issue where bulk alert rules would show as successful when adding users to a non-existent alert rule`.

CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 ## Unreleased
 
+### Fixes
```
unparalleled-js, line 13: Should be `Fixed`

CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
...
+
+- Fixes issue where `code42 alert-rules bulk add` would show as successful when adding users to a non-existent alert rule.
```
unparalleled-js, line 15: Prob `Fixed` here as well

unparalleled-js (approved):

PR #224, opened 2021-02-04
Feature/extension scripts

timabrmsn:
This PR adds some helpers for writing code42cli powered extensions with a minimal amount of boilerplate, giving the extension-writer cli profiles, exception handling, and sdk initialization out of the box, so all they need to do is their custom script logic. Extensions can be standalone scripts that require no installation, or can be installed as a plugin so the commands become part of the main CLI usage (using the `click_plugins` package that helpfully keeps everything else working if an extension breaks). 

### Standalone scripts
To write a cli-powered standalone script, you just need to import the `script` custom `click.Group` and `sdk_options`, then add your click commands. This example is a script that takes a guid from an option and pretty-prints the resulting settings dict: 

```
import click
from code42cli.extensions import script, sdk_options
from pprint import pprint

@click.option("--guid", required=True)
@sdk_options()
def get_device_settings(state, guid):
    settings_response = state.sdk.devices.get_settings(guid)
    pprint(settings_response.data)


if __name__ == "__main__":
    script.add_command(get_device_settings)
    script()
```

If there's only a single defined command, the custom group makes it the default, so it can be invoked just from the base script itself: 

```
# python script.py -h
Usage: script.py  [OPTIONS]

Options:
  --guid TEXT     [required]
  -d, --debug     Turn on debug logging.
  --profile TEXT  The name of the Code42 CLI profile to use when executing this command.
  -h, --help      Show this message and exit.
```

If multiple commands are defined, then `script` operates as a normal group and enumerates them, requiring the subcommand be called: 

```
import click
from code42cli.extensions import script, sdk_options
from pprint import pprint


@click.option("--guid", required=True)
@sdk_options()
def get_device_settings(state, guid):
    settings_response = state.sdk.devices.get_settings(guid)
    pprint(settings_response.data)


@click.argument("guid")
@sdk_options()
def get_device_user(state, guid):
    settings_response = state.sdk.devices.get_settings(guid)
    print(settings_response.user_id)


if __name__ == "__main__":
    script.add_command(get_device_settings)
    script.add_command(get_device_user)
    script()
```

```
# python script.py -h
Usage: script.py [OPTIONS] COMMAND [ARGS]...

Options:
  -h, --help  Show this message and exit.

Commands:
  get-device-settings
  get-device-user
```

To help when a user might have multiple python installations in their path and `python script.py` isn't finding the `code42cli` package (but the regular `code42 -h` command works), a `--python` option has been added to print out the path to the interpreter that code42 is installed in. So the user can then just call `$(code42 --python) script.py` and it should just work without having to fiddle with their paths. 

### Installable extensions

To make the above script installable as a permanent extension to the CLI, you can put it in a package structure with a setup.py where the defined entrypoints are `code42cli.plugins`, the CLI will then be able to find them. 

I've created an example repo with two installable packages here: https://github.com/timabrmsn/cli_plugin_test

To install the settings example command from above you can run 

```
$(code42 --python) -m pip install "git+https://github.com/timabrmsn/cli_plugin_test#egg=code42cli_settings&subdirectory=settings"
```

And to install the users one: 

```
$(code42 --python) -m pip install "git+https://github.com/timabrmsn/cli_plugin_test#egg=code42cli_users&subdirectory=users"
```

This allows support/PS to have a single repo of their extension scripts, and they can either link directly to the standalone one, or instruct the user to install it, depending on the need/use-case.


src/code42cli/main.py
```diff
@@ -50,9 +50,14 @@ def exit_on_interrupt(signal, frame):
 }
 
 
-@click.group(cls=ExceptionHandlingGroup, context_settings=CONTEXT_SETTINGS, help=BANNER)
+@click.group(
+    cls=CLIGroup, context_settings=CONTEXT_SETTINGS, help=BANNER,
+)
+@click.option(
```
timabrmsn, line 56: This option doesn't actually do anything except put it in the help text, since the parsing is short-circuited by the InterpreterGroup.

alanag13:
A couple thoughts here:
- It may be valuable to look into a way to see if it would be possible to dynamically discover/import these scripts. For example, could we have a designated directory in `.code42cli` where these could be dropped, and then they would all show up as additional options in the main help text? It seems like it would be _really_ nice to be able to keep a collection of these and not have to remember the names/paths of custom scripts as they get developed over time.

- We should probably look over our API surface and come up with a better organizational scheme for what is public (okay to call via an extension) and what isn't, and make sure to adjust our coding practices to reflect that the terminal itself is no longer the only interface.

- We should demo this to PS, perhaps even only in this state, to give them a feel for what the experience would be like for them.

src/code42cli/click_ext/groups.py
```diff
@@ -19,6 +23,26 @@
...
+    """A `click.Group` subclass that captures the `--script` option and short-circuits
+    the normal CLI parsing, passing all subsequent args to the python interpreter that
+    the CLI is installed into, allows for reliably executing custom scripts that import
+    code42cli.
+    """
+
+    def parse_args(self, ctx, args):
+        if len(args) > 1 and args[0] == "--script":
+            args[0] = sys.executable
+            if platform.system() == "Windows":
```
unparalleled-js, line 36: Can we sanitize the args somehow?
Probably a far-fetched concern, but if the CLI was running on a server, like maybe as part of an integration of some sort, and users were able to call custom scripts remotely, could this could be used as part of a shell-injection attack?

Bandit scan:
```
 Issue: [B605:start_process_with_a_shell] Starting a process with a shell, possible injection detected, security issue.
   Severity: High   Confidence: High
   Location: src/code42cli/click_ext/groups.py:40
   More Info: https://bandit.readthedocs.io/en/latest/plugins/b605_start_process_with_a_shell.html
39                      cmd = shlex.join(args)
40                  status = os.system(cmd)
41                  sys.exit(status)
```

Oddly, `bandit` claims it is a HIGH severity but the documentation says it should be LOW/

timabrmsn: I don't think we need to worry about that since this function is explicitly _for_ running user-provided code. If someone can hijack the actual execution of the CLI to get it to call `--script` on something unintended, they likely already have the ability to just run any arbitrary command they want.

unparalleled-js: I was imagining a scenario where some integration or app intentionally allowing user input to put into the CLI command, which like I said, is maybe a far-fetched scenario. But, in this scenario, they expose the CLI essentially, so it's not hijacked (yet).

unparalleled-js: but it would be outside of our control if someone did that, albeit a dangerous decision on their part, but sanitization on our end could be an extra layer of security

unparalleled-js:
A command that executes raw input from the user is kind of a red flag to me. That should be avoided whenever possible.

What if this PR could be made into a private extension for internal Code42 use? They would use this private CLI package (who's upstream is this public repo), and from there, they can add their own scripts?

Just a thought.

unparalleled-js:
What if - we made another python package, like `code42cli-lib` that contains the parts of the code42cli that we want to share.  Then, they can just write normal python scripts.

kiran-chaudhary:
I am sure this might have been discussed but for my benefit I would like to raise it,  if someone knows python then they can use `py42`, the only advantage cli can provide here to use script is about `incremental` results. Also one cannot use the custom script to go ahead and try something that is not supported in `Py42`!!  So do we have real use cases here? the whole purpose of having it open source was others can come and add commands they feel are useful and raise PR against it!!

kiran-chaudhary:
the standard way to support such thing would be to have `-i` or `--interactive` option instead of a `--script` option. As we do `python -i` and that would load some default modules/objects and one can start doing as explained in the description.

kiran-chaudhary:
Also tested out the same,

`code42 --script -h` throws error `AttributeError: module 'shlex' has no attribute 'join'`

timabrmsn:
@kiran-chaudhary the background for this is that while we _do_ want internal teams who write scripts that go to customers (Pro Service, Support, SST, etc) to contribute to the CLI development, the CLI isn't going to be able to accommodate every specific use-case. Sometimes a customer has a very particular need that the general goal of the CLI doesn't accomplish, and so support/PS/SST will need to write something custom for them.

Exposing this "extension" functionality in the CLI relieves the writer of the burden of having to do any boilerplate logic around getting credentials from the user, initializing the sdk, and handling/logging every possible Py42 error. 

So remember the use case here is primarily for internal teams to be able to quickly and easily write custom scripts that are intended to be given to and run by customers (who often don't have a lot of programming experience). 

With these, our teams can write up a script quickly, give it to the customer and just say "run `pip3 install code42cli` and create a profile". And this only needs to be done once for all future custom scripts to just work!

And to @unparalleled-js's question about is it important if they call it `code42 --script script.py` instead of `python script.py`, it's important to have an easy option to run it if there are path issues that prevent `python script.py` from working (like if their `pip` cmd is linked to a different python install than their `python` cmd). This way we know that if `code42 <regular_cmd>` works, then our script should just work too.

timabrmsn:
> AttributeError: module 'shlex' has no attribute 'join'

Apparently `.join()` was only added in 3.8. I'll update.

src/code42cli/click_ext/groups.py
timabrmsn: I did start out playing around with just having `code42 --python` just return the full path to the interpreter `code42` was installed under. So then if a user was having problems with the regular `python script.py` they could try `$(code42 --python) script.py`.

unparalleled-js:
Can we at least get reach out to the security guild or someone and get their advice?

unparalleled-js:
@timabrmsn If this is intended for internal people to use, as you say, shouldn't it not be part of the main/publicy available CLI?

timabrmsn:
The primary _writers_ of these scripts are going to be internal teams, yes. But the primary _runners_ of these scripts will be customers, so they have to work with regular public-facing CLI installs.

src/code42cli/click_ext/groups.py
timabrmsn: I think I'll just revert the exec python logic back to just printing the python path. It keeps the code simple and gives users a relatively easy path to figuring things out if their environment is broken.

unparalleled-js:
> The primary _writers_ of these scripts are going to be internal teams, yes. But the primary _runners_ of these scripts will be customers, so they have to work with regular public-facing CLI installs.

Ah ok I did not realize that.  I will process this.

alanag13:
@timabrmsn @unparalleled-js @kiran-chaudhary there's a lot of chatter happening here and comments are getting buried. Lets set up a meeting and get everyone on the same page.

src/code42cli/extensions.py
```diff
@@ -0,0 +1,5 @@
+from code42cli.click_ext.groups import ExtensionGroup
+from code42cli.main import CONTEXT_SETTINGS
+from code42cli.options import sdk_options  # noqa: F401
```
kiran-chaudhary, line 3: wondering why are ignoring unused imports or suppressing warnings, `noqa...` statement? I need to read more on the usage of this statement, but the first thing that comes to my mind is will this suppress warnings only in context to this module or will it suppress all-over `code42cli`!!?

kiran-chaudhary (commented):
LGTM from code's point of view. Wondering why we haven't documented this feature, the description given in the PR should be documented for future reference.

src/code42cli/extensions.py
kiran-chaudhary: Also can this be resolved by abstraction/design, breaking up the modules in such a way that we don't have to suppress any warnings or unused imports!

timabrmsn: I'm pretty sure the comment only suppresses it for the line being commented, I still got unused import style failures after this commit went in. 

I imported sdk_options here just for the convenience of script-writers, so they can just do:

```
from code42cli.extensions import script, sdk_options
```

timabrmsn:
> Wondering why we haven't documented this feature

Documentation will be coming, I'm just holding off for now until I can demo the feature to our customer-facing teams to get any feedback before we finalize this implementation. I'm also not sure where the best spot is for the documentation to live. It could be in the README, ReadTheDocs, or the Dev Portal articles...

src/code42cli/extensions.py
unparalleled-js: Another way to get around this would be to create a directory named `extensions` with an `__init__.py` importing things for convenience; it seems like that is ok to do because we do it other places to make importing things more convenient.

timabrmsn: Ah, good call. Making it a module would also allow us to better expand/organize things in the future.

timabrmsn: Huh, so the style cop still balks at unused imports even in the module __init__. I wonder why it doesn't do that in py42? I'm not seeing any config differences...

timabrmsn: nevermind. I found where we're ignoring it in the setup.cfg

alanag13:
LGTM too, but I'll leave this for now in case you get feedback and want to add changes here.

unparalleled-js (approved):

unparalleled-js:
Documentation would be a plus

unparalleled-js (changes requested):
Sorry, I accidentally approved early - everything looks good, just need a changelog and perhaps a readme update

timabrmsn:
Built user-guide can be found here: https://clidocs.code42.com/en/feature-extension_scripts/userguides/extensions.html

docs/guides.md
```diff
@@ -5,3 +5,4 @@
 * [Ingest file events or alerts into a SIEM](userguides/siemexample.md)
 * [Manage detection list users](userguides/detectionlists.md)
 * [Manage legal hold users](userguides/legalhold.md)
+* [Write custom extension scripts using the code42cli and py42](userguides/extensions.md)
```
unparalleled-js, line 8: Should `the code42cli` be `the Code42 CLI`?

timabrmsn: yeah, probably.

src/code42cli/extensions/__init__.py
```diff
@@ -0,0 +1,49 @@
+from code42cli.click_ext.groups import ExtensionGroup
+from code42cli.main import CONTEXT_SETTINGS
+from code42cli.options import debug_option
+from code42cli.options import pass_state
+from code42cli.options import profile_option
+
+
+def sdk_options(f):
```
unparalleled-js, line 8: Maybe you just want to be really explicit, but this could just be 

```
from code42cli.options import sdk_options as sdk_opts

def sdk_options(f):
    """Decorator that adds two `click.option`s (--profile, --debug) to wrapped command, as well as
    passing the `code42cli.options.CLIState` object using the [click.make_pass_decorator](https://click.palletsprojects.com/en/7.x/api/#click.make_pass_decorator),
    which automatically instantiates the `py42` sdk using the Code42 profile provided from the `--profile`
    option. The `py42` sdk can be accessed from the `state.sdk` attribute.

    Example:

        @click.command()
        @sdk_options
        def get_current_user_command(state):
            my_user = state.sdk.users.get_current()
            print(my_user)
    """
    f = sdk_opts()(f)
    return f
```

unparalleled-js: Or even `sdk_options = sdk_opts`

unparalleled-js (approved):
Very cool!

docs/userguides/extensions.md
```diff
@@ -0,0 +1,82 @@
+# Write custom extension scripts using the Code42 CLI and py42
+
+While the Code42 CLI aims to provide an easy way to automate many common Code42 tasks, there will likely be times when you need to script something the CLI doesn't have out-of-the-box.
+
+To accommodate for those scenarios, the Code42 CLI exposes a few helper objects in the `code42cli.extensions` module that make it easy to write custom scripts with `py42` that use features of the CLI (like profiles) to reduce the amount of boilerplate needed to be productive.
+
+## Before you begin
+
+The Code42 CLI is a python application written using the [click framework](https://click.palletsprojects.com/en/7.x/), and the exposed extension objects are custom `click` classes. A basic knowledge of how to define `click` commands, arguments, and options is required.
```
unparalleled-js, line 9: I am a fan of line breaks in markdown because of horizontal scrolling in text editors. It seems like most markdown renderers can handle converting the newlines to spaces too. Maybe this is the issue with my IDE -- but I have seen markdown rules around line length too.

src/code42cli/extensions/__init__.py
timabrmsn: I mainly wanted to have the object we expose be it's own thing in case we ever want to change our internal `sdk_options` we're not having to work around the extension backwards compatibility.

docs/userguides/extensions.md
timabrmsn: Huh, for some reason I had it in my mind that the newlines could cause odd linebreaks in the rendered sphinx text, but I tested it out and it doesn't seem to matter. Made it much more readable now!

PR #225, opened 2021-02-04
Disallow TLS specific args for non-TLS

unparalleled-js:
* Fixes issue where the user was still allowed to pass in `--certs` and `--ignore-cert-validation` when not using the `TLS-TCP` protocol for `send-to` commands.

### DIFF
--- a/src/code42cli/cmds/search/__init__.py
+++ b/src/code42cli/cmds/search/__init__.py
@@ -46,5 +46,5 @@ def _handle_incompatible_args(protocol, ignore_cert_validation, certs):
     )
     if arg is not None:
         raise Code42CLIError(
-            f"'{arg}' must be used with '--protocol {ServerProtocol.TLS_TCP}'."
+            f"'{arg}' can only be used with '--protocol {ServerProtocol.TLS_TCP}'."
         )
--- a/src/code42cli/cmds/search/__init__.py
+++ b/src/code42cli/cmds/search/__init__.py
@@ -43,6 +43,6 @@ def _handle_incompatible_args(protocol, ignore_cert_validation, certs):
     elif certs is not None:
         arg = "--certs"
     if arg is not None:
-        raise Code42CLIError(
-            f"'{arg}' can only be used with '--protocol {ServerProtocol.TLS_TCP}'."
+        raise click.BadOptionUsage(
+            arg, f"'{arg}' can only be used with '--protocol {ServerProtocol.TLS_TCP}'."
         )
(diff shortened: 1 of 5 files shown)


src/code42cli/cmds/search/__init__.py
```diff
@@ -21,10 +22,29 @@ def invoke(self, ctx):
...
             hostname, protocol, output_format, certs
         )
         return super().invoke(ctx)
+
+
+def _handle_incompatible_args(protocol, ignore_cert_validation, certs):
+    if protocol == ServerProtocol.TLS_TCP:
+        return
+
+    arg = (
```
alanag13, line 40: Nested ternaries are a little hard to parse, could we just use if blocks?

tests/cmds/test_alerts.py
```diff
@@ -789,6 +789,52 @@ def test_send_to_when_given_ignore_cert_validation_uses_certs_equal_to_ignore_st
     )
 
 
+@pytest.mark.parametrize("protocol", (ServerProtocol.UDP, ServerProtocol.TCP))
```
alanag13, line 792: These tests are good examples of when parameterized tests make sense.

timabrmsn:
Oh hey, I had started a branch to add a generic `depends_on` option class factory like `incompatible_with` not too long ago, but since we didn't have a need for it yet I hadn't wrapped it up. This seems like a perfect time to add that.

timabrmsn:
Actually, my `depends_on` won't work as is here, since we're depending on the specific value of an option, not just the presence of it..

src/code42cli/cmds/search/__init__.py
unparalleled-js: I did that originally but felt the opposite. I can put it back and show you what I mean

tests/cmds/test_alerts.py
unparalleled-js: Right!

src/code42cli/cmds/search/__init__.py
```diff
@@ -21,10 +22,29 @@ def invoke(self, ctx):
...
+        "--ignore-cert-validation"
+        if ignore_cert_validation is not None
+        else "--certs"
+        if certs is not None
+        else None
+    )
+    if arg is not None:
+        raise Code42CLIError(
```
timabrmsn, line 48: This should instead raise `click.BadOptionUsage`.

src/code42cli/cmds/search/__init__.py
unparalleled-js: Ah I guess it's not too bad, nevermind

unparalleled-js:
> Actually, my `depends_on` won't work as is here, since we're depending on the specific value of an option, not just the presence of it..

That seems like it would work?  

`--certs` depends on a specific value of `--protocol`, etc

timabrmsn (approved):

PR #229, opened 2021-02-04
fix state-dependent tests

alanag13:
The nightly build was failing due to some tests that only worked when run in a certain order. Going through all tests and running them individually exposed which tests needed adjusting.

### DIFF
--- a/tests/cmds/test_cases.py
+++ b/tests/cmds/test_cases.py
@@ -18,23 +18,54 @@ from code42cli.main import cli
 ALL_EVENTS = """{
   "events": [
     {
-      "eventId": "0_1d71796f-af5b-4231-9d8e-df6434da4663_984418168383179707_986472527798692818_971",
-      "eventTimestamp": "2020-12-23T12:41:38.592Z"
+      "eventId": "0_147e9445-2f30-4a91-8b2a-9455332e880a_973435567569502913_986467523038446097_163",
+      "eventTimestamp": "2020-12-23T14:24:44.593Z",
+      "exposure": [
+        "OutsideTrustedDomains",
+        "IsPublic"
+      ],
+      "fileName": "example.docx",
+      "filePath": "/Users/casey/Documents/"
     }
-  ]
+  ],
+  "totalCount": 42
 }"""
 ALL_CASES = """{
   "cases": [
     {
-      "number": 3,
-      "name": "test@test.test",
-      "updatedAt": "2021-01-24T11:00:04.217878Z",
-      "subject": "942897"
+      "assignee": 273411254592236320,
+      "assigneeUsername": "test@example.com",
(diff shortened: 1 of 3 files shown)


timabrmsn (approved):

tests/cmds/test_cases.py
```diff
@@ -209,8 +202,12 @@ def test_show_with_include_file_events_calls_file_events_get_all_with_expected_p
     cli_state.sdk.cases.file_events.get_all.assert_called_once_with(1)
 
 
-def test_show_when_py42_raises_exception_prints_error_message(runner, cli_state, error):
-    cli_state.sdk.cases.file_events.get_all.side_effect = Py42NotFoundError(error)
+def test_show_when_py42_raises_exception_prints_error_message(
```
unparalleled-js, line 205: For `CASE_DETAILS`, can we make that the actual response for getting a case? And maybe rename it to `CASE_DETAILS_RESPONSE_TEXT` or something similar

unparalleled-js:
can we make the text for the `custom_error` fixture JSON text?

tests/cmds/test_cases.py
```diff
@@ -273,9 +270,9 @@ def test_file_events_add_calls_add_event_with_expected_params(runner, cli_state)
```
unparalleled-js, line 270: On line 229, it seems to me like the wrong method is being mocked ... I don't understand what is going. `get_all` does not get called in `show`, as far as I can tell.

shouldnt we be mocking `cases.get()` instead?

unparalleled-js:
Consider it approved on my end, but I found some other oddities and ways that would make the tests less likely to break from py42 changes.

tests/cmds/test_cases.py
kiran-chaudhary: the test method at line 229, `test_show_prints_expected_data_with_include_file_events_option` ,  refers to `file-events` `get_all`, so its fine. When `--include-file-events` is passed we want all the events listed against the case.

kiran-chaudhary (approved):
Changes LGTM. 

Honestly, Looking at the refactor, I am not able to tell what issue has been resolved, some more detail in the description, a link to the nightly build execution, would have helped to understand the scenario.

Having said that, I will try to find the nightly build execution, that shouldn't be hard assuming it should be in last few executions somewhere.

kiran-chaudhary:
> Changes LGTM.
> 
> Honestly, Looking at the refactor, I am not able to tell what issue has been resolved, some more detail in the description, a link to the nightly build execution, would have helped to understand the scenario.
> 
> Having said that, I will try to find the nightly build execution, that shouldn't be hard assuming it should be in last few executions somewhere.

Please ignore this. I started my day with reviews and show the slack conversation later,  it helps to understand the investigation part.

PR #231, opened 2021-02-05
handle windows log error from ssl connection

unparalleled-js:
Fixes issue where Windows would not raise an error when sending TCP to a TLS server.
The fix works by catching a common base class between the operating systems instead of the BrokenPipError that we saw originally.
On WIndows, I observed the error was ConnectionResetError.
Both BrokenPipeError and ConnectionResetError share the base class ConnectionError.

Questions:

Are there any concerns catching the other errors that inherit the base class? Subclasses are BrokenPipeError, ConnectionAbortedError, ConnectionRefusedError and ConnectionResetError.

Is there a more preferred way to check the base classes?

### DIFF
--- a/src/code42cli/logger/handlers.py
+++ b/src/code42cli/logger/handlers.py
@@ -85,7 +85,7 @@ class NoPrioritySysLogHandler(SysLogHandler):
         them nowhere.
         """
         t, _, _ = sys.exc_info()
-        if t == BrokenPipeError:
+        if ConnectionError in t.__bases__:
             raise SyslogServerNetworkConnectionError()
         super().handleError(record)
 
--- a/src/code42cli/logger/handlers.py
+++ b/src/code42cli/logger/handlers.py
@@ -14,7 +14,7 @@ class SyslogServerNetworkConnectionError(Exception):
         super().__init__(
             "The network connection broke while sending results. "
             "This might happen if your connection requires TLS and you are attempting "
-            "unencrypted TCP communication"
+            "unencrypted TCP communication."
         )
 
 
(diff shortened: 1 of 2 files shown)


src/code42cli/logger/handlers.py
```diff
@@ -85,7 +85,7 @@ def handleError(self, record):
         them nowhere.
         """
         t, _, _ = sys.exc_info()
-        if t == BrokenPipeError:
+        if ConnectionError in t.__bases__:
```
alanag13, line 88: `issubclass(t, ConnectionError)` is the standard way to check this

unparalleled-js: ha, thanks!

unparalleled-js: I am going to push the changes, then pull it down to Windows and test there to make sure it works

alanag13:
> Are there any concerns catching the other errors that inherit the base class? Subclasses are BrokenPipeError, ConnectionAbortedError, ConnectionRefusedError and ConnectionResetError.

I don't have any concerns about these being caught, they all seem to fall into the general category of "something went wrong with your connection". It may be helpful to output a hint for this specific scenario though, like "This might happen if your connection requires TLS and you are attempting unencrypted TCP communication".

src/code42cli/logger/handlers.py
unparalleled-js: everything looks good, I have tested all cases on both mac and windows

tests/logger/test_handlers.py
```diff
@@ -49,10 +49,20 @@ def socket_mocks(mocker):
...
+
+@pytest.fixture()
+def broken_pipe_error(system_exception_info):
+    system_exception_info.return_value = (BrokenPipeError, None, None)
+    return system_exception_info
+
+
+@pytest.fixture()
+def connection_reset_error(system_exception_info):
+    system_exception_info.return_value = (ConnectionResetError, None, None)
```
unparalleled-js, line 64: This is the error I saw during Windows, so the test explicitly checks for that rather

alanag13 (approved):

PR #244, opened 2021-02-12
Nightly build posts results to slack app

peterbriggs42:
https://github.com/marketplace/actions/action-slack

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -33,35 +33,11 @@ jobs:
         env:
             SSH_AUTH_SOCK: /tmp/ssh_agent.sock
         run: tox -e nightly  # Run tox using latest master branch from py42/c42eventextractor
-      # - name: Notify Slack Action
-      #   uses: ravsamhq/notify-slack-action@v1.1
-      #   if: always()
-      #   with:
-      #     status: ${{ job.status }}
-      #     notify_when: 'failure'
-      #   env:
-      #     SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
-      #
-      # - name: Notify Slack Action (v2)
-      #   uses: rtCamp/action-slack-notify@v2
-      #   env:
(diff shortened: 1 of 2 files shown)


.github/workflows/nightly.yml
```diff
@@ -33,3 +33,12 @@ jobs:
         env:
             SSH_AUTH_SOCK: /tmp/ssh_agent.sock
         run: tox -e nightly  # Run tox using latest master branch from py42/c42eventextractor
+      - name: Notify Slack Action
+        uses: ravsamhq/notify-slack-action@v1.1
+        if: always()
+        with:
+          status: ${{ job.status }}
+          notify_when: 'success,failure,warnings'
```
alanag13, line 41: can we make this only post on failure? In general I think its a good idea to only put something in the alerts channel when action is needed.

peterbriggs42: Yeah I just added the other two statuses for testing purposes, to avoid needing to trigger a failure to verify the connection works. Will remove.

tests/test_bulk.py
```diff
@@ -70,7 +70,7 @@ def test_generate_template_cmd_factory_when_using_defaults_returns_expected_comm
 def test_run_bulk_process_calls_run(bulk_processor, bulk_processor_factory):
     errors.ERRORED = False
     run_bulk_process(func_with_one_arg, None)
-    assert bulk_processor.run.call_count
+    assert bulk_processor.run.call_count == 5
```
alanag13, line 73: flip this back so that the test is no longer failing and we should be good here!

.github/workflows/nightly.yml
peterbriggs42: We are using a different action now, but it is configured to only post to Slack in the event of an action failure.

unparalleled-js (approved):
Seems good to me! I like the different links in the alerts

Feel free to still get other devs opinions if you want.

alanag13 (approved):

PR #245, opened 2021-02-16
Chore/dataframe formatter handle nulls

timabrmsn:
This updates our DataFrameOutputFormatter to automatically convert "null" values in the dataframe (`None`, `numpy.NaN`) to empty strings when outputting to table or CSV. For json output pandas automatically converts these to a json `null`. 

I refactored it a bit to hopefully be a little clearer, and updated the tests to validate the expected result instead of a specific implementation detail.


src/code42cli/output_formats.py
```diff
@@ -81,41 +80,51 @@ def _requires_list_output(self):
...
+            defaults = {"index": False}
+            defaults.update(kwargs)
+            df = df.fillna("")
+            return df.to_string(**defaults)
 
-    def _format_output(self, output, *args, **kwargs):
-        self._output_args.update(kwargs)
-        return self._format_func(output, *args, **self._output_args)
+        else:
+            raise Exception(
```
unparalleled-js, line 121: Where does this get raised? Should this be a ` Code42CLIError` or a `click.Exception` ?

src/code42cli/output_formats.py
```diff
@@ -81,41 +80,51 @@ def _requires_list_output(self):
...
-                {
-                    "orient": "records",
-                    "lines": True,
-                    "index": True,
-                    "default_handler": str,
-                }
-            )
+    def get_formatted_output(self, df, **kwargs):
+        if self.output_format == OutputFormat.JSON:
+            defaults = {
```
unparalleled-js, line 89: could use a helper method `_get_json_defaults(lines=True)` to share this part with the `.RAW` part

src/code42cli/output_formats.py
alanag13: At the very least this should have a more specific exception type (doesn't have to be a custom exception necessarily) to make catching this specific error easier if we want

tests/test_output_formats.py
```diff
@@ -771,53 +773,61 @@ def test_security_data_output_format_has_expected_options():
...
+            {"string_column": "string2", "int_column": 43, "null_column": NaN},
+        ]
+    )
+
+    def test_format_when_none_passed_defaults_to_table(self):
+        formatter = DataFrameOutputFormatter(output_format=None)
+        assert formatter.output_format == OutputFormat.TABLE
+
+    def test_format_when_unknown_format_raises_cli_exception(self):
+        with pytest.raises(Exception):
```
alanag13, line 788: when we use a bare exception, anything could have gone wrong here, not necessarily the thing that we're testing went wrong. A more specific type of exception (doesn't necessarily have to be a custom exception) would improve this

src/code42cli/output_formats.py
timabrmsn: I did initially make it a Code42CLIError, but that caused a circular import. This should only get raised if someone makes a programming error somewhere, since the format options the user passes in _should_ be bounded by a `click.Choice` type. So we'd hopefully catch it in development/testing. 

Previously we were not raising anything if an unknown value was passed as the format, and just defaulting to Table. I figured it would be better to raise to make it clear where the problem was than have everything still function but maybe not do what was expected, for our own debugging/sanity.

The ExceptionHandlingGroup does catch all exceptions raised and if any are not known it just prints "unknown error, see logs" and logs the full stack trace. If we use a ClickException the message will get printed, but no other details logged.

tests/test_output_formats.py
```diff
@@ -771,53 +773,61 @@ def test_security_data_output_format_has_expected_options():
...
+            {"string_column": "string1", "int_column": 42, "null_column": None},
+            {"string_column": "string2", "int_column": 43, "null_column": NaN},
+        ]
+    )
+
+    def test_format_when_none_passed_defaults_to_table(self):
+        formatter = DataFrameOutputFormatter(output_format=None)
+        assert formatter.output_format == OutputFormat.TABLE
+
+    def test_format_when_unknown_format_raises_cli_exception(self):
```
alanag13, line 787: the expected type of exception raised called out by the name of this test should be changed to `ValueError`.

alanag13 (approved):

unparalleled-js (approved):

Issue #205, opened 2021-01-20
[Enhancement] Updating profile does not know to use the default profile when not specifying `-n`


unparalleled-js:
## Summary
`code42 profile update` command requires `-n`.  Like other profile commands, when `-n` is not specified, it should know to use the default profile.

Also, if the user only wishes to change a single property, they still have to specify all of the other ones.  For example, if I just want to change my server address, I have to still supply my username.

## Proposed API

Allow command:

`code42 profile update -s 127.0.0.1:4200`

that only updates the server address for the default profile.

## Intended Use Case
To be able to easily change the server address.

3nin6:
Should the server and username options also be made required=False under the similar assumption that if the flag is not set, then use what already exists?

PR #246, opened 2021-02-17
feature/205-profile-update

3nin6:
This PR aims to resolve issue #205 

Args `--name --server --username` are no longer `required=True`. In order to prevent this from allowing users to create profiles with missing information a catch was added to create_profile.

### DIFF
--- a/src/code42cli/profile.py
+++ b/src/code42cli/profile.py
@@ -102,8 +102,6 @@ def switch_default_profile(profile_name):
 def create_profile(name, server, username, ignore_ssl_errors):
     if profile_exists(name):
         raise Code42CLIError("A profile named '{}' already exists.".format(name))
-    if None in {name, server, username}:
-        raise Code42CLIError("You must provide a profile name, server, and username.")
     config_accessor.create_profile(name, server, username, ignore_ssl_errors)
 
 
--- a/src/code42cli/profile.py
+++ b/src/code42cli/profile.py
@@ -102,7 +102,8 @@ def switch_default_profile(profile_name):
 def create_profile(name, server, username, ignore_ssl_errors):
     if profile_exists(name):
         raise Code42CLIError("A profile named '{}' already exists.".format(name))
-
+    if None in {name, server, username}:
+        raise Code42CLIError("You must provide a profile name, server, and username.")
     config_accessor.create_profile(name, server, username, ignore_ssl_errors)
 
 
--- a/tests/cmds/test_profile.py
+++ b/tests/cmds/test_profile.py
@@ -254,6 +254,34 @@ def test_update_profile_updates_existing_profile(
     )
 
 
+def test_update_profile_updates_default_profile(
+    runner, mock_cliprofile_namespace, user_agreement, valid_connection, profile
+):
+    name = "foo"
+    profile.name = name
+    mock_cliprofile_namespace.get_profile.return_value = profile
+    runner.invoke(
+        cli, ["profile", "update", "-s", "bar", "-u", "baz", "--disable-ssl-errors",],
+    )
+    mock_cliprofile_namespace.update_profile.assert_called_once_with(
+        name, "bar", "baz", True
+    )
+
+
+def test_update_profile_updates_name_alone(
+    runner, mock_cliprofile_namespace, user_agreement, valid_connection, profile
+):
+    name = "foo"
+    profile.name = name
+    mock_cliprofile_namespace.get_profile.return_value = profile
+    runner.invoke(
+        cli, ["profile", "update", "-u", "baz", "--disable-ssl-errors",],
+    )
+    mock_cliprofile_namespace.update_profile.assert_called_once_with(
+        name, None, "baz", True
+    )
+
+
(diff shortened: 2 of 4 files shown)


3nin6:

src/code42cli/profile.py
```diff
@@ -102,7 +102,8 @@ def switch_default_profile(profile_name):
 def create_profile(name, server, username, ignore_ssl_errors):
     if profile_exists(name):
         raise Code42CLIError("A profile named '{}' already exists.".format(name))
-
+    if None in {name, server, username}:
```
unparalleled-js, line 105: This causes the `[required]` documentation help text to go away from the `create` command, where these are actually required. An alternative approach that will allow the help text to remain accurate would be to use parameterized methods for creating the `click.options`.

Example:

```
def usename_option(required=False):
  return click.option(
    "-u", "--username", required=required, help="The username of the Code42 API user.",
```

Then, in the update function, use it like this:

```
@profile.command()
...
@username_option()
def update()...
```

And in the create method,

```
@profile.command()
...
@username_option(required=True)
def create()...
```

unparalleled-js:
Thank you so much for this contribution!

A couple of things that will be needed before merging are tests and a CHANGELOG entry. Let me know if you any help with those.

I also added a comment on an alternative way to achieve this without sacrificing the parts of the help text that specify which args are required (for the create command).

3nin6:
Thank you for the feedback, I'll look at making changes tonight to improve the PR.

3nin6:
Hi @unparalleled-js , I have made the changes for the functionality and added a preliminary changelog entry based off of others PRs I checked. I am having trouble getting a test to check this without making style checks fail permanently - they make the changes every time I run tox.

first pass at a test for this is a clone of the updates_existing_profile test it but without passing the `-n` option:

`def test_update_profile_updates_default_profile(
    runner, mock_cliprofile_namespace, user_agreement, valid_connection, profile
):
    name = "foo"
    profile.name = name
    mock_cliprofile_namespace.get_profile.return_value = profile
    runner.invoke(
        cli,
        [
            "profile",
            "update",
            "-s",
            "bar",
            "-u",
            "baz",
            "--disable-ssl-errors",
        ],
    )
    mock_cliprofile_namespace.update_profile.assert_called_once_with(
        name, "bar", "baz", True
    )
`

unparalleled-js:
> Hi @unparalleled-js , I have made the changes for the functionality and added a preliminary changelog entry based off of others PRs I checked. I am having trouble getting a test to check this without making style checks fail permanently - they make the changes every time I run tox.
> 
> first pass at a test for this is a clone of the updates_existing_profile test it but without passing the `-n` option:
> 
> ```
> def test_update_profile_updates_default_profile(
>     runner, mock_cliprofile_namespace, user_agreement, valid_connection, profile
> ):
>     name = "foo"
>     profile.name = name
>     mock_cliprofile_namespace.get_profile.return_value = profile
>     runner.invoke(
>         cli,
>         [
>             "profile",
>             "update",
>             "-s",
>             "bar",
>             "-u",
>             "baz",
>             "--disable-ssl-errors",
>         ],
>     )
>     mock_cliprofile_namespace.update_profile.assert_called_once_with(
>         name, "bar", "baz", True
>     )
> ```

Correct me I am not understand correctly, but

If you run `tox -e style`, it will correct any style issues.

The test you provided here looks like the right idea.  We could add another test as well that checks that `code42 profile update -u <username>` works `test_update_profile_updates_username_alone()`

3nin6:
I have committed the tests following running `tox` and also just `tox -e style` for review. I received the same failure+correction diff each time I ran it.

alanag13:
> If you run `tox -e style`, it will correct any style issues.

flake8 issues get checked with `tox -e style` as well, and those need to be corrected manually.
In your case, the output shows:

```
tests/cmds/test_profile.py:264:84: E231 missing whitespace after ','
tests/cmds/test_profile.py:278:71: E231 missing whitespace after ','
```

It looks like in both cases this is caused by an unnecessary trailing comma at the end of a list.

tests/cmds/test_profile.py
```diff
@@ -254,6 +254,34 @@ def test_update_profile_updates_existing_profile(
...
 
 
+def test_update_profile_updates_default_profile(
+    runner, mock_cliprofile_namespace, user_agreement, valid_connection, profile
+):
+    name = "foo"
+    profile.name = name
+    mock_cliprofile_namespace.get_profile.return_value = profile
+    runner.invoke(
+        cli, ["profile", "update", "-s", "bar", "-u", "baz", "--disable-ssl-errors",],
```
alanag13, line 264: here and only line 278 removing the last `,` before the closing `]` should resolve the style errors.

3nin6: Thank you for pointing that out, I didn't realize the style corrections weren't fully automated.

unparalleled-js (approved):

tests/cmds/test_profile.py
unparalleled-js: Thanks Alan

PR #247, opened 2021-02-18
Fix CLI FileCategory choice breakage

timabrmsn:
Works around file category camelcase conversion to allow backwards compatible values by creating a new `click.Choice` subclass that can take a mapping of additional "valid" keys that map back to the listed choice values.


src/code42cli/click_ext/types.py
```diff
@@ -116,3 +116,19 @@ def _get_dt_from_date_time_pair(date, time):
             raise BadParameter("Unable to parse date string: {}.".format(date_string))
         else:
             return dt
+
+
+class MapChoice(click.Choice):
+    """Choice subclass that takes an extra map of additional 'valid' keys to map to correct
+    choices list, allowing backward compatible choice changes. The extra keys don't show up
```
unparalleled-js, line 123: Would it make more sense to have the extra keys be the ones that show up in the help text and to just have this be a way to make things backwards compatible more silently?

I guess I could go either way, just curious if you thought about that

timabrmsn: I did, but if a new category gets added, it will _only_ be camelcase in py42 going forward, so we'd have to keep adding new upper-case maps as well to keep it consistent.

src/code42cli/cmds/securitydata.py
```diff
@@ -97,7 +98,25 @@
...
+            "PDF": "Pdf",
+            "PRESENTATION": "Presentation",
+            "SCRIPT": "Script",
+            "SOURCE_CODE": "SourceCode",
+            "SPREADSHEET": "Spreadsheet",
+            "VIDEO": "Video",
+            "VIRTUAL_DISK_IMAGE": "VirtualDiskImage",
+            "ARCHIVE": "Archive",
+            "ZIP": "Archive",
+            "Zip": "Archive",
```
timabrmsn, line 117: These didn't exist in the CLI, previously but since the Console text is "Zip", and py42's attribute value is ".ZIP" I figured it can't hurt to let those just work.

src/code42cli/cmds/securitydata.py
```diff
@@ -97,7 +98,25 @@
 file_category_option = click.option(
     "--file-category",
     multiple=True,
-    type=click.Choice(list(FileCategory.choices())),
+    type=MapChoice(
+        choices=list(FileCategory.choices()),
+        extras_map={
```
unparalleled-js, line 103: I got really annoyed with this similarly in the XSOAR integration and just made it always work with any cases or underscore/hyphen combo using this logic: https://github.com/code42/content/blob/master/Packs/Code42/Integrations/Code42/Code42.py#L530-L544

```
def get_file_category_value(key):
    # Meant to handle all possible cases
    key = key.lower().replace("-", "").replace("_", "")
    category_map = {
        "sourcecode": FileCategory.SOURCE_CODE,
        "audio": FileCategory.AUDIO,
        "executable": FileCategory.EXECUTABLE,
        "document": FileCategory.DOCUMENT,
        "image": FileCategory.IMAGE,
        "pdf": FileCategory.PDF,
        "presentation": FileCategory.PRESENTATION,
        "script": FileCategory.SCRIPT,
        "spreadsheet": FileCategory.SPREADSHEET,
        "video": FileCategory.VIDEO,
        "virtualdiskimage": FileCategory.VIRTUAL_DISK_IMAGE,
        "archive": FileCategory.ZIP,
    }
    return category_map.get(key, "UNCATEGORIZED")
```
Just thought I'd share because this way we are resilient to any mapping changes.  There have been breaking changes in that past that broke the XSOAR integration on this specific mapping

src/code42cli/cmds/securitydata.py
```diff
@@ -97,7 +98,25 @@
...
+            "AUDIO": "Audio",
```
unparalleled-js, line 104: If these mapping are already in the latest py42, you could use the actual constants
`FileCategory.DOCUMENT` etc

unparalleled-js: We actually may need a release in order for that to happen

src/code42cli/cmds/securitydata.py
unparalleled-js, line 117: https://github.com/code42/code42cli/pull/247/files#r578513304
^ Something like this will make it so we don't have to include multiple entries for the same value

src/code42cli/cmds/securitydata.py
timabrmsn: Ah, true. Then if (god forbid) they ever change again, it'll just work (although we'd need a new set of mappings for the now changed-again values).

src/code42cli/cmds/securitydata.py
alanag13: this approach also gives some additional flexibility like allowing `pdf` as valid in addition to `PDF` and `Pdf`.

src/code42cli/click_ext/types.py
alanag13: agree with @timabrmsn , we do not have to support both casings in newer categories that never used the old casing.

kiran-chaudhary:
Imo, tests should fail often with changes, if we try to cover all possible combinations we kind of hide the failures.

src/code42cli/cmds/securitydata.py
timabrmsn: this would also mean `__p__d__f___` will be a valid value too, is that ok?

alanag13: I don't see why not, unless we think that there's a chance that could become a name for a different category.

alanag13: i.e. would we ever have two categories where the only difference in the names is the presence of underscores?

src/code42cli/cmds/securitydata.py
alanag13: we do need a version bump / new py42 release, yeah -- that's why the tests are failing on this PR.

src/code42cli/cmds/securitydata.py
unparalleled-js: Lol. Would multiple underscores/hyphens really work? Values like `P-D_F` would work and that is odd but would have to be _discovered_ accidentally or something and even then, wouldn't have too surprising of an effect.

But would multiple underscores/hyphens really work with this advanced mapper suggestion? That would mean values like `P-_---_--Df` would work?

alanag13: actually, even then it wouldn't matter because that would be a new category that we wouldn't do this mapping on

unparalleled-js: Oh wait I see now how it works.

unparalleled-js:
> Imo, tests should fail often with changes, if we try to cover all possible combinations we kind of hide the failures.

That is good point. I think I have just been burned by this specific mapping two times now. But I agree it does enlighten us more when it happens.

src/code42cli/cmds/securitydata.py
timabrmsn: If we normalize and allow any casing/mix of underscores here, we should probably do it everywhere, no? It'd be odd to allow `sourcecode` on this one option when the choice lists `SourceCode` but not allow `--type applicationread`.

alanag13: @timabrmsn standardizing that sounds like a separate effort -- let's keep this as-is for now since it addresses the main problem.

alanag13 (approved):

PR #250, opened 2021-02-23
Raise usage error when updating profile does not receive any argument

unparalleled-js:
Error when `profile update` does not receive any arguments

Will take care of some feedback from QA

### DIFF
--- a/src/code42cli/cmds/profile.py
+++ b/src/code42cli/cmds/profile.py
@@ -60,6 +60,7 @@ disable_ssl_option = click.option(
     is_flag=True,
     help="For development purposes, do not validate the SSL certificates of Code42 servers. "
     "This is not recommended, except for specific scenarios like testing.",
+    default=None
 )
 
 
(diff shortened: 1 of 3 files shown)


src/code42cli/cmds/profile.py
```diff
@@ -104,11 +105,19 @@ def create(name, server, username, password, disable_ssl_errors):
 def update(name, server, username, password, disable_ssl_errors):
     """Update an existing profile."""
     c42profile = cliprofile.get_profile(name)
+
+    if not server and not username and not password and disable_ssl_errors is None:
+        raise click.UsageError(
+            "Must provide either `--username`, `--server`, `--password`, or `--disable-ssl-errors` "
```
alanag13, line 111: `Must provide at least one of...` ?

src/code42cli/cmds/profile.py
```diff
@@ -60,6 +60,7 @@ def username_option(required=False):
     is_flag=True,
     help="For development purposes, do not validate the SSL certificates of Code42 servers. "
     "This is not recommended, except for specific scenarios like testing.",
+    default=None,
```
unparalleled-js, line 63: When creating a new profile, it still defaults to `false` with this change..
In fact, this is actually a bugfix because it used to override `true`to false if excluded and updating profile.... Migt need to verify that and possibly include in the changelog

unparalleled-js:
Note that the current CHANGELOG entry also basically overs this change

unparalleled-js:
I get a test that fails locally but no in CI/CD

src/code42cli/cmds/profile.py
unparalleled-js: Yes, I just verified... This a bug but it is not a released bug

unparalleled-js: Reproduce the bug from master branch by doing:

```
code42 profile update --disable-ssl-errors
code42 profile show
# notice you disable-ssl-errors set to True

code42 profile update -u username
# say no to updating password
code42 profile show

# Notice you have disable-ssl-errors set to False
```

unparalleled-js: Actually, it might in sense be in the released version... will verify

alanag13 (approved):
Talked offline, there is a separate PR open to address the flaky test in question.

PR #253, opened 2021-03-02
hooking up integration tests with mock servers

unparalleled-js:


.github/workflows/build.yml
```diff
@@ -25,8 +25,30 @@ jobs:
...
+        run: cd code42cli; tox -e py  # Run tox using the version of Python in `PATH`
       - name: Submit coverage report
         uses: codecov/codecov-action@v1.0.7
         with:
-          file: ./coverage.xml
+          file: code42cli/coverage.xml
+      - name: Checkout mock servers
+        uses: actions/checkout@v2
+        with:
+          ref: network
```
unparalleled-js, line 36: TODO: remove this once https://github.com/code42/code42-mock-servers/pull/9 merges

tests/integration/test_alerts.py
```diff
@@ -24,16 +24,13 @@ def test_alerts_search_command_returns_success_return_code(
 
 
 @pytest.mark.integration
-@pytest.mark.parametrize(
```
unparalleled-js, line 27: Was seeing random send-to tests fail all the time.
I tried 3 things to fix it with no avail:

1.) sleeping everywhere possible, even large amounts of time

2.) Using different ports per instance of the ncat server

The error is that the connection is refused or breaks somehow.

unparalleled-js: I know at one point we said that testing both TCP and UDP is superfluous so I opted to remove these tests for now...

unparalleled-js: If anyone knows why these tests fail like this, please enlighten me and I can try to work it out!

timabrmsn: Do we need to install nmap in the image?

```
FileNotFoundError: [Errno 2] No such file or directory: 'ncat': 'ncat'
```

unparalleled-js: Yes, somehow, I forgot about that, will try to do this now

unparalleled-js: The failing tests I was describing was happening locally

timabrmsn: Yeah, they're intermittently failing for me to locally. Investigating.

unparalleled-js: If the `UDP` test executes before the `TCP`, it seems to pass

unparalleled-js: Somehow the tests are still failing here, so weird. They seem to pass consistently locally now at least...

unparalleled-js: It seems it either has to be `yield_fixtures` or there is something else going on...

timabrmsn:
I think I figured out these failures. I was able to reproduce a few times locally, and it always failed on the very first test that used the `tcp_dataserver` fixture, so I added a 1 sec sleep on the __enter__ method to make sure there's enough wiggle room to make sure the process is up and actually listening before any tests run. 

I triggered the build actions manually about 5-6 times after the change and they all built ok, so I think that was probably it.

kiran-chaudhary (approved):

timabrmsn (approved):

PR #254, opened 2021-03-02
Fix intermittent send-to integration failures

timabrmsn:
This change moves the DataServers into session-scoped fixtures with different ports to prevent intermittent connection problems, since they'll only be setup/torn down once per run.

### DIFF
--- a/tests/integration/test_auditlogs.py
+++ b/tests/integration/test_auditlogs.py
@@ -5,7 +5,6 @@ from shlex import split as split_command
 import pytest
 from tests.integration.conftest import append_profile
 from tests.integration.util import assert_test_is_successful
-from tests.integration.util import DataServer
 
 from code42cli.main import cli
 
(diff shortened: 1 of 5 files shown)


unparalleled-js (approved):
Thanks - I was trying to do this same thing but I think the scope-correction fixtures was the missing key

tests/integration/test_auditlogs.py
```diff
@@ -17,17 +16,24 @@
 
 
 @pytest.mark.integration
-@pytest.mark.parametrize(
-    "protocol", ["TCP", "UDP"],
-)
-def test_auditlogs_send_to_command_returns_success_return_code(
-    runner, integration_test_profile, protocol
+def test_auditlogs_send_to_tcp_command_returns_success_return_code(
+    runner, integration_test_profile, tcp_dataserver
```
unparalleled-js, line 20: Do actually need to pass this in or will it work without it?

unparalleled-js: I believe it should work without passing them in as fixture params

unparalleled-js: The `tcp_dataserver` and `udp_dataserver` fixtures are what I am referring to

timabrmsn: If no tests use the fixture, it won't get set up unless we set `autouse=True` on it. Which we don't want, since we only want it to get set up on integration tests.

PR #262, opened 2021-03-29
Feature/select profile

unparalleled-js:
Allows you to select a profile rather than having to first list all your profiles and then select the one you want (in the case where you forgot the name).

`code42 profile select`

vs

`code42 profile list` + `code42 profile use`

__________

This was a fun thing I made because I kept forgetting exact names of my profiles and I wanted to switch from Core Int to Console to Mock Servers etc.

Since it is rather simple and I use it myself, I figured I'd add a couple tests and open PR.

However, this _does_ take on a dependency of `PyInquirer` (which we could potentially use elsewhere).. If we are not comfortable doing that, I can keep this to myself..  I would completely understand.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 The intended audience of this file is for py42 consumers -- as such, changes that don't affect
 how a consumer would use the library (e.g. adding unit tests, updating documentation, etc) are not captured here.
 
+## Unreleased
+
+### Added
+
+- `code42 profile select` command that allows you to select a default profile
+    from a list of all your profiles.
+
 ## 1.4.0 - 2021-03-09
 
 ### Added
--- a/setup.py
+++ b/setup.py
@@ -39,6 +39,7 @@ setup(
         "keyrings.alt==3.2.0",
         "pandas>=1.1.3",
         "py42>=1.11.1",
+        "PyInquirer>=1.0.3",
     ],
     extras_require={
         "dev": [
--- a/src/code42cli/cmds/profile.py
+++ b/src/code42cli/cmds/profile.py
@@ -10,6 +10,7 @@ from code42cli.options import yes_option
 from code42cli.profile import CREATE_PROFILE_HELP
 from code42cli.sdk_client import validate_connection
 from code42cli.util import does_user_agree
+from code42cli.util import get_user_selected_item
 
 
 @click.group()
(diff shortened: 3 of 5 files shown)


alanag13:
If we can accomplish the same thing without a dependency (and hopefully without errors that happen when you click in the terminal), I'm in favor of that approach. I like the functionality of having this happen when you say you call `use` without any arguments, but we could also allow a `select` command since that verbiage feels a little more natural.

I think we should be conservative about where we apply this feature. It would be easy to get carried away and try to use it for every cli flag / command whose choices are limited to a list. For example, In detection lists, a user could have thousands of options for what user to remove from a list. What would that experience look like? I think it is best to limit this feature to instances where the set of choices are likely to be very small, and profile selection is a very good fit for that.

unparalleled-js:
Thanks for the feedback everyone!

Seems like a fair workaround! I like that we can achieve close to the same thing without needing a fancy dependency.

I am not really prioritizing this work, after work sometime soon, when I have time, I will implement it as suggested.

src/code42cli/click_ext/types.py
```diff
@@ -132,3 +132,19 @@ def convert(self, value, param, ctx):
         if value in self.extras_map:
             value = self.extras_map[value]
         return super().convert(value, param, ctx)
+
+
+class PromptChoices(click.ParamType):
+    def __init__(self, choices):
```
unparalleled-js, line 138: Thanks @timabrmsn for this code

unparalleled-js:
The test here is kind of lacking the user-input portion.  I plan to try and add a better test.

timabrmsn:
Since now `select` and `use` are essentially the same when `use` has no argument, would it make sense to have _only_ `use` and just document in help text that no arg prompts from a list?

unparalleled-js:
> Since now `select` and `use` are essentially the same when `use` has no argument, would it make sense to have _only_ `use` and just document in help text that no arg prompts from a list?

Yeah `select` was intended to just be alias of use with no args, but yeah maybe it's overkill if the functionality is already in `use`

docs/commands/securitydata.rst
```diff
@@ -1,3 +1,3 @@
-.. click:: code42cli.cmds.securitydata:security_data
+.. click:: code42cli.cmds.security_data:security_data
```
unparalleled-js, line 1: Renamed this because for modules like `departing_employee.py`, we were using a `_` to deliminate where the `-` are in the command name, so just a little consistency improvement...

PR #263, opened 2021-04-01
Update gettingstarted.md

annie-payseur:
Removed info about not supporting SSO


annie-payseur:
Statement about SSO is still valid. Closing this PR.

PR #264, opened 2021-04-01
Add events command to legal-hold

maddie-vargo:
This PR addresses Phase II of #176 to allow users to list legal hold administrative events. This command includes options to filter the results based on a specific legal hold uid, a beginning timestamp, an end timestamp, and a list of event types. 

**Testing Procedure**
```
code42 legal-hold events
code42 legal-hold --matter-id <matter id> --begin 30d --end 04/01/2021 --event-type MembershipCreated 
```

**PR Checklist**


CHANGELOG.md
```diff
@@ -8,6 +8,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 The intended audience of this file is for py42 consumers -- as such, changes that don't affect
 how a consumer would use the library (e.g. adding unit tests, updating documentation, etc) are not captured here.
 
+## Unreleased
+
+### Added
+
+- `code42 legal-hold events` command:
```
unparalleled-js, line 15: Should the command be called `search-events` instead?

* It'd be a verb, like the rest of our commands
* It'd mirror the `file-events search` command

docs/userguides/legalhold.md
```diff
@@ -93,4 +93,12 @@ To view all custodians (including inactive) for a legal hold matter, enter
 
 `code42 legal-hold show <matterID> --include-inactive`
 
+### List legal hold events
+
+To view a list of legal hold administrative events, use the following command:
+
+`code42 legal-hold events`
+
+This command takes the optional filters of a specific matter uid, beginning time stamp, end time stamp, and event type.
```
unparalleled-js, line 102: I think `time stamp` should be ` word, like `timestamp`

src/code42cli/cmds/legal_hold.py
```diff
@@ -25,6 +27,18 @@
 _MATTER_KEYS_MAP["description"] = "Description"
 _MATTER_KEYS_MAP["creator_username"] = "Creator"
 _MATTER_KEYS_MAP["creationDate"] = "Creation Date"
+_EVENT_KEYS_MAP = OrderedDict()
```
unparalleled-js, line 30: We no longer need to use `OrderedDict` in the CLI, we no longer support any python version that requires that.
Regular `{}` syntax should suffice

unparalleled-js: The above `_MATTER_KEYS_MAP`  could be changed as well

src/code42cli/cmds/legal_hold.py
```diff
@@ -230,6 +275,24 @@ def _get_all_active_matters(sdk):
     return matters
 
 
+def _get_all_events(sdk, legal_hold_uid, begin_date, end_date, event_type):
```
timabrmsn, line 278: Can we move the filtering logic out into the top-level events command function? That way the name of this function is more accurate.

So it would be like:

```
events = _get_all_events(state.sdk, matter_id, begin, end)
if event_type:
    events = [e for e in events if e["eventType"] == event_type]
if events:
    formatter.echo_formatted_list(events)
```

src/code42cli/cmds/legal_hold.py
```diff
@@ -124,6 +138,37 @@ def show(state, matter_id, include_inactive=False, include_policy=False):
         echo("")
 
 
+@legal_hold.command()
+@click.option(
```
unparalleled-js, line 142: Could you add the `-m` as a short flag for `matter_id_option`?

src/code42cli/cmds/legal_hold.py
```diff
@@ -124,6 +138,37 @@ def show(state, matter_id, include_inactive=False, include_policy=False):
...
+    "--matter-id", type=str, help="Filter results by legal hold UID",
```
unparalleled-js, line 143: Small refactor opportunity: we have matter_id_option already defined in this module.  We could create a shared method

```
def matter_id_option(required, help):
  return click.option(
    "-m",
    "--matter-id",
    required=required,
    type=str,
    help=help,
)
```

And use that for all matter_id options and then only have set two fields

src/code42cli/cmds/legal_hold.py
```diff
@@ -124,6 +138,37 @@ def show(state, matter_id, include_inactive=False, include_policy=False):
...
+            "MembershipCreated",
+            "MembershipReactivated",
+            "MembershipDeactivated",
+            "HoldCreated",
+            "HoldDeactivated",
+            "HoldReactivated",
+            "Restore",
+        ]
+    ),
+    help="Filter results by event types",
```
unparalleled-js, line 158: Missing `.` at end of help text

src/code42cli/cmds/legal_hold.py
unparalleled-js, line 143: Missing `.` at end of help text

src/code42cli/cmds/legal_hold.py
```diff
@@ -124,6 +138,37 @@ def show(state, matter_id, include_inactive=False, include_policy=False):
...
+)
+@click.option("--begin", **BEGIN_DATE_DICT)
```
unparalleled-js, line 160: This should use the `begin_option` defined in `code42cli.options`
There will be several advantages to doing this.

unparalleled-js: Same with `end_option`

src/code42cli/cmds/legal_hold.py
```diff
@@ -124,6 +138,37 @@ def show(state, matter_id, include_inactive=False, include_policy=False):
...
+@click.option("--end", **END_DATE_DICT)
+@format_option
+@sdk_options()
+def events(state, matter_id, event_type, begin, end, format):
+    """Report on legal hold events"""
```
unparalleled-js, line 165: Missing period after help text

`Report on legal hold events.`

src/code42cli/cmds/legal_hold.py
```diff
@@ -124,6 +138,37 @@ def show(state, matter_id, include_inactive=False, include_policy=False):
...
+    formatter = OutputFormatter(format, _EVENT_KEYS_MAP)
+    events = _get_all_events(state.sdk, matter_id, begin, end, event_type)
+    if events:
+        formatter.echo_formatted_list(events)
```
unparalleled-js, line 169: `echo_formatted_list` works well with small lists, but we should handle larger lists via a pager.

Like this:

```
        if len(events) > 10:
            output = formatter.get_formatted_output(events)
            click.echo_via_pager(output)
        else:
            formatter.echo_formatted_list(events)
```

tests/cmds/test_legal_hold.py
```diff
@@ -212,6 +246,21 @@ def active_and_inactive_legal_hold_memberships_response(mocker):
...
+    return _create_py42_response(mocker, EVENTS_RESPONSE)
+
+
+@pytest.fixture
+def empty_events_response(mocker):
+    return _create_py42_response(mocker, EMPTY_EVENTS_RESPONSE)
+
+
+@pytest.fixture
+def all_events_response(mocker):
```
unparalleled-js, line 260: It is better to have mock generator methods actually be generators and not just lists.  I know it probably works fine, but I have seen bugs show up because the tests were doing this. There are some subtle differences.

tests/cmds/test_legal_hold.py
```diff
@@ -575,6 +629,28 @@ def test_list_with_csv_format_returns_no_response_when_response_is_empty(
...
+def test_events_shows_events_that_respect_type_filters(
+    runner, cli_state, get_all_events_success, all_events_response
+):
+
+    cli_state.sdk.legalhold.get_all_events.return_value = all_events_response
+    result = runner.invoke(
+        cli, ["legal-hold", "events", "--event-type", "HoldCreated"], obj=cli_state
+    )
+
+    assert "564564654566" in result.output
```
unparalleled-js, line 641: From the test alone, it is not immediately clear why these values are expected.
Maybe a comment saying `// From `all_events_response` would help

tests/cmds/test_legal_hold.py
```diff
@@ -575,6 +629,28 @@ def test_list_with_csv_format_returns_no_response_when_response_is_empty(
...
+
+def test_events_with_csv_returns_no_events_when_response_is_empty(
+    runner, cli_state, get_all_events_success, empty_events_response
+):
+    cli_state.sdk.legalhold.get_all_events.return_value = empty_events_response
+    result = runner.invoke(cli, ["legal-hold", "events", "-f", "csv"], obj=cli_state)
+
+    assert "Matter ID,Name,Description,Creator,Creation Date" not in result.output
+
+
```
unparalleled-js, line 653: Could we get test(s) for `begin_date` / `end_date` validation?
It could be to just assert the values were passed into py42 correctly.

src/code42cli/cmds/legal_hold.py
maddie-vargo: Sure, I avoided this option initially because the `begin` option requires a cursor initialized for state, which I didn't think was necessary. I've added a LegalHoldEvents cursor to the cursor_store.py to make it work, but not sure if that's appropriate.

timabrmsn: I wonder if we should implement the `--use-checkpoint` option for this command and probably also make a `send-to`. As I could see customers wanting to be able to automate sending this data into a SIEM.

unparalleled-js: Oh!  @maddie-vargo Sorry, I did not forget about that when I made my initial comments.

I do like the idea from @timabrmsn, supporting checkingpointing would be a nice feature. We can always add it later though

CHANGELOG.md
maddie-vargo: Changed to `search-events` > let me know if this is not what you had in mind

tests/cmds/test_legal_hold.py
maddie-vargo: Let me know if I'm still missing the mark here.

tests/cmds/test_legal_hold.py
maddie-vargo: Defined variables at the top here. I don't love the names

```
OLDER_LEGAL_HOLD_CREATED_EVENT = "564564654566"
NEWER_LEGAL_HOLD_MEMBERSHIP_EVENT = "74533457745"
```

src/code42cli/cmds/legal_hold.py
maddie-vargo: @unparalleled-js  would it be possible to use the `begin_option` but somehow designate it as optional? I left the original options in there for now. 

I left checkpoint'ing off for now.

tests/cmds/test_legal_hold.py
maddie-vargo: I'll work on this. I had tried this initially, but had trouble getting a test to recognize the date inputs in the runner.

maddie-vargo: I added a test titled `test_search_events_is_called_with_expected_begin_timestamp`

src/code42cli/cmds/legal_hold.py
```diff
@@ -51,7 +69,10 @@ def legal_hold(state):
 
 
 @legal_hold.command()
-@matter_id_option
+@matter_id_option(
+    True,
+    "Identification number of the legal hold matter the custodian will be added to.",
```
unparalleled-js, line 74: @annie-payseur When you have chance, could you help review these option help texts? I will tag you in spots.

src/code42cli/cmds/legal_hold.py
```diff
@@ -124,6 +148,30 @@ def show(state, matter_id, include_inactive=False, include_policy=False):
...
+    "--event-type",
+    type=click.Choice(LEGAL_HOLD_EVENT_TYPES),
+    help="Filter results by event types.",
+)
+@click.option("--begin", **BEGIN_DATE_DICT)
+@click.option("--end", **END_DATE_DICT)
+@format_option
+@sdk_options()
+def search_events(state, matter_id, event_type, begin, end, format):
+    """Report on legal hold events."""
```
unparalleled-js, line 163: @annie-payseur This is the command's help text: `Report on legal hold events.`

tests/cmds/test_legal_hold.py
```diff
@@ -169,6 +173,75 @@
     ]
 }
 """
+EVENTS_RESPONSE = """
```
unparalleled-js, line 176: Something seems off about this response.... It is not valid JSON. Should these objects be under a list with key: `legalHoldEvents`

unparalleled-js: Or else I believe line 290 will create a response where this a `str` type (I think).

tests/cmds/test_legal_hold.py
```diff
@@ -212,6 +285,20 @@ def active_and_inactive_legal_hold_memberships_response(mocker):
     return [_create_py42_response(mocker, ALL_ACTIVE_AND_INACTIVE_CUSTODIANS_RESPONSE)]
 
 
+@pytest.fixture
+def events_response(mocker):
```
unparalleled-js, line 289: Is this fixture used anywhere? I am not seeing.
Perhaps this and `EVENTS_RESPONSE` can be removed?

tests/cmds/test_legal_hold.py
unparalleled-js: I don't think the fixture that uses this is ever used, if that is the, case can this be removed?

tests/cmds/test_legal_hold.py
unparalleled-js: The names are fine..

but alternatively they could be called 
`_CREATE_EVENT_ID` and `_MEMBERSHIP_EVENT_ID`

Reasons:
1.) they are already in a legal hold module so don't need the prefix,
2.) they can be internal since nothing should be importing or using them outside of this test module, and 
3.) ending with ID makes it more apparent that it's not a JSON object

tests/cmds/test_legal_hold.py
unparalleled-js: This works!
You could add more events to the list so that it yields more than once, that way it is more like what will happen in real life (unless there is often only a single event). Not a big deal though.

tests/cmds/test_legal_hold.py
maddie-vargo: I took your suggestions. Thanks!

tests/cmds/test_legal_hold.py
maddie-vargo: Yes, that is no longer needed, since I moved to a generator format for the mocker. I forgot to take it out on my last push along with it's corresponding fixture.

tests/cmds/test_legal_hold.py
maddie-vargo: Yes, this is no longer needed. See my comment on the item above.

src/code42cli/cmds/legal_hold.py
```diff
@@ -60,7 +81,10 @@ def add_user(state, matter_id, username):
 
 
 @legal_hold.command()
-@matter_id_option
+@matter_id_option(
+    True,
+    "Identification number of the legal hold matter the custodian will be removed from.",
```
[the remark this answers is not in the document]
annie-payseur: Looks good, thanks!

unparalleled-js (approved):

src/code42cli/cmds/legal_hold.py
annie-payseur: I suggest changing to `Tools for getting legal hold event data.` (which is more consistent with the help text for `security-data`)

maddie-vargo:
Does anything else need to happen on this PR, or can it be merged and closed?

unparalleled-js:
> Does anything else need to happen on this PR, or can it be merged and closed?

Was just waiting for a bugfix release to happen (just did). I can merge this now, but do you mind fixing the CHANGELOG one last time?  Just need to add your changed to a new Unreleased section (above the version section that was released today). After that, I will merge this in and get it officially tested and release it ASAP.

unparalleled-js (approved):

PR #265, opened 2021-04-02
Bugfix/attempt to autodetect file encodings when reading files

timabrmsn:
This arose from Pro Services attempting to deactivate devices running the commands on Windows:

```
code42 devices list --<some-filter-options> -f CSV > devices.csv
code42 devices bulk deactivate devices.csv
``

Powershell apparently writes the `> devices.csv` file as UTF-16 encoded, so the `csv_reader` function erroring saying the headers were wrong even though they appeared fine in a text editor. 

This change adds a new click type extension to wrap `click.File` with the `chardet` lib's `detect()` method (`chardet` is already a dependency of requests, so we're not adding anything new).

### DIFF
--- a/src/code42cli/click_ext/types.py
+++ b/src/code42cli/click_ext/types.py
@@ -17,7 +17,7 @@ class AutoDecodedFile(click.File):
             with open(value, "rb") as file:
                 self.encoding = chardet.detect(file.read())["encoding"]
             if self.encoding is None:
-                CliLogger.log_error(f"Failed to detect encoding of file: {value}")
+                CliLogger().log_error(f"Failed to detect encoding of file: {value}")
         finally:
             return super().convert(value, param, ctx)
 
--- a/src/code42cli/click_ext/types.py
+++ b/src/code42cli/click_ext/types.py
@@ -25,7 +25,7 @@ class AutoDecodedFile(click.File):
         return super().convert(value, param, ctx)
 
 
-class FileOrString(click.File):
+class FileOrString(AutoDecodedFile):
     """Declares a parameter to be a file (if the argument begins with `@`), otherwise accepts it as
     a string.
     """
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by unparalleled-js** [comment]
The method `code42cli.bulk.write_template_file` creates a bulk template file in utf8. I don't think that will break anything since the reads always check the encoding first, but pointing it out just in case.

File: src/code42cli/click_ext/types.py [hunk 1]
```diff
@@ -3,9 +3,27 @@
...
+class AutoDecodedFile(click.File):
+    """Attempts to autodetect file's encoding prior to normal click.File processing."""
+
+    def convert(self, value, param, ctx):
+        try:
+            with open(value, "rb") as file:
+                self.encoding = chardet.detect(file.read())["encoding"]
+            if self.encoding is None:
+                CliLogger().log_error(f"Failed to detect encoding of file: {value}")
+        except Exception:
```
unparalleled-js, line 22: Is this try / except block needed

timabrmsn: If the filename doesn't exist the `open()` will raise an exception.

unparalleled-js:
Missing CHANGELOG entry

src/code42cli/click_ext/types.py
kiran-chaudhary: Shouldn't we `raise` instead of `pass` if we intend `click.File` to do the exception handling.... by `pass` aren't we suppressing the exception!?

kiran-chaudhary: Would it be nice to add a (file not exists) test case to confirm if the exception cases works as expected!

timabrmsn: click.File does its own `open()` on the filename. We're only opening here to do the encoding detection, not actually using the result. 

The "doesn't exist" exception test case is a good call.

unparalleled-js (approved):

PR #269, opened 2021-04-20
Bugfix/Win-Newlines-and-use-pager

unparalleled-js:
* Fixes Windows bug where CSV outputs would have extra newlines between the output

^ Verify via `code42 high-risk-employee list -f csv >> test.csv` and then open the file (on Windows and Mac)

* Uses pager when outputting > 10 alert rules

^ Verify via `code42 alert-rules list`. Notice you can page.

* Uses pager when outputting > 10 case

^ Verify via `code42 cases list`. Notice you can page.

also updates the extractor to fix that weird timestamp bug

### DIFF
--- a/src/code42cli/cmds/cases.py
+++ b/src/code42cli/cmds/cases.py
@@ -151,10 +151,7 @@ def _list(
     formatter = OutputFormatter(format, _get_cases_header())
     cases = [case for page in pages for case in page["cases"]]
     if cases:
-        if len(cases) > 10:
-            click.echo_via_pager(formatter.get_formatted_output(cases))
-        else:
-            formatter.echo_formatted_list(cases)
+        formatter.echo_formatted_list(cases)
     else:
         click.echo("No cases found.")
 
--- a/src/code42cli/cmds/cases.py
+++ b/src/code42cli/cmds/cases.py
@@ -151,7 +151,10 @@ def _list(
     formatter = OutputFormatter(format, _get_cases_header())
     cases = [case for page in pages for case in page["cases"]]
     if cases:
-        formatter.echo_formatted_list(cases)
+        if len(cases) > 10:
+            click.echo_via_pager(formatter.get_formatted_output(cases))
+        else:
+            formatter.echo_formatted_list(cases)
     else:
         click.echo("No cases found.")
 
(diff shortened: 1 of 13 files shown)


src/code42cli/cmds/cases.py
```diff
@@ -151,7 +151,10 @@ def _list(
     formatter = OutputFormatter(format, _get_cases_header())
     cases = [case for page in pages for case in page["cases"]]
     if cases:
-        formatter.echo_formatted_list(cases)
+        if len(cases) > 10:
```
alanag13, line 154: is this logic everywhere where paging happens? could we abstract it somehow? For example, if we decided to increase the number of records required in order for paging to occur, how many places would we need to touch?

src/code42cli/cmds/devices.py
```diff
@@ -154,11 +155,11 @@ def _change_device_name(sdk, guid, name):
 @devices.command()
 @device_guid_argument
 @sdk_options()
-def show(state, device_guid, format=None):
+def show(state, device_guid):
     """Print individual device details. Requires device GUID."""
 
-    formatter = OutputFormatter(format, _device_info_keys_map())
-    backup_set_formatter = OutputFormatter(format, _backup_set_keys_map())
+    formatter = OutputFormatter(OutputFormat.TABLE, _device_info_keys_map())
```
alanag13, line 161: why is `TABLE` the only valid output format for this command?

unparalleled-js: show commands only output a single record and we don't have any in the CLI that allow other formats for show commands.

For some reason, there was a format param here but there was no `format_option` so it just always uses `None` (`TABLE`) anyway

src/code42cli/cmds/cases.py
unparalleled-js: If we ever do the ticket that switched the formatters to always use the `DataFrameFormatter`, it would be resolved. I am not sure of what all that entails though. In the short term, I can put something in

unparalleled-js: I just moved the pager logic to the `echo_formatted_output` anyway since it is always used at the same time

alanag13 (approved):

Issue #259, opened 2021-03-18
[Enhancement] add Users command


ceciliastevens:
## Summary
Code42 administrators occasionally need to manage users in bulk, and may not have SCIM providers or other methods. Even if SCIM is present, occasionally there are one-time cleanup jobs that would benefit from the bulk processing offered by the CLI, but not available in the web UI.

One way to address this need would be to add a `users` command that could accomplish tasks such as:

- List users visible to the user running the CLI
- List users who have a particular role
- Add or remove a role from a user
- List users by org
- List users who are licensed
- Bulk deactivate/reactivate users
- Bulk change usernames
- Bulk change organization
- Bulk block/unblock

## Proposed API

```
code42 users list --having-role "Customer Cloud Admin" --active-only --org-name "Example org" 
code42 users list --licensed --deactivated-only
code42 users bulk deactivate users_to_deactivate.csv
code42 users bulk block users_to_block.csv
code42 users bulk change-username users_to_rename.csv
code42 users bulk change-organization users_to_move.csv --target-org "Example org"
```

## Intended Use Case
There are a variety of use-cases for this command:

- Admins needing to do one-time cleanup or rename operations prior to introducing a SCIM provider
- Admins needing to audit who has particular roles in their environment
- Admins needing to determine who is consuming licenses and when
- Building temporary user access controls (i.e. granting a role, and then removing it after 10 minutes)

These are just a few examples from the top of my head, doubtless more is possible in the future.

alanag13:
This all seems like super useful functionality that I think we can safely say we would be happy to include, but I think this should probably be broken up into smaller enhancements -- otherwise this would be a monstrous PR.

ceciliastevens:
Yeah, that makes sense. Is there any particular way you'd prefer to have it broken up?

The one that seems clearest to me should be separate would be "list licensed users", as that's currently a standalone script and isn't especially straightforward.

alanag13:
I'd probably break it up like:
- non-license modifications to the `list` command
- licensed users
- bulk block / deactivate (should include unblock / reactivate, and non-bulk versions of these commands)
- bulk change username / org

alanag13:
the smaller the better (I wouldn't be upset a PR that did just block/unblock, for example), but the above is probably the largest each PR should reasonably be, imo.

ceciliastevens:
That makes sense. Should I file separate issues, or can we have multiple PRs all pointing back to (but not closing) this issue?

[...]

PR #270, opened 2021-04-21
List users

ceciliastevens:
**Description of Change**

This adds a "users list" command and parameters to filter on active status, org Uid, and role name.

This partially addresses #259, but more work will follow in additional PRs.

**Testing Procedure**

```
code42 users list
code42 users list --role-name "Customer Cloud Admin"
code42 users list --org-uid "12345"
code42 users list --inactive
code42 users list --active
```

**PR Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,9 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 ## Unreleased
 
-- `code42 users list` command
+### Added
+
+- New command `code42 users list` with options:
     - `--org-uid` filters on org membership.
     - `--role-name` filters on users having a particular role.
     - `--active` and `--inactive` filter on user status.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 ## Unreleased
 
+- `code42 users list` command
+    - `--org-uid` filters on org membership.
+    - `--role-name` filters on users having a particular role.
+    - `--active` and `--inactive` filter on user status.
+
 ### Fixed
 
 - Bug where some CSV outputs on Windows would have an extra newline between the rows.
--- a/src/code42cli/cmds/users.py
+++ b/src/code42cli/cmds/users.py
@@ -1,40 +1,38 @@
 import click
+from pandas import DataFrame
 
-from code42cli.options import format_option
-from code42cli.options import sdk_options
-from code42cli.output_formats import DataFrameOutputFormatter
 from code42cli.click_ext.groups import OrderedGroup
 from code42cli.click_ext.options import incompatible_with
 from code42cli.errors import Code42CLIError
+from code42cli.options import format_option
+from code42cli.options import sdk_options
+from code42cli.output_formats import DataFrameOutputFormatter
 from code42cli.output_formats import OutputFormatter
-from pandas import DataFrame
 
 
 @click.group(cls=OrderedGroup)
 @sdk_options(hidden=True)
 def users(state):
-	"""Manage users within your Code42 environment"""
-	pass
+    """Manage users within your Code42 environment"""
+    pass
+
 
 org_uid_option = click.option(
-	"--org-uid",
-	required=False,
+    "--org-uid",
+    required=False,
     type=str,
     default=None,
-    help="Limit users to only those in the organization you specify. "
+    help="Limit users to only those in the organization you specify. ",
 )
 role_name_option = click.option(
-	"--role-name",
-	required=False,
-	type=str,
-	default=None,
-	help="Limit results to only users having the specified role."
+    "--role-name",
+    required=False,
+    type=str,
+    default=None,
+    help="Limit results to only users having the specified role.",
 )
 active_option = click.option(
-    "--active",
-    is_flag=True,
-    help="Limits results to only active users.",
-    default=None,
+    "--active", is_flag=True, help="Limits results to only active users.", default=None,
 )
 inactive_option = click.option(
     "--inactive",
@@ -43,6 +41,7 @@ inactive_option = click.option(
     cls=incompatible_with("active"),
 )
 
+
 @users.command(name="list")
 @org_uid_option
 @role_name_option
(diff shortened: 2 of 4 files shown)


src/code42cli/cmds/users.py
```diff
@@ -0,0 +1,84 @@
...
+from code42cli.errors import Code42CLIError
+from code42cli.options import format_option
+from code42cli.options import sdk_options
+from code42cli.output_formats import DataFrameOutputFormatter
+
+
+@click.group(cls=OrderedGroup)
+@sdk_options(hidden=True)
+def users(state):
+    """Manage users within your Code42 environment"""
```
unparalleled-js, line 15: Descriptions should end in periods

src/code42cli/cmds/users.py
```diff
@@ -0,0 +1,84 @@
...
+    pass
+
+
+org_uid_option = click.option(
+    "--org-uid",
+    required=False,
+    type=str,
```
unparalleled-js, line 22: `type=str` is redundant

src/code42cli/cmds/users.py
```diff
@@ -0,0 +1,84 @@
...
+    default=None,
```
unparalleled-js, line 23: `default=None` is redundant

src/code42cli/cmds/users.py
```diff
@@ -0,0 +1,84 @@
...
+    help="Limit users to only those in the organization you specify. ",
```
unparalleled-js, line 24: Help text ends with extra unneeded space

src/code42cli/cmds/users.py
```diff
@@ -0,0 +1,84 @@
...
+)
+role_name_option = click.option(
+    "--role-name",
+    required=False,
```
unparalleled-js, line 28: `required=False` is redundant for options

src/code42cli/cmds/users.py
unparalleled-js: It is needed for the flag options only

src/code42cli/cmds/users.py
```diff
@@ -0,0 +1,84 @@
...
+def _get_role_id(sdk, role_name):
+    try:
+        roles_dataframe = DataFrame.from_records(sdk.users.get_available_roles().data)
+        return str(
+            roles_dataframe.loc[
+                roles_dataframe["roleName"] == role_name, "roleId"
+            ].array[0]
+        )
+    except KeyError:
+        raise Code42CLIError("Role with name {} not found.".format(role_name))
```
unparalleled-js, line 76: We can use f strings now
`f"Role with name {role_name} not found."

src/code42cli/cmds/users.py
```diff
@@ -0,0 +1,84 @@
...
+@role_name_option
+@active_option
+@inactive_option
+@format_option
+@sdk_options()
+def list_users(state, org_uid, role_name, active, inactive, format):
+    """List users in your Code42 environment."""
+    if inactive:
+        active = False
+    if role_name:
```
unparalleled-js, line 55: Could be a ternary here
`role_id = _get_role_id(state.sdk, role_name) if role_name else None`

tests/cmds/test_users.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+def test_get_users_dataframe_returns_dataframe(cli_state, get_all_users_success):
+    result = _get_users_dataframe(
+        cli_state.sdk, org_uid=None, role_id=None, active=None
+    )
+    assert isinstance(result, DataFrame)
+
+
+def test_get_users_dataframe_calls_get_all_users_with_correct_parameters(
```
unparalleled-js, line 79: We don't typically test the implementation details of commands (the protected methods).
Can we change these tests such that their entry point occurs from command execution?

CHANGELOG.md
```diff
@@ -10,6 +10,11 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 ## Unreleased
 
+- `code42 users list` command
```
unparalleled-js, line 13: This should be under the `Added` header and be phrased like this

### Added

- New command `code42 users list` with options:
...

src/code42cli/cmds/users.py
unparalleled-js, line 24: Does this include child orgs?

tests/cmds/test_users.py
```diff
@@ -0,0 +1,88 @@
...
+from code42cli.cmds.users import _get_role_id
+from code42cli.cmds.users import _get_users_dataframe
+from code42cli.main import cli
+
+TEST_ROLE_RETURN_DATA = {"data": [{"roleName": "Customer Cloud Admin", "roleId": "12"}]}
+
+TEST_USERS_RESPONSE = {
+    "users": [
+        {
+            "userId": "1234",
```
unparalleled-js, line 17: `userId` is a number and not a string on real responses I thought

src/code42cli/cmds/users.py
```diff
@@ -0,0 +1,84 @@
...
+        click.echo("No results found.")
+    else:
+        formatter = DataFrameOutputFormatter(format)
+        formatter.echo_formatted_dataframe(df)
+
+
+def _get_role_id(sdk, role_name):
+    try:
+        roles_dataframe = DataFrame.from_records(sdk.users.get_available_roles().data)
+        return str(
```
unparalleled-js, line 70: Lines 70 - 73 is quite complex. Can we break this into multiple statements?

tests/cmds/test_users.py
ceciliastevens: Updated!

src/code42cli/cmds/users.py
ceciliastevens: I believe so. I'll update the text.

tests/cmds/test_users.py
ceciliastevens: Yes, it is. Updated.

src/code42cli/cmds/users.py
ceciliastevens: I did break this out, and also found a way to do it using DataFrame.at[] instead of DataFrame.loc[]

tests/cmds/test_users.py
ceciliastevens: There are also a couple other tests (test_get_role_id_returns_appropriate_role_id and test_get_users_dataframe_returns_dataframe) that test the details of internal methods, should those be updated as well? I think it might be difficult to test some of those things from command execution as they aren't directly exposed in the result.

unparalleled-js: Yes, that would be nice. 

Tests like `test_get_users_dataframe_returns_dataframe` are not needed. The fact that we use dataframes under the hood shouldn't matter.

For testing `role_ids` logic, the test would look something this 

```
def test_list_users_includes_expected_role_ids()
```

And then you would verify that the role IDs from the mock responses have shown up in the output

```
assert "12" in result.output
```

(Mayne make the role ID more specific, like `1230123` or something)


Additionally, I would really appreciate some tests around the logic that figures out `active`.

```
def test_list_users_when_given_active_uses_inactivate_uses_active_equals_false()

def test_list_users_when_given_active_and_inactive_raises_error()

def test_list_users_when_given_excluding_active_and_inactive_uses_active_equals_none()
```

^ These tests can just verify that py42 was called correctly if we are unable to verify these states via the output.

ceciliastevens: I've added the additional tests for the active logic. I updated the role Id test as well, but roleId isn't actually included in the output, it's just passed as a parameter to the users.get_all() method that gets the users (where it's used as a filter). So I ended up just testing that the users.get_all() call has the appropriate parameter.

tests/cmds/test_users.py
```diff
@@ -0,0 +1,110 @@
...
+def test_list_users_when_given_active_and_inactive_raises_error(
+    runner, cli_state, get_available_roles_success, get_all_users_success
+):
+    result = runner.invoke(
+        cli, ["users", "list", "--active", "--inactive"], obj=cli_state
+    )
+    assert "Error: --inactive can't be used with: --active" in result.output
+
+
+def test_list_users_when_given_excluding_active_and_inactive_uses_active_equals_none(
```
unparalleled-js, line 104: technically don't need the word `_given_` in the test signature (not a blocker though)

unparalleled-js (approved):

PR #271, opened 2021-04-21
Bugfix/too many py42s

unparalleled-js:
*Fixes issue in de bulk add where we would instantiate a py42 per row in the csv file.

### DIFF
--- a/src/code42cli/cmds/departing_employee.py
+++ b/src/code42cli/cmds/departing_employee.py
@@ -66,8 +66,6 @@ def _list(state, format, filter):
 @sdk_options()
 def add(state, username, cloud_alias, departure_date, notes):
     """Add a user to the Departing Employees detection list."""
-    if departure_date:
-        departure_date = departure_date.strftime(DATE_FORMAT)
     _add_departing_employee(state.sdk, username, cloud_alias, departure_date, notes)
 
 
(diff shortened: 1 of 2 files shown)


src/code42cli/cmds/departing_employee.py
```diff
@@ -102,26 +100,21 @@ def bulk(state):
 )
 @read_csv_arg(headers=DEPARTING_EMPLOYEE_CSV_HEADERS)
 @sdk_options()
-@click.pass_context
-def bulk_add(ctx, state, csv_rows):
+def bulk_add(state, csv_rows):
+    sdk = state.sdk  # Force initialization of py42 to only happen once.
```
unparalleled-js, line 104: it seems as though all the other bulk commands already are doing this

src/code42cli/cmds/departing_employee.py
```diff
@@ -102,26 +100,21 @@ def bulk(state):
...
                 )
                 raise Code42CLIError(message)
-        ctx.invoke(
-            add,
-            username=username,
-            cloud_alias=cloud_alias,
-            departure_date=departure_date,
-            notes=notes,
-        )
+        _add_departing_employee(sdk, username, cloud_alias, departure_date, notes)
```
unparalleled-js, line 117: `ctx.invoke()` was the culprit here but I still like this better. I don't know what `ctx.invoke()` does behind the scenes and it seems like unnecessary to handle off the execution back to the context

unparalleled-js:
Should I include a changelog like `Fix performance issues with code42 departing-employee bulk add` or something?

alanag13 (approved):

alanag13:
> Should I include a changelog like `Fix performance issues with code42 departing-employee bulk add` or something?

Sure, that wouldn't hurt.

PR #274, opened 2021-04-23
Users fixes

unparalleled-js:
* Limits the columns for TABLE output for `list users` cmds.

### DIFF
--- a/src/code42cli/cmds/users.py
+++ b/src/code42cli/cmds/users.py
@@ -6,7 +6,8 @@ from code42cli.click_ext.options import incompatible_with
 from code42cli.errors import Code42CLIError
 from code42cli.options import format_option
 from code42cli.options import sdk_options
-from code42cli.output_formats import DataFrameOutputFormatter, OutputFormat
+from code42cli.output_formats import DataFrameOutputFormatter
+from code42cli.output_formats import OutputFormat
 
 
 @click.group(cls=OrderedGroup)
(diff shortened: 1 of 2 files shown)


src/code42cli/cmds/users.py
```diff
@@ -46,7 +47,12 @@ def list_users(state, org_uid, role_name, active, inactive, format):
     if inactive:
         active = False
     role_id = _get_role_id(state.sdk, role_name) if role_name else None
-    df = _get_users_dataframe(state.sdk, org_uid, role_id, active)
+    columns = (
+        ["userUid", "status", "username", "orgUid"]
+        if format == OutputFormat.TABLE
+        else None
```
unparalleled-js, line 53: None means all columns

timabrmsn:
One unrelated thing I found while testing this change was we don't cleanly handle a bad orgUid. The service returns a 400 we could handle better:
```
Failure in HTTP call 400 Client Error: Bad Request for url: https://www.crashplan.com/api/User?orgUid=10&pgNum=1&pgSize=500 

response.text: [{"name":"SYSTEM","description":"Organization was not found"}]
```

tests/cmds/test_users.py
```diff
@@ -61,6 +63,16 @@ def test_list_outputs_appropriate_columns(runner, cli_state, get_all_users_succe
     assert "modificationDate" in result.output
 
 
+def test_list_when_table_format_outputs_expected_columns(
```
timabrmsn, line 66: Should we test the columns we're filtering out _aren't_ in the output?

unparalleled-js:
> One unrelated thing I found while testing this change was we don't cleanly handle a bad orgUid. The service returns a 400 we could handle better:
> 
> ```
> Failure in HTTP call 400 Client Error: Bad Request for url: https://www.crashplan.com/api/User?orgUid=10&pgNum=1&pgSize=500 
> 
> response.text: [{"name":"SYSTEM","description":"Organization was not found"}]
> ```

Yeah @terimurie also point this out. I was thinking we should just add it to py42 and let it naturally propagate to the CLI

timabrmsn (approved):

PR #277, opened 2021-04-29
Feature/2fa

timabrmsn:
Implements support for MFA accounts.

When a profile for an MFA user is created, the user will see the message:
```
Multi-factor account detected. `--totp <token>` option will be required for all code42 invocations.
```

Then any commands that initialize the sdk will either require the `--totp <token>` option, or if the option isn't provided, the CLI will prompt for the token.

### DIFF
--- a/docs/userguides/gettingstarted.md
+++ b/docs/userguides/gettingstarted.md
@@ -72,9 +72,15 @@ python3 -m pip install code42cli --upgrade
 .. important:: The Code42 CLI currently only supports token-based authentication.
 ```
 
-Create a user in Code42 to authenticate (basic authentication) and access data via the CLI. The CLI returns data based on the roles assigned to this user. To ensure that the user's rights are not too permissive, create a user with the lowest level of privilege necessary. See our [Role assignment use cases](https://support.code42.com/Administrator/Cloud/Monitoring_and_managing/Role_assignment_use_cases) for information on recommended roles. We recommend you test to confirm that the user can access the right data.
+Create a user in Code42 to authenticate (basic authentication) and access data via the CLI. The CLI returns data based 
+on the roles assigned to this user. To ensure that the user's rights are not too permissive, create a user with the lowest 
+level of privilege necessary. See our [Role assignment use cases](https://support.code42.com/Administrator/Cloud/Monitoring_and_managing/Role_assignment_use_cases) 
+for information on recommended roles. We recommend you test to confirm that the user can access the right data.
 
 If you choose not to store your password in the CLI, you must enter it for each command that requires a connection.
+
+The Code42 CLI supports local accounts with MFA (multi-factor authentication) enabled. The Time-based One-Time 
+Password (TOTP) must be provided at every invocation of the CLI, either via the `--totp` option, or when prompted.
 
 The Code42 CLI currently does **not** support SSO login providers or any other identity providers such as Active
 Directory or Okta.
--- a/docs/userguides/gettingstarted.md
+++ b/docs/userguides/gettingstarted.md
@@ -72,14 +72,14 @@ python3 -m pip install code42cli --upgrade
 .. important:: The Code42 CLI currently only supports token-based authentication.
 ```
 
-Create a user in Code42 to authenticate (basic authentication) and access data via the CLI. The CLI returns data based 
-on the roles assigned to this user. To ensure that the user's rights are not too permissive, create a user with the lowest 
-level of privilege necessary. See our [Role assignment use cases](https://support.code42.com/Administrator/Cloud/Monitoring_and_managing/Role_assignment_use_cases) 
+Create a user in Code42 to authenticate (basic authentication) and access data via the CLI. The CLI returns data based
+on the roles assigned to this user. To ensure that the user's rights are not too permissive, create a user with the lowest
+level of privilege necessary. See our [Role assignment use cases](https://support.code42.com/Administrator/Cloud/Monitoring_and_managing/Role_assignment_use_cases)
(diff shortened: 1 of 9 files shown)


alanag13:
Can/does that message also get printed when the user switches to that account using `profile use`?

timabrmsn:
> Can/does that message also get printed when the user switches to that account using `profile use`?

It does not, as it only knows the user requires MFA when we validate the password. So I guess it also doesn't print if the user doesn't supply a password when creating the profile.

alanag13:
Maybe we could validate upon switching too? Might be nice to know if your credentials are no longer good when you switch anyway.

docs/userguides/gettingstarted.md
```diff
@@ -72,10 +72,16 @@ python3 -m pip install code42cli --upgrade
...
+Create a user in Code42 to authenticate (basic authentication) and access data via the CLI. The CLI returns data based
+on the roles assigned to this user. To ensure that the user's rights are not too permissive, create a user with the lowest
+level of privilege necessary. See our [Role assignment use cases](https://support.code42.com/Administrator/Cloud/Monitoring_and_managing/Role_assignment_use_cases)
+for information on recommended roles. We recommend you test to confirm that the user can access the right data.
 
 If you choose not to store your password in the CLI, you must enter it for each command that requires a connection.
 
+The Code42 CLI supports local accounts with MFA (multi-factor authentication) enabled. The Time-based One-Time
```
unparalleled-js, line 82: It would be great to be able to login only once. We have a story for this, right? For storing the token and only prompting when it expires

docs/userguides/profile.md
```diff
@@ -33,3 +33,9 @@ To see all your profiles, do:
 ```bash
 code42 profile list
 ```
+
+## Profiles with Multi-Factor Authentication
+
+If your Code42 user account requires multi-factor authentication, the token is not required to create your profile, but
```
unparalleled-js, line 39: comma not needed before `but`

src/code42cli/cmds/profile.py
```diff
@@ -192,7 +193,13 @@ def _prompt_for_allow_password_set(profile_name):
 def _set_pw(profile_name, password):
     c42profile = cliprofile.get_profile(profile_name)
     try:
-        validate_connection(c42profile.authority_url, c42profile.username, password)
+        validate_connection(
+            c42profile.authority_url, c42profile.username, password, None
+        )
+    except Py42MFARequiredError:
+        echo(
```
unparalleled-js, line 200: Could we prompt at this point so that we can still validate?

src/code42cli/cmds/profile.py
```diff
@@ -192,7 +193,13 @@ def _prompt_for_allow_password_set(profile_name):
 def _set_pw(profile_name, password):
     c42profile = cliprofile.get_profile(profile_name)
     try:
-        validate_connection(c42profile.authority_url, c42profile.username, password)
+        validate_connection(
+            c42profile.authority_url, c42profile.username, password, None
```
unparalleled-js, line 197: Rather than having to explicitly pass in `None` here, we could just make `validate_connection` default to `None` for `totp` arg,

src/code42cli/sdk_client.py
```diff
@@ -30,18 +32,24 @@ def create_sdk(profile, is_debug_mode):
...
         logger.log_error(str(err))
-        raise LoggedCLIError("Problem connecting to {}".format(authority_url))
+        raise LoggedCLIError(f"Problem connecting to {authority_url}.")
+    except Py42MFARequiredError:
+        totp = prompt("Multi-factor authentication required. Enter TOTP", type=int)
+        return validate_connection(authority_url, username, password, totp)
     except Py42UnauthorizedError as err:
         logger.log_error(str(err))
-        raise Code42CLIError("Invalid credentials for user {}".format(username))
+        if "INVALID_TIME_BASED_ONE_TIME_PASSWORD" in err.response.text:
```
unparalleled-js, line 49: This is fine for now but this seems like a common enough scenario that we should have a custom py42 error for this

docs/userguides/gettingstarted.md
```diff
@@ -72,10 +72,16 @@ python3 -m pip install code42cli --upgrade
...
+Password (TOTP) must be provided at every invocation of the CLI, either via the `--totp` option, or when prompted.
```
unparalleled-js, line 83: comma not needed before `or`

tests/test_sdk_client.py
```diff
@@ -114,5 +120,46 @@ def mock_get_password():
...
+    assert str(err.value) == "Invalid TOTP token for user Test."
+
+
+def test_totp_option_when_passed_is_passed_to_sdk_initialization(
+    mocker, profile, runner
+):
+    mock_py42 = mocker.patch("code42cli.sdk_client.py42.sdk.from_local_account")
+    cli_state = CLIState()
+    totp = "1234"
+    profile.authority_url = "test.com"
```
unparalleled-js, line 158: `example.com` is generally better.

I once accidentally made a network request to something like `test.com`

src/code42cli/cmds/profile.py
timabrmsn: What do we need to validate? If we get the Py42MFARequired error, it means the password was correct but the token was missing.

docs/userguides/gettingstarted.md
timabrmsn: Ohh, you mean the v3 token?

unparalleled-js: Yes I meant the v3 token

unparalleled-js (approved):

alanag13 (approved):

PR #282, opened 2021-05-12
pin click version to fix failing tests

alanag13:
Click 8.0 was released ~14hrs ago and is causing unit tests to fail due to some breaking changes it made. We should eventually try to update to the latest and greatest and make the appropriate corresponding changes but for now let's just get everything to pass again.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,14 +31,14 @@ setup(
     zip_safe=False,
     python_requires=">3, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, !=3.5.*, <4",
     install_requires=[
-        "click>=7.1.1",
+        "click>=7.1.1, <8",
         "click_plugins>=1.1.1",
         "colorama>=0.4.3",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -31,14 +31,14 @@
...
-        "click>=7.1.1",
+        "click>=7.1.1, <8",
         "click_plugins>=1.1.1",
         "colorama>=0.4.3",
         "c42eventextractor==0.4.1",
         "keyring==18.0.1",
         "keyrings.alt==3.2.0",
         "pandas>=1.1.3",
-        "py42>=1.11.1",
+        "py42>=1.13.0",
```
alanag13, line 41: I also found that the current min py42 version is wrong, 1.13.0 added some methods that the latest CLI release uses

peterbriggs42 (approved):

setup.py
alanag13: woops nvm this was caused by my branch not being caught up

PR #283, opened 2021-05-14
Users command: Functionality to add/remove roles + Users user guide

maddie-vargo:
This PR addresses a portion of Issue #259 by adding functionality to add and remove user roles to an individual user.  This command takes a username and a user role name as two positional arguments. 

**Testing Procedure:** 

```
cod42 users add-role <username> <role name>
code42 users remove-role <username> <role name>
```

**PR Checklist**

 


src/code42cli/cmds/users.py
```diff
@@ -60,6 +61,50 @@ def list_users(state, org_uid, role_name, active, inactive, format):
         formatter.echo_formatted_dataframe(df)
 
 
+@users.command()
+@click.argument("username")
```
unparalleled-js, line 65: When commands have multiple arguments, we elect to use required options instead.
This is because then the order does not matter, like I don't have to use `--help` to see if the username goes first or the role name.

src/code42cli/cmds/users.py
```diff
@@ -60,6 +61,50 @@ def list_users(state, org_uid, role_name, active, inactive, format):
...
+@click.argument("role-name")
+@sdk_options()
+def add_role(state, username, role_name):
+    """Add the specified role to the specified username"""
```
unparalleled-js, line 69: Maybe just `... specified user` at the end of the help text (rather than `username`)?  

Or `the user with the specified username`? 

Just because we are not adding a role to a username, but to the user themselves

src/code42cli/cmds/users.py
```diff
@@ -60,6 +61,50 @@ def list_users(state, org_uid, role_name, active, inactive, format):
...
+@click.argument("role-name")
+@sdk_options()
+def remove_role(state, username, role_name):
+    """Add the specified role to the specified username"""
+    _remove_user_role(state.sdk, role_name, username)
+
+
+def _add_user_role(sdk, role_name, username):
+    user_id = _get_user_id(sdk, username)
+    try:
```
unparalleled-js, line 84: I don't understand what this `try` catch is doing.
It looks like it catches an error just to raise it again.
Is there any technical difference of just letting the error raise naturally?  If so, we should add a comment explaining what it is because it is not obvious.

unparalleled-js: Also, if it is needed, it could be a helper method and `_remove_user_role()` can share it

src/code42cli/cmds/users.py
```diff
@@ -60,6 +61,50 @@ def list_users(state, org_uid, role_name, active, inactive, format):
...
+        _get_role_id(sdk, role_name)
+    except Code42CLIError:
+        raise
+    sdk.users.remove_role(user_id, role_name)
+
+
+def _get_user_id(sdk, username):
+    user = sdk.users.get_by_username(username)["users"]
+    if not user:
+        raise UserDoesNotExistError
```
unparalleled-js, line 103: We have a share method that handles this nicely, located in `code42cli.cmds.shared.get_user_id()`... It will even cache IDs for users! We should use that instead.

Also, when raising errors, you need to use an instance of the error, like `UserDoesNotExistError(username)` (see `code42cli.cmds.shared.get_user_id()`)

tests/cmds/test_users.py
```diff
@@ -153,3 +161,44 @@ def test_list_users_when_given_excluding_active_and_inactive_uses_active_equals_
...
+    )
+
+
+def test_remove_user_role_raises_error_when_role_does_not_exist(
+    runner, cli_state, get_user_id_success, get_available_roles_success
+):
+    command = ["users", "remove-role", "test.username@example.com", "test"]
+    result = runner.invoke(cli, command, obj=cli_state)
+    assert result.exit_code == 1
+    assert "Role with name test not found." in result.output
```
unparalleled-js, line 204: It might be clearer if we put quotes around the role name in the error so that the error message is 

`Role with name 'test' not found`

src/code42cli/cmds/users.py
unparalleled-js: Ah I just noticed that this uses the legacy ID so that won't work.

However, we still need to raise an instance of the error here

src/code42cli/cmds/users.py
maddie-vargo: The try/except block isn't doing anything, actually - I will remove it.

We don't actually even need the roleID from _get_role_id to proceed. However, I like the error handling Cecilia had built for an invalid username. It handles cases where an invalid username was entered and provides the following error message: 

```
Error: Role with name Fake_Role not found.
```



Otherwise, the error is: 


```
Error: Problem making request to server.
```


Py42 doesn't really have a method for determining if a role name is valid.

src/code42cli/cmds/users.py
maddie-vargo: Great, I added them as functions with flexible help text, assuming that they would be re-used in further expansions of the `users` command.

src/code42cli/cmds/users.py
```diff
@@ -60,6 +66,44 @@ def list_users(state, org_uid, role_name, active, inactive, format):
         formatter.echo_formatted_dataframe(df)
 
 
+@users.command()
+@username_option("Username of the target user")
```
unparalleled-js, line 70: These help texts should end in period

unparalleled-js: Same with ones above `remove_role()`

src/code42cli/cmds/users.py
```diff
@@ -60,6 +66,44 @@ def list_users(state, org_uid, role_name, active, inactive, format):
...
+@role_name_option("Name of role to add")
+@sdk_options()
+def add_role(state, username, role_name):
+    """Add the specified role to the user with the specified username"""
```
unparalleled-js, line 74: The command help texts should also end in period, the part in the `""" """`

unparalleled-js (approved):
Some missing periods in help text but otherwise looks good.

src/code42cli/cmds/users.py
unparalleled-js: Might be worth adding a comment about that, or even extracting the error handling into a separate method
(still approved but thought I'd mention that)

PR #284, opened 2021-05-17
Added note to deactivate devices user guide about Excel behavior

ryanvanasse:
If a user just opens the CSV with Excel, it sets the guid column to a data type that does not accept numbers longer than 15 digits. This wipes out the last few digits of the guid and replaces them with zeros. For users who are not working with our CSVs all the time, this behavior is confusing, and unfortunately when it comes to feeding this CSV back into the tool to do a bulk deactivate, will cause failure (at least based on all of the other documented behavior).

I've added a note here to the user guide so that users aren't surprised by this.

In other news, I think I hate restructured text syntax.

### DIFF
--- a/docs/userguides/deactivatedevices.md
+++ b/docs/userguides/deactivatedevices.md
@@ -38,6 +38,9 @@ curating a list of devices to deactivate:
 * `--exclude-most-recently-connected INTEGER` - allows you to exclude the most recently connected device (per user) from the results. This allows you to ensure that every user is left with at least N device(s), regardless of how recently they have connected.
 * `--created-before DATE|TIMESTAMP|SHORT_TIME` - allows you to only see devices created before a particular date.
 
+```eval_rst
+.. note:: If you use Excel to filter or edit the list of devices, you will find that the ``guid`` column is displayed in scientific notation—when you change the column type to text, you'll see that the full number now has had its last digits replaced by zeros. This modification makes the guid invalid for deactivating devices. In order to edit the list in Excel without overwriting this, you will need to `import the data into a new workbook <https://support.microsoft.com/en-us/office/keeping-leading-zeros-and-large-numbers-1bf7b935-36e1-4985-842f-5dfa51f85fe7>`_. rather than just opening the CSV. If you want to work on this in Excel later, you will also want to save as an xlsx file and only 'save as' csv when you're ready to make the final export for running the bulk deactivate command.
+```
 ## Deactivate devices
 
 Once you have a list of devices that you want to remove, you can


docs/userguides/deactivatedevices.md
```diff
@@ -38,6 +38,9 @@ curating a list of devices to deactivate:
 * `--exclude-most-recently-connected INTEGER` - allows you to exclude the most recently connected device (per user) from the results. This allows you to ensure that every user is left with at least N device(s), regardless of how recently they have connected.
 * `--created-before DATE|TIMESTAMP|SHORT_TIME` - allows you to only see devices created before a particular date.
 
+```eval_rst
+.. note:: If you use Excel to filter or edit the list of devices, you will find that the ``guid`` column is displayed in scientific notation—when you change the column type to text, you'll see that the full number now has had its last digits replaced by zeros. This modification makes the guid invalid for deactivating devices. In order to edit the list in Excel without overwriting this, you will need to `import the data into a new workbook <https://support.microsoft.com/en-us/office/keeping-leading-zeros-and-large-numbers-1bf7b935-36e1-4985-842f-5dfa51f85fe7>`_. rather than just opening the CSV. If you want to work on this in Excel later, you will also want to save as an xlsx file and only 'save as' csv when you're ready to make the final export for running the bulk deactivate command.
```
unparalleled-js, line 42: Is the data modified by merely opening the file in Excel or does changing the column type cause the truncation?

unparalleled-js: I just tried this with the `user list` command (And I really am not an Excel user), and what happened for me is that I noticed the `userUid` column was sci-not and when I changed it to text, nothing happened.  It remained in sci-not but the type changed.  When I `cat` when .csv file, the original IDs are still there. Does this mean anything to you?

docs/userguides/deactivatedevices.md
```diff
@@ -38,6 +38,9 @@ curating a list of devices to deactivate:
 * `--exclude-most-recently-connected INTEGER` - allows you to exclude the most recently connected device (per user) from the results. This allows you to ensure that every user is left with at least N device(s), regardless of how recently they have connected.
 * `--created-before DATE|TIMESTAMP|SHORT_TIME` - allows you to only see devices created before a particular date.
 
+```eval_rst
```
unparalleled-js, line 41: I noticed this also happens with `userUid`s... I am not sure if this is the best place for this note.

unparalleled-js: Perhaps it'd be a good idea to have a whole guide dedicated to working with CSVs.  It can contain examples for outputting CSV files and using CSV files in bulk commands. It can mention potential issues and workarounds, such as dealing with scientific notation false auto-detection.

alanag13: could this be considered a bug? would we see this behavior if all these types of ids got exported to the csvs wrapped in quotes?

alanag13: @unparalleled-js to more directly answer your question, I think that if there are issues related to csvs in general rather than specific cli commands that it's worth having their own section. Otherwise we'll find ourself repeating this info for user uids, device uids, etc, etc. If the problem is the format and not the command, let's have the guidance be reflective of that.

It still may be helpful to have a note like:
```
This command exports data to a csv file. To troubleshoot common issues when exporting to csv, see `here`
```

Near any potentially impacted commands.

ceciliastevens: I just tested with numbers wrapped in quotes and Excel still rounded them, so I don't think that would make a difference.

docs/userguides/deactivatedevices.md
timabrmsn: did you save the file anytime after opening it?

docs/userguides/deactivatedevices.md
timabrmsn: I wonder if it'd be possible to detect that the IDs might have been modified. Like if when we hit a "not found" error for any "long" ID type, we could check to see if the last digits are all `0`s and log a message indicating Excel as possibly the culprit?

docs/userguides/deactivatedevices.md
ryanvanasse: The data in the underlying csv isn't changed except when saving, but at the same time, if you open the CSV, Excel converts the guids to scientific notation, and if you try to convert that back to text, it has already thrown away the "non-significant" numbers at the end of the guid, so you won't be able to save the CSV with the correct guids. 

In other words, opening the CSV file directly in Excel, rather than importing, already sets the column type to one that throws away numbers in the guid. The csv itself is only modified after you save in Excel, though. 

This is not an uncommon Excel issue even outside of Code42 tooling—searching for "large numbers csv" is mostly users encountering this with Excel. This is a [pretty good writeup of the issue](https://support.3dcart.com/knowledgebase/article/View/619/7/why-does-my-exported-csv-data-get-converted-to-weird-formats) from an e-commerce software provider.

docs/userguides/deactivatedevices.md
ryanvanasse: > could this be considered a bug? would we see this behavior if all these types of ids got exported to the csvs wrapped in quotes or something? I'd be curious what the difference is between this file as-is and when it is opened into a new playbook.

I doubt this would be considered a bug. I don't think that *we* are doing anything unusual, it's more of just the result of how Excel chooses to interpret those CSVs. Searching for "long number csv excel" shows a lot of other applications having similar issues with Excel, and the standard fix/workaround has basically always been "do an import, don't open directly." While Excel is almost certainly the most widely used software for our customers to view/manipulate those CSVs, it's not the only tool (I use the [Visidata](https://www.visidata.org/) utility mostly, myself) and doing things "the Excel way" may result in other tools being confused or having to remove excess characters, like quotes or backticks.

I don't actually *have* an Excel license for testing, but it looks like one may be able to coerce Excel into not changing the format on open by [various kludgy tricks](https://superuser.com/questions/861964/how-do-i-stop-excel-from-converting-value-0503e000-to-5-03e02-automatically/862073) (e.g. `="136479852364713598"` or ``` `13518797813518```)


> It still may be helpful to have a note like:
> 
> ```
> This command exports data to a csv file. To troubleshoot common issues when exporting to csv, see `here`
> ```
> 
> Near any potentially impacted commands.

Definitely. A "Working with CSVs" seems like a useful thing to have, but we'd definitely want to have a link to that from any articles directly. I don't think that people would likely think of the CSV being the culprit when something went wrong, and think to look for that.

ryanvanasse: I didn't realize this but we have a [Import CSV Reports to Microsoft Excel](https://support.code42.com/Administrator/Cloud/Monitoring_and_managing/Import_CSV_reports_to_Microsoft_Excel) guide in the official Code42 documentation...perhaps this should be rewritten to point to that guide instead.

PR #288, opened 2021-05-26
Have profile commands use create_sdk so disabling ssl verification works

timabrmsn:
This allows profile creation, update, and reset-pw commands to work when a server's SSL cert is invalid/self-signed and the user has set the --disable-ssl-errors flag on their profile.

### DIFF
--- a/src/code42cli/sdk_client.py
+++ b/src/code42cli/sdk_client.py
@@ -33,29 +33,29 @@ def create_sdk(profile, is_debug_mode, password=None, totp=None):
         )
         py42.settings.verify_ssl_certs = False
     password = password or profile.get_password()
-    return validate_connection(profile.authority_url, profile.username, password, totp)
+    return _validate_connection(profile.authority_url, profile.username, password, totp)
 
 
-def validate_connection(authority_url, username, password, totp=None):
+def _validate_connection(authority_url, username, password, totp=None):
     try:
         return py42.sdk.from_local_account(authority_url, username, password, totp=totp)
     except SSLError as err:
-        logger.log_error(str(err))
+        logger.log_error(err)
         raise LoggedCLIError(
             f"Problem connecting to {authority_url}, SSL certificate verification failed.\nUpdate profile with --disable-ssl-errors to bypass certificate checks (not recommended!)."
         )
     except ConnectionError as err:
-        logger.log_error(str(err))
+        logger.log_error(err)
         raise LoggedCLIError(f"Problem connecting to {authority_url}.")
     except Py42MFARequiredError:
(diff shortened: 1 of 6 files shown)


timabrmsn:
I see if have some tests to update as well..

src/code42cli/sdk_client.py
```diff
@@ -31,13 +32,18 @@ def create_sdk(profile, is_debug_mode, totp=None):
...
-    password = profile.get_password()
+    password = password or profile.get_password()
     return validate_connection(profile.authority_url, profile.username, password, totp)
 
 
 def validate_connection(authority_url, username, password, totp=None):
     try:
         return py42.sdk.from_local_account(authority_url, username, password, totp=totp)
+    except SSLError as err:
+        logger.log_error(str(err))
```
unparalleled-js, line 43: `log_error()` does already call `str()` on the err
https://github.com/code42/code42cli/blob/master/src/code42cli/logger/__init__.py#L116

unparalleled-js:
I think we could make `validate_connection` protected (`_validate_connection()` and migrate any specific tests of it to just `create_sdk()` so that we know just to always use `create_sdk()`

unparalleled-js:
Missing CHANGELOG entry?

tests/cmds/test_profile.py
```diff
@@ -37,12 +38,13 @@ def mock_getpass(mocker):
 
 @pytest.fixture
 def mock_verify(mocker):
-    return mocker.patch("{}.cmds.profile.validate_connection".format(PRODUCT_NAME))
+    return mocker.patch("{}.cmds.profile.create_sdk".format(PRODUCT_NAME))
```
unparalleled-js, line 41: Could use `fstr` here or just type it out

tests/test_sdk_client.py
```diff
@@ -54,99 +64,84 @@ def test_create_sdk_when_profile_has_ssl_errors_disabled_sets_py42_setting_and_p
...
+def test_create_sdk_uses_given_credentials(
+    mock_sdk_factory, mock_profile_with_password
+):
+    create_sdk(mock_profile_with_password, False)
+    mock_sdk_factory.assert_called_once_with(
+        "example.com", "foo", "Test Password", totp=None
+    )
 
 
 def test_validate_connection_when_mfa_required_exception_raised_prompts_for_totp(
```
unparalleled-js, line 122: Forgot to rename some of these `test_create_sdk_` rather than `test_validate_connection_`

unparalleled-js (approved):
Looks good, just some outdated test method names

PR #290, opened 2021-05-27
Users command: option to report on legal hold membership

maddie-vargo:
This PR addresses the final component of Issue #176 by adding functionality to the `users` command to report on legal hold membership. This option will report on the legal hold matter name and ID for any user on legal hold. This work largely mimics the legal hold reporting option in the `devices` command

**Testing Procedure**

```
code42 users list --include-legal-hold-membership
```

**PR Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 - New command `code42 users remove-role` to remove a user role from a single user.
 
+- New command 'code42 users list --include-legal-hold-membership' to print the legal hold matter name and ID for any user on legal hold
+
 ## 1.6.1 - 2021-05-27
 
 ### Fixed
--- a/src/code42cli/cmds/users.py
+++ b/src/code42cli/cmds/users.py
@@ -1,5 +1,6 @@
 import click
 from pandas import DataFrame
+from pandas import json_normalize
 
 from code42cli.click_ext.groups import OrderedGroup
 from code42cli.click_ext.options import incompatible_with
@@ -46,9 +47,19 @@ def username_option(help):
 @role_name_option("Limit results to only users having the specified role.")
 @active_option
 @inactive_option
+@click.option(
+    "--include-legal-hold-membership",
+    required=False,
+    type=bool,
+    default=False,
+    is_flag=True,
+    help="Include legal hold membership in output.",
+)
 @format_option
 @sdk_options()
-def list_users(state, org_uid, role_name, active, inactive, format):
+def list_users(
+    state, org_uid, role_name, active, inactive, include_legal_hold_membership, format
+):
     """List users in your Code42 environment."""
     if inactive:
         active = False
@@ -59,6 +70,8 @@ def list_users(state, org_uid, role_name, active, inactive, format):
         else None
     )
     df = _get_users_dataframe(state.sdk, columns, org_uid, role_id, active)
+    if include_legal_hold_membership:
+        df = _add_legal_hold_membership_to_user_dataframe(state.sdk, df)
     if df.empty:
         click.echo("No results found.")
     else:
@@ -122,3 +135,37 @@ def _get_users_dataframe(sdk, columns, org_uid, role_id, active):
         users_list.extend(page["users"])
 
     return DataFrame.from_records(users_list, columns=columns)
+
+
+def _add_legal_hold_membership_to_user_dataframe(sdk, df):
+    columns = ["legalHold.legalHoldUid", "legalHold.name", "user.userUid"]
+
+    legal_hold_member_dataframe = (
+        json_normalize(list(_get_all_active_hold_memberships(sdk)))[columns]
+        .groupby(["user.userUid"])
+        .agg(",".join)
+        .rename(
+            {
+                "legalHold.legalHoldUid": "legalHoldUid",
+                "legalHold.name": "legalHoldName",
+            },
+            axis=1,
+        )
+    )
+    df = df.merge(
+        legal_hold_member_dataframe,
+        how="left",
+        left_on="userUid",
+        right_on="user.userUid",
+    )
+
+    return df
+
+
+def _get_all_active_hold_memberships(sdk):
+    for page in sdk.legalhold.get_all_matters(active=True):
+        for matter in page["legalHolds"]:
+            for _page in sdk.legalhold.get_all_matter_custodians(
+                legal_hold_uid=matter["legalHoldUid"], active=True
+            ):
+                yield from _page["legalHoldMemberships"]
--- a/tests/cmds/test_users.py
+++ b/tests/cmds/test_users.py
@@ -1,9 +1,11 @@
 import json
 
 import pytest
+from pandas import DataFrame
 from py42.response import Py42Response
 from requests import Response
 
+from code42cli.cmds.users import _add_legal_hold_membership_to_user_dataframe
 from code42cli.main import cli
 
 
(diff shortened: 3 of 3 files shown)


src/code42cli/cmds/users.py
```diff
@@ -122,3 +135,37 @@ def _get_users_dataframe(sdk, columns, org_uid, role_id, active):
...
+    )
+
+    return df
+
+
+def _get_all_active_hold_memberships(sdk):
+    for page in sdk.legalhold.get_all_matters(active=True):
+        for matter in page["legalHolds"]:
+            for _page in sdk.legalhold.get_all_matter_custodians(
+                legal_hold_uid=matter["legalHoldUid"], active=True
```
unparalleled-js, line 169: `userUid` is also a param on this method... And we already have the list of users.. Would it more efficient to instead   loop through the users and get their memberships that way?

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 - New command `code42 users remove-role` to remove a user role from a single user.
 
+- New command 'code42 users list --include-legal-hold-membership' to print the legal hold matter name and ID for any user on legal hold
```
unparalleled-js, line 19: I know this has probably already been said, but what is the use-case for this functionality? I had heard most customers have a single matter for legal hold.

Is this flag general enough or this just for a specific use-case? I am also wondering if this is the right place to include this flag... What if I wanted to filter users by matter ID instead? Or what if I want to know which detection list or alert rules each user belong to as well?

maddie-vargo: The primary use case that this addresses is "show me all the users on legal hold." 

There are customers that have a lot of legal hold matters. Therefore, the `legal-hold show <matter ID>` command would not meet their needs if they need to see users on all legal holds.

Also, `legal-hold show <matter ID>` produces the users in column format. This makes sense for a `show` command, which is meant to be a high-level overview (This was discussed in the [PR](https://github.com/code42/code42cli/pull/192#discussion_r562810149) that added the `devices list --include-legal-hold-membership`).  However, for an admin to work with the data and put it in different formats, the `list` command provided some more flexibility (at least for device info)

src/code42cli/cmds/users.py
maddie-vargo: I could see it both ways. As written, it pulls the full custodian list of a matter custodians and merges that table into the users table.

If, say, >10% of the users in the list output are in fact on legal hold, I can see it being more straightforward to write a method using the user_uid parameter.

However, if 0.5% of users in the list output are on legal hold, that seems like more processing than necessary.

CHANGELOG.md
timabrmsn: I'm good with this being in the CLI proper. As it could be useful to any customer using legal hold. 

I think I'd also be ok with future flags for including if a user is on a detection list as well. Having the CLI `users` command be a kind of one-stop reporting tool to get all users info makes sense. Especially for larger customers who might have multiple teams (or automations) managing who gets put on what list/hold.

maddie-vargo:

src/code42cli/cmds/users.py
```diff
@@ -46,9 +47,19 @@ def username_option(help):
 @role_name_option("Limit results to only users having the specified role.")
 @active_option
 @inactive_option
+@click.option(
+    "--include-legal-hold-membership",
+    required=False,
```
unparalleled-js, line 52: `required=False` not needed, that is the default

src/code42cli/cmds/users.py
```diff
@@ -46,9 +47,19 @@ def username_option(help):
...
+    type=bool,
```
unparalleled-js, line 53: `type=bool` not needed when `is_flag=True`

src/code42cli/cmds/users.py
```diff
@@ -122,3 +135,37 @@ def _get_users_dataframe(sdk, columns, org_uid, role_id, active):
...
+    return df
+
+
+def _get_all_active_hold_memberships(sdk):
+    for page in sdk.legalhold.get_all_matters(active=True):
+        for matter in page["legalHolds"]:
+            for _page in sdk.legalhold.get_all_matter_custodians(
+                legal_hold_uid=matter["legalHoldUid"], active=True
+            ):
+                yield from _page["legalHoldMemberships"]
```
unparalleled-js, line 171: TIL `yield from`

src/code42cli/cmds/users.py
```diff
@@ -122,3 +135,37 @@ def _get_users_dataframe(sdk, columns, org_uid, role_id, active):
...
+def _add_legal_hold_membership_to_user_dataframe(sdk, df):
+    columns = ["legalHold.legalHoldUid", "legalHold.name", "user.userUid"]
+
+    legal_hold_member_dataframe = (
+        json_normalize(list(_get_all_active_hold_memberships(sdk)))[columns]
+        .groupby(["user.userUid"])
+        .agg(",".join)
+        .rename(
+            {
+                "legalHold.legalHoldUid": "legalHoldUid",
```
unparalleled-js, line 149: In the command for `code42 legal-hold list`, for table format, we should this as `Matter ID`... Will that also happen here?

CHANGELOG.md
```diff
@@ -8,6 +8,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 The intended audience of this file is for py42 consumers -- as such, changes that don't affect
 how a consumer would use the library (e.g. adding unit tests, updating documentation, etc) are not captured here.
 
+## Unreleased
+
+### Added
+
+- New command 'code42 users list --include-legal-hold-membership' to print the legal hold matter name and ID for any user on legal hold
```
unparalleled-js, line 15: Missing period at end

tests/cmds/test_users.py
```diff
@@ -169,6 +226,32 @@ def test_list_users_when_given_excluding_active_and_inactive_uses_active_equals_
     )
 
 
+def test_add_legal_hold_membership_to_user_dataframe_adds_legal_hold_columns_to_dataframe(
```
unparalleled-js, line 229: This test is not really needed, `_add_legal_hold_membership_to_user_dataframe` is protected and an implementation detail.

We only really care that when we do `--include-legal-hold-membership` that the CLI outputs that.

src/code42cli/cmds/users.py
unparalleled-js: It would be nice to be able to pass in user friendly keys for Table format headers for when using `DataFrameOutputFormatter`, as that was the norm before.

Table format should use label-esque keys while all the other ones should use raw API values.

tests/cmds/test_users.py
```diff
@@ -169,6 +226,32 @@ def test_list_users_when_given_excluding_active_and_inactive_uses_active_equals_
...
+    get_all_matter_success,
+):
+    result = runner.invoke(
+        cli, ["users", "list", "--include-legal-hold-membership"], obj=cli_state
+    )
+
+    assert "Legal Hold #1,Legal Hold #2" in result.output
+    assert "123456789,987654321" in result.output
+
+
```
unparalleled-js, line 254: Can we get a test for the following two cases:

1.) A user is not on any legal hold but the `--include-legal-hold-membership` flag was passed in (keys should be there but their values should be `None`)
2.) A user is on a legal hold but the flag `--include-legal-hold-membership` was not passed in (keys and values should not be present in the output)

unparalleled-js:
The build failures will go away if you catch your branch up with upstream `main`
Also, can you target `main` as the target branch? `master` is no longer the `main` branch.

unparalleled-js:
when using Table format, the spacing between the end of the Users data in the beginning of the Legal Hold data is really really big... Even when the user is part of many matters, it is still almost twice as big as it needs to be. 
Screen Shot 2021-07-08 at 12 54 41


What is going on with this?

unparalleled-js:
I am little curious what the legal hold data is a comma-separated string rather than a list type.

For example, when I create a CSV, the license entry is `['admin.securityTools']` while the `legalHoldUid` entry is `"1002360211054033697,996367185330237359,995898853137248175,995896307393194971,995894227423915951"` 

Same with JSON.

Would using lists be easier for programmatically handling the output for CSV and JSON outputs?

maddie-vargo:
**Re: Build Failures**
Resolved the build failures by syncing the fork, and set `main` as the target branch. Thanks!


**Re: Column spacing for a long list of legal holds:**
 I find that this column will be as long as the longest list. In the output referenced by your screenshot, is there someone listed that has many more legal holds listed? 
I'm not able to reproduce this. I've added a user to 5 legal holds and ran a `users list --include-legal-hold-membership` command. The most legal holds a user has is 5, and the spacing is appropriate. 

**Re: Legal Hold Info as a comma-separated string**
The user `license` info is coming directly from the api, where the value is a list type. The Legal Hold info is being formatted by the code. Currently, I'm only aggregating it into a comma-separated string, but I can certainly change it to a standard list.

`"955060991270602172,946193444244471969" ` >> `[955060991270602172, 946193444244471969]`

However, I'm not sure how to programmatically have that display as a list for some formats, and as a string for the table. If I had to choose, I err on the side of making it a list for all formats, as opposed to the comma-separated string.

tests/cmds/test_users.py
maddie-vargo: I added these tests. The code didn't handle cases where 0 custodians were returned. Thanks for suggesting this test case.

I also added a test for cases where the flag`--include-legal-hold-membership` and no matters were returned.

tests/cmds/test_users.py
maddie-vargo: Removed.

src/code42cli/cmds/users.py
maddie-vargo: The strange format of `"legalHold.legalHoldUid"` does not (should not) come through the final output. This syntax gets introduced with the `json_normalize` command, which flattens the json structure into nested columns names.

tests/cmds/test_users.py
```diff
@@ -72,6 +108,22 @@ def get_all_users_generator():
     yield TEST_USERS_RESPONSE
 
 
+def matter_list_generator():
+    yield TEST_MATTER_RESPONSE
+
+
+def custodian_list_generator():
+    yield TEST_CUSTODIANS_RESPONSE
```
unparalleled-js, line 116: These should use the `create_mock_response()` method:

```
yield create_mock_response(mocker, data=TEST_CUSTODIANS_RESPONSE)
```

Anything that is a response from a py42 method should use this method, so all these ones here,

it is being uses in some of the other fixtures below as an example!

Please and thank you! This is looking really good.

PR #291, opened 2021-06-02
Bugfix: allow disable ssl on profile cmds

timabrmsn:
This PR makes `validate_connection()` internal and has the profile commands use `create_sdk()` instead so they can succeed when ssl verification is turned off and they hit a server with a bad cert. 

It also adds the `--enable-ssl-errors` option to `profile update` to allow a user to re-enable verification on a profile that had it disabled (the only way prior to this was to either delete/recreate profile or manually edit the config.cfg


timabrmsn:
Wow, apparently we already merged/released this fix..

PR #292, opened 2021-06-02
Fstrs everywhere

unparalleled-js:
Migrating code to be more py3-y by using f-strs everywhere where it makes sense. Note, it does not always make sense, such as in the following conditions:

* You are interpolating an expression involving slashes `/`
* You are building a dict of values to pass into format
* You are using a str-template pattern

F-str makes sense in most cases however.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,7 +3,7 @@ repos:
       rev: v2.7.1
       hooks:
         - id: pyupgrade
-          args: ["--py3-plus"]
+          args: ["--py36-plus"]
     - repo: https://github.com/asottile/reorder_python_imports
       rev: v2.3.0
       hooks:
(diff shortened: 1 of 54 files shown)


alanag13:
@unparalleled-js I think you could get this to be something that `tox style` would do for you automatically if we modified `.pre-commit-config.yaml` to pass different args to `pyupgrade`, namely:

```
 args: ["--py36-plus"]
```

https://github.com/asottile/pyupgrade#f-strings

unparalleled-js:
> @unparalleled-js I think you could get this to be something that `tox style` would do for you automatically if we modified `.pre-commit-config.yaml` to pass different args to `pyupgrade`, namely:
> 
> ```
>  args: ["--py36-plus"]
> ```
> 
> https://github.com/asottile/pyupgrade#f-strings

It works

alanag13 (approved):

PR #293, opened 2021-06-10
auto console w sdk

unparalleled-js:
Adds a new command that automatically opens a shell with a pre-initialized py42 SDK.

```
code42 console
```

```python
 dP""b8  dP"Yb  8888b. 888888  dP88  oP"Yb.
dP   `" dP   Yb 8I  Yb 88__   dP 88  "' dP'
Yb      Yb   dP 8I  dY 88""  d888888   dP'
 YboodP  YbodP  8888Y" 888888    88  .d8888

code42cli version 1.6.1, by Code42 Software.
powered by py42 version 1.14.2.
In [1]: sdk.users.get_current()
Out[1]: <Py42Response [status=200, data={'active': True, 'blocked': False, 'creationDate': '2019-09-30T21:03:08.587Z', 'email': 'juliya.smith...rs@code42.com', ...}]>
```

This is useful for running more complex commands using py42, observing responses from py42, and would save time for QA individual testing py42 - they can instead call

with a code42cli profile already setup and in-use... Then, they can start using the `sdk`.

### DIFF
--- a/src/code42cli/__init__.py
+++ b/src/code42cli/__init__.py
@@ -1,2 +1,15 @@
+from py42.__version__ import __version__ as py42version
+
+from code42cli.__version__ import __version__ as cliversion
+
+
 PRODUCT_NAME = "code42cli"
 MAIN_COMMAND = "code42"
+BANNER = f"""\b
+ dP""b8  dP"Yb  8888b. 888888  dP88  oP"Yb.
+dP   `" dP   Yb 8I  Yb 88__   dP 88  "' dP'
+Yb      Yb   dP 8I  dY 88""  d888888   dP'
+ YboodP  YbodP  8888Y" 888888    88  .d8888
+
+code42cli version {cliversion}, by Code42 Software.
+powered by py42 version {py42version}."""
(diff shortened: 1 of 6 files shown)


unparalleled-js:
I also think it'd be cool to have commands:

`code42 alerts console` and
`code42 security-data console`

that have all the filters / query objects auto imported so you can just start querying

src/code42cli/__init__.py
```diff
@@ -1,2 +1,15 @@
+from py42.__version__ import __version__ as py42version
+
+from code42cli.__version__ import __version__ as cliversion
+
+
 PRODUCT_NAME = "code42cli"
 MAIN_COMMAND = "code42"
+BANNER = f"""\b
```
peterbriggs42, line 8: Love the banner! Great idea

peterbriggs42 (approved):

timabrmsn:
I love this and think it'll be really useful! 

The only thing I think we might reconsider is the command name. Since "Code42 console" is the generally used name of the web experience, and the console even has a "CLI", I think it would be better to rename it to something without meaning overlap. 

Maybe `code42 shell`?

src/code42cli/main.py
```diff
@@ -73,12 +64,13 @@ def cli(state, python):
 
 cli.add_command(alerts)
```
unparalleled-js, line 65: I alphabetized these :/

timabrmsn (approved):

PR #294, opened 2021-06-17
release prep

unparalleled-js:
Also pins py42 to 1.15.0 to take advantage of the bugfix for creating users


peterbriggs42 (approved):

setup.py
```diff
@@ -39,7 +39,7 @@
         "keyrings.alt==3.2.0",
         "ipython>=7.16.1",
         "pandas>=1.1.3",
-        "py42>=1.14.2",
+        "py42>=1.15.0",
```
unparalleled-js, line 42: Hmm this actually might not be necessary... but does it hurt?

PR #299, opened 2021-06-23
release prep

unparalleled-js:
Release?


peterbriggs42 (commented):
Don't we need to bump the Python version to take advantage of the auth change?

unparalleled-js:
> Don't we need to bump the Python version to take advantage of the auth change?

That was done on a previous PR

PR #300, opened 2021-06-24
Update user

ceciliastevens:
**Description of Change**

This change adds a method `code42 users update` to update a single user, and the corresponding bulk method `code42 users bulk update`

It partially addresses (but does not fully address) issue #259

**Testing Procedure**

```
code42 users update --user-id <user-id> --email <new email>
code42 users bulk update users_to_update.csv
```

**PR Checklist**

### DIFF
--- a/src/code42cli/cmds/users.py
+++ b/src/code42cli/cmds/users.py
@@ -1,15 +1,15 @@
 import click
-from code42cli.bulk import run_bulk_process
 from pandas import DataFrame
 
 from code42cli.bulk import generate_template_cmd_factory
+from code42cli.bulk import run_bulk_process
 from code42cli.click_ext.groups import OrderedGroup
 from code42cli.click_ext.options import incompatible_with
 from code42cli.errors import Code42CLIError
 from code42cli.errors import UserDoesNotExistError
+from code42cli.file_readers import read_csv_arg
 from code42cli.options import format_option
 from code42cli.options import sdk_options
-from code42cli.file_readers import read_csv_arg
 from code42cli.output_formats import DataFrameOutputFormatter
 from code42cli.output_formats import OutputFormat
 from code42cli.output_formats import OutputFormatter
@@ -129,6 +129,7 @@ def update_user(
         archive_size_quota,
     )
 
+
 _bulk_user_update_headers = [
     "user_id",
     "username",
@@ -137,8 +138,9 @@ _bulk_user_update_headers = [
     "first_name",
     "last_name",
     "notes",
-    "archive_size_quota"
+    "archive_size_quota",
 ]
+
 
 @users.group(cls=OrderedGroup)
 @sdk_options(hidden=True)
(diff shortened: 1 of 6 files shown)


src/code42cli/cmds/users.py
```diff
@@ -84,6 +93,95 @@ def remove_role(state, username, role_name):
...
+bulk.add_command(users_generate_template)
+
+
+@bulk.command(name="update")
+@read_csv_arg(headers=_bulk_user_update_headers)
+@format_option
+@sdk_options()
+def bulk_update(state, csv_rows, format):
+    """Update a list of users from the provided CSV."""
+    csv_rows[0]["updated"] = False
```
unparalleled-js, line 166: `updated` is a boolean here but a string on line 174

unparalleled-js, line 166: Why do we need this line `csv_rows[0]["updated"] = False`?

What happens if someone passes in a csv with no rows, does this through some IndexError?
^ if that is the case, and this is needed, we should check the length first and fail early with a better error

unparalleled-js:
In `code42cli.click_ext.groups.py`, can you import `Py42UsernameMustBeEmailError` from py42 (gets raised during `update_user()`) and add it to the exception list caught in lines 59-69?

tests/cmds/test_users.py
```diff
@@ -263,3 +275,48 @@ def test_remove_user_role_raises_error_when_username_does_not_exist(
     result = runner.invoke(cli, command, obj=cli_state)
     assert result.exit_code == 1
     assert "User 'not_a_username@example.com' does not exist." in result.output
+
+
+def test_update_user_calls_update_user_with_correct_parameters(
+    runner, cli_state, update_user_success
+):
+    command = ["users", "update", "--user-id", "12345", "--email", "test_email"]
```
unparalleled-js, line 283: Can you include all the args here so the test confirms they all get passed to the right spot?

src/code42cli/cmds/users.py
ceciliastevens: Updated to use a string here.

ceciliastevens: This is to add the column, for some reason if we don't add it beforehand, the addition during handle_row wasn't working properly. I copied this approach from how `devices bulk deactivate` does it.

The CSV is validated by read_csv(), which will raise an error if the CSV does not contain the proper rows.

tests/cmds/test_users.py
ceciliastevens: I wanted to test also that it properly passed `None` where no argument was given (since we don't want the method to accidentally update all unchanged fields to blank values). I've added additional tests to validate that all parameters are passed properly.

ceciliastevens (commented):
> In `code42cli.click_ext.groups.py`, can you import `Py42UsernameMustBeEmailError` from py42 (gets raised during `update_user()`) and add it to the exception list caught in lines 59-69?

I've added this.

unparalleled-js (approved):
Do you mind also changing `cmd/devices/.py` where it says
`csv_rows[0]["deactivated"] = False`
to 
`csv_rows[0]["deactivated"] = "False"`

that would be helpful.

This PR looks good!

ceciliastevens:
> Do you mind also changing `cmd/devices/.py` where it says
> `csv_rows[0]["deactivated"] = False`
> to
> `csv_rows[0]["deactivated"] = "False"`
> 
> that would be helpful.
> 
> This PR looks good!

Sure, I'll do that now.

PR #301, opened 2021-06-25
add move and bulk move methods

ceciliastevens:
**Description of Change**

This change adds `users move` and `users bulk move` methods to change a user's organization.

**Testing Procedure**

```
code42 users move --username <username> --org-id <org_id>
code42 users bulk move bulk_move_spreadsheet.csv
```

**Checklist**

### DIFF
--- a/src/code42cli/cmds/users.py
+++ b/src/code42cli/cmds/users.py
@@ -290,5 +290,5 @@ def _update_user(
 
 
 def _change_organization(sdk, username, org_id):
-    user_id=_get_user_id(sdk, username)
+    user_id = _get_user_id(sdk, username)
     return sdk.users.change_org_assignment(user_id=int(user_id), org_id=int(org_id))
--- a/src/code42cli/cmds/users.py
+++ b/src/code42cli/cmds/users.py
@@ -41,6 +41,13 @@ user_uid_option = click.option(
     "--user-id", help="The unique identifier of the user to be modified.", required=True
 )
 
+org_id_option = click.option(
+    "--org-id",
+    help="The identifier for the organization to which the user will be moved.",
+    required=True,
+    type=int,
+)
+
 
 def role_name_option(help):
     return click.option("--role-name", help=help)
@@ -141,6 +148,19 @@ _bulk_user_update_headers = [
     "archive_size_quota",
 ]
 
+_bulk_user_move_headers = ["user_id", "org_id"]
+
+
+@users.command(name="move")
+@click.option(
+    "--user-id", help="The identifier for the user to be moved", required=True, type=int
+)
+@org_id_option
+@sdk_options()
+def change_organization(state, user_id, org_id):
+    """Move the user with the given user ID to the org with the given org ID"""
+    _change_organization(state.sdk, user_id, org_id)
+
 
 @users.group(cls=OrderedGroup)
 @sdk_options(hidden=True)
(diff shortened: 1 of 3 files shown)


src/code42cli/cmds/users.py
```diff
@@ -41,13 +41,20 @@ def users(state):
...
+)
+
 
 def role_name_option(help):
     return click.option("--role-name", help=help)
 
 
-def username_option(help):
-    return click.option("--username", help=help)
+def username_option(help, required=False):
```
unparalleled-js, line 56: options are not required by default

src/code42cli/cmds/users.py
```diff
@@ -141,6 +148,17 @@ def update_user(
...
 ]
 
+_bulk_user_move_headers = ["username", "org_id"]
+
+
+@users.command(name="move")
+@username_option("The username of the user to be moved.", required=True)
+@org_id_option
+@sdk_options()
+def change_organization(state, username, org_id):
```
unparalleled-js, line 158: Would org name be a more friendly input param?

ceciliastevens: The difficulty with org name is that we don't have a good way to get the org ID from the org name, we'd have to do `sdk.orgs.get_all` and then iterate through the result to find what we want. Org names are also not guaranteed to be unique even within a tenant (they should be, but there are a few different ways to end up with non-unique org names).

src/code42cli/cmds/users.py
```diff
@@ -141,6 +148,17 @@ def update_user(
     "archive_size_quota",
 ]
 
+_bulk_user_move_headers = ["username", "org_id"]
+
+
+@users.command(name="move")
+@username_option("The username of the user to be moved.", required=True)
```
unparalleled-js, line 155: Don't need the `be`, could just say:

`The username of the user to move`

src/code42cli/cmds/users.py
```diff
@@ -141,6 +148,17 @@ def update_user(
...
+@org_id_option
+@sdk_options()
+def change_organization(state, username, org_id):
+    """Move the user with the given username to the org with the given org ID"""
```
unparalleled-js, line 159: Missing period

src/code42cli/cmds/users.py
ceciliastevens: Yes, but I needed to require it for the `users move` method, but I didn't want to change any of the other methods which have it as an optional option, so I added this way to pass that in from the decorator if we wanted, but left it defaulting to False. If there's a better way to do this, I'm happy to change it.

src/code42cli/cmds/users.py
```diff
@@ -182,6 +203,29 @@ def handle_row(**row):
     formatter.echo_formatted_list(result_rows)
 
 
+@bulk.command(name="move")
+@read_csv_arg(headers=_bulk_user_move_headers)
+@format_option
+@sdk_options()
+def bulk_move(state, csv_rows, format):
+    """Move the list of users from the provided CSV."""
```
unparalleled-js, line 211: In the help text here, we should say `Change the organization of ... ` to further explain what moving means

tests/cmds/test_users.py
```diff
@@ -388,3 +398,26 @@ def test_bulk_deactivate_uses_expected_arguments_when_all_are_passed(
...
+
+def test_bulk_move_uses_expected_arguments(runner, mocker, cli_state):
+    bulk_processor = mocker.patch(f"{_NAMESPACE}.run_bulk_process")
+    with runner.isolated_filesystem():
+        with open("test_bulk_move.csv", "w") as csv:
+            csv.writelines(["username,org_id\n", f"{TEST_USERNAME},4321\n"])
+        runner.invoke(
+            cli, ["users", "bulk", "move", "test_bulk_move.csv"], obj=cli_state
+        )
+    assert bulk_processor.call_args[0][1] == [
```
unparalleled-js, line 421: If this is just called once, you can use `assert_called_once_with()` and pass in the args and kwargs (the same as they get passed in the real call).

It helps ensure the method was also called the expected number of times.

ceciliastevens: What we're testing her is the call to `run_bulk_process`, which takes as its argument the `handle_row` method, the csv rows themselves, and a progress label. The test is only testing that the csv rows are passed through as expected, which is the same approach we take with testing the other bulk methods.

I've added another assertion to this test to verify that it's only called once.

src/code42cli/cmds/users.py
ceciliastevens: Updated

src/code42cli/cmds/users.py
ceciliastevens: Updated.

src/code42cli/cmds/users.py
ceciliastevens: Updated.

tests/cmds/test_users.py
unparalleled-js: Oh I see, it is because that method-call contains like a method and what-not.
Thanks, `assert_called_once()` is probably best here.

src/code42cli/cmds/users.py
unparalleled-js: Oh my bad,  I overlooked the wrapper

unparalleled-js (approved):

PR #303, opened 2021-07-01
Fix 2 Bulk Bugs in the CLI (addressing QA feedback)

unparalleled-js:
* `code42 users move` (unreleased) now uses the uid of the org as the ID instead of the legacy ID. Legacy IDs should not be in the CLI. This hinders performance, but there are plans to make API endpoints (soon I think) that accepts `orgUid`s instead of `orgId`s. We should rely on the backend to fix this issue and not have that in the CLI, especially if the fix is coming soon, that way we are not confusing users and we are using the most modern IDs and **to avoid breaking changes**.
* Fixes bug in certain bulk methods where it would output the incorrect successes and failures by the end of the process because it was swallowing all exceptions within its handlers.
* Additionally, Nikunj reported that newlines are not being handled correctly on Windows machines. I wrote some unit tests to prove that they at least work for Linux / OSX machines... Further investigation will be needed there.

Note of Caution: remember that weird shell issue where the text you type suddenly disappears and does not come back in the same shell process? If that happens, you will not see the stats at all in any circumstance... Weird right?

### DIFF
--- a/src/code42cli/cmds/devices.py
+++ b/src/code42cli/cmds/devices.py
@@ -592,18 +592,12 @@ def bulk_deactivate(state, csv_rows, change_device_name, purge_date, format):
             stats.increment_total_errors()
         return row
 
-    def handle_if_errors(*args, **kwargs):
-        """Still output rows to show individual row status by ignoring
-        the default CLI Log Error message.
-        """
-        pass
-
     result_rows = run_bulk_process(
         handle_row,
         csv_rows,
         progress_label="Deactivating devices:",
         stats=stats,
-        handle_if_errors=handle_if_errors,
+        raise_global_error=False,
     )
     formatter.echo_formatted_list(result_rows)
 
@@ -628,17 +622,11 @@ def bulk_reactivate(state, csv_rows, format):
             stats.increment_total_errors()
         return row
 
-    def handle_if_errors(*args, **kwargs):
-        """Still output rows to show individual row status by ignoring
-        the default CLI Log Error message.
-        """
-        pass
-
     result_rows = run_bulk_process(
         handle_row,
         csv_rows,
         progress_label="Reactivating devices:",
         stats=stats,
-        handle_if_errors=handle_if_errors,
+        raise_global_error=False,
     )
     formatter.echo_formatted_list(result_rows)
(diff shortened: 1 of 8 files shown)


unparalleled-js:
@ceciliastevens

src/code42cli/cmds/devices.py
```diff
@@ -611,6 +612,7 @@ def handle_row(**row):
             row["reactivated"] = "True"
         except Exception as err:
             row["reactivated"] = f"False: {err}"
+            raise
```
ceciliastevens, line 615: Will this change (and similar in the other handle_row methods) cause the output at the end to not include the failures?

unparalleled-js: It does include failures now _with_ this change. Before, it was all successes.
This error is again caught somewhere that manages that stats (worker.py)

unparalleled-js: If no error is raised in the handler, the stats incrementation logic considers the row a success

unparalleled-js: hmm...  I guess that individual status of the rows failures no longer show though...
ugh.

terimurie (approved):

src/code42cli/cmds/devices.py
ceciliastevens: Yeah, it was the output of all rows at the end that concerned me. Ideally it would be nice if we could both show the failures in the status as the job runs and also output them all at the end, but I'm not sure how best to actually accomplish that with the way it's structured currently.

unparalleled-js: I have an idea to make both worlds work. BRB!

unparalleled-js:
Also noticing that I don't always get the stats.. Seems like some race condition there

timabrmsn:
Can we add the expected CSV format in the bulk command help text like the other bulk commands?: https://github.com/code42/code42cli/blob/f8a4c334dadded7ba741b995333b963d72dab517/src/code42cli/cmds/departing_employee.py#L98-L99

src/code42cli/cmds/devices.py
unparalleled-js: Ok now it should work with both -- the bulk commands that need to display the row status as it goes along can do that and still have their final results correct at the end.

timabrmsn (approved):

PR #305, opened 2021-07-02
BugFixes: Allow checkpointing to work when API returns timestamps with nanoseconds

unparalleled-js:
* Fixes bug where search for audit logs would fail to store timestamps when the API would include nanoseconds for event timestamps in the response.
* Fix bug where if error occurred during processing audit log events, it would still store a timestamp and not process any of the events.
* Fixes 2fa bug for the new users bulk commands where you'd have to input your 2fa code for each row.
* Fixes bug where if using username with no value in csv for bulk users methods, it will use the first user from the list of all users...
* Adjusts error message for when user does not exist to also say that `or you do not have permission to view them` to handle cases where we call this method before making the real call that would normally result in 403.

I am copying this logic for Splunk and I need it to work in the CLI for the sake of testing.

* TEST *

Run a command (or similar logic of) `code42 jules verify-audit-log-dates` https://github.com/unparalleled-js/code42cli-extensions/blob/main/jules42.py#L64-L75
It outputs if it finds any unhandled timestamps.


src/code42cli/util.py
```diff
@@ -40,28 +41,31 @@ def get_user_project_path(*subdirs):
...
-def find_format_width(record, header, include_header=True):
+def find_format_width(records, header, include_header=True):
     """Fetches needed keys/items to be displayed based on header keys.
 
     Finds the largest string against each column so as to decide the padding size for the column.
 
     Args:
-        record (dict): data to be formatted.
-        header (dict): key-value where keys should map to keys of record dict and
+        records (list or dict): A list of data to be formatted.
```
unparalleled-js, line 50: This was always a list before but documented as a dict

src/code42cli/util.py
```diff
@@ -173,3 +177,10 @@ def hash_event(event):
     if isinstance(event, dict):
         event = json.dumps(event, sort_keys=True)
     return md5(event.encode()).hexdigest()
+
+
+def parse_timestamp(date_str):
+    # example: {"property": "bar", "timestamp": "2020-11-23T17:13:26.239647Z"}
+    ts = date_str[:-1]
```
unparalleled-js, line 184: borrowed from phcode42v2 and makes life easy

src/code42cli/util.py
```diff
@@ -40,28 +41,31 @@ def get_user_project_path(*subdirs):
...
     if include_header:
         if not header:
-            header = _get_default_header(record)
+            header = _get_default_header(records)
         rows.append(header)
     widths = dict(header.items())  # Copy
-    for record_row in record:
-        row = OrderedDict()
+    for record_row in records:
+        row = {}
```
unparalleled-js, line 68: all dicts are ordered in our supported versions of python

src/code42cli/util.py
unparalleled-js: really should just be using this everywhere, right? Not wanting to do that right now. Also, this PR makes me wish we were using the df formatter everywhere, we should try to prioritize that soon... And mock responses, either use `create_mock_response()` everywhere or your idea with generating response data...

timabrmsn (approved):
LGTM, nice job reworking the formatters to handle generators!

src/code42cli/util.py
timabrmsn: yeah, there have been a few things recently that I thought would be nice to just push into py42 to not have to rewrite everywhere..

PR #307, opened 2021-07-08
Release prep - 1.8.0

unparalleled-js:
Fixes some Audit Log bugs as well as new Users Commands


CHANGELOG.md
```diff
@@ -33,8 +33,6 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 ### Changed
 
-- `code42 alerts search` now uses microsecond precision when searching by alerts' date observed.
```
unparalleled-js, line 36: This does not appear to be true for some reason, so removing it for now.

unparalleled-js: Not true in the sense that when I do `-b "2021-07-06 12:23:23.235322"`, it is sent as `2021-07-06T12:23:23.000000Z`

timabrmsn: This would need to be updated to allow for sub-second precision:

https://github.com/code42/code42cli/blob/main/src/code42cli/click_ext/types.py#L123-L130

timabrmsn (approved):

PR #308, opened 2021-07-14
explicitly require chardet since requests 2.26.0 no longer depends on it

timabrmsn:
https://docs.python-requests.org/en/latest/community/updates/#release-history


unparalleled-js (approved):

setup.py
```diff
@@ -31,6 +31,7 @@
     zip_safe=False,
     python_requires=">3, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, !=3.5.*, <4",
     install_requires=[
+        "chardet>=4.0.0",
```
alanag13, line 34: is this needed for install or only to run tests?

If actually for install, we should get this released right away since fresh installs won't work.
If only needed for tests, we can move chardet to dev dependencies.

timabrmsn: For install. We're using it to detect file encodings when we read bulk files.

alanag13: sounds good, is master in a state where we can safely bump the version and cut a release? Might as well bump the version in this PR too if that is the case.

timabrmsn: I bumped the version and made a changelog note so we could release on this PR.

PR #310, opened 2021-08-09
Bugfix/2FA logic

timabrmsn:
Updates CLI MFA logic to key off of new py42's new Py42Unauthorized exception text that contains the LoginConfig value.

Now profile commands that validate password (`create`, `update`, `reset-pw`) require a TOTP token to be provided (either via `--totp` option or when prompted), as the server no longer returns any indication the password is correct but the 2FA token is not. 

I also added the `--debug` option to profile commands that make API calls, to aide future potential troubleshooting needs.

And since I noticed that if you try to delete a profile name that doesn't exist it didn't print a clear error (just pointed you to the log file), I added a fix for that as well.

FYI: This is dependent on next release of py42, so builds/tests will fail until then.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,13 +10,13 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 ## Unreleased
 
-### Added 
+### Added
 
 - `code42 profile` commands that validate passwords (`create`, `update`, `reset-pw`) now have `--totp` and `--debug` options available.
 
 ### Changed
 
-- A TOTP token is now required on `code42 profile` commands that check for password validity when a user has MFA enabled. 
+- A TOTP token is now required on `code42 profile` commands that check for password validity when a user has MFA enabled.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 ### Added
 
-- `code42 profile` commands that validate passwords (`create`, `update`, `reset-pw`) now have `--totp` and `--debug` options available.
+- `code42 profile` commands that validate passwords (`create`, `update`, `reset-pw`) now have the `--debug` option available, and `create` and `update` can now also pass in `--totp` as an option.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 The intended audience of this file is for py42 consumers -- as such, changes that don't affect
 how a consumer would use the library (e.g. adding unit tests, updating documentation, etc) are not captured here.
 
+## Unreleased
+
+### Added 
+
+- `code42 profile` commands that validate passwords (`create`, `update`, `reset-pw`) now have `--totp` and `--debug` options available.
+
+### Changed
+
+- A TOTP token is now required on `code42 profile` commands that check for password validity when a user has MFA enabled. 
+
+### Fixed
+
+- `code42 profile delete` command now prints a clear error message when deletion target doesn't exist.
+
 ## 1.8.1 - 2021-07-14
 
 ### Fixed
--- a/src/code42cli/cmds/profile.py
+++ b/src/code42cli/cmds/profile.py
@@ -23,6 +23,7 @@ debug_option = click.option(
 )
 totp_option = click.option("--totp", help="TOTP token for multi-factor authentication.")
 
+
 def profile_name_arg(required=False):
     return click.argument("profile_name", required=required)
 
--- a/src/code42cli/cmds/profile.py
+++ b/src/code42cli/cmds/profile.py
@@ -3,7 +3,6 @@ from getpass import getpass
 import click
 from click import echo
 from click import secho
-from py42.exceptions import Py42MFARequiredError
 
 import code42cli.profile as cliprofile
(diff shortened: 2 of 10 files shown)


src/code42cli/cmds/profile.py
```diff
@@ -124,12 +133,13 @@ def update(name, server, username, password, disable_ssl_errors):
...
 @profile_name_arg()
-def reset_pw(profile_name):
+@debug_option
+def reset_pw(profile_name, debug):
     """\b
     Change the stored password for a profile. Only affects what's stored in the local profile,
     does not make any changes to the Code42 user account."""
     password = getpass()
-    profile_name_saved = _set_pw(profile_name, password)
+    profile_name_saved = _set_pw(profile_name, password, debug)
```
unparalleled-js, line 142: It seems there are missing tests for the cases when passing `debug` and `totp` to these profile commands that show that those arguments are used when creating the sdk

timabrmsn: good catch

src/code42cli/cmds/profile.py
```diff
@@ -19,6 +18,12 @@ def profile():
     pass
 
 
+debug_option = click.option(
+    "-d", "--debug", is_flag=True, help="Turn on debug logging.",
+)
+totp_option = click.option("--totp", help="TOTP token for multi-factor authentication.")
```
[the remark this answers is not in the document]
timabrmsn: yeah, I suppose it is better to validate in the argument parsing rather than waiting until a couple of http requests complete..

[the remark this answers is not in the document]
timabrmsn: created a new click.ParamType `TOTP` to validate token args.

src/code42cli/sdk_client.py
```diff
@@ -47,13 +46,18 @@ def _validate_connection(authority_url, username, password, totp=None):
...
         raise LoggedCLIError(f"Problem connecting to {authority_url}.")
-    except Py42MFARequiredError:
-        totp = prompt("Multi-factor authentication required. Enter TOTP", type=int)
-        return _validate_connection(authority_url, username, password, totp)
     except Py42UnauthorizedError as err:
         logger.log_error(err)
-        if "INVALID_TIME_BASED_ONE_TIME_PASSWORD" in err.response.text:
-            raise Code42CLIError(f"Invalid TOTP token for user {username}.")
+        if "LoginConfig: LOCAL_2FA" in str(err):
+            if totp is None:
```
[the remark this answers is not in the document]
timabrmsn: The new `TOTP` param type validation handles this case now.

CHANGELOG.md
```diff
@@ -8,6 +8,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 The intended audience of this file is for py42 consumers -- as such, changes that don't affect
 how a consumer would use the library (e.g. adding unit tests, updating documentation, etc) are not captured here.
 
+## Unreleased
+
+### Added
+
+- `code42 profile` commands that validate passwords (`create`, `update`, `reset-pw`) now have `--totp` and `--debug` options available.
```
unparalleled-js, line 15: It actually does not look like `reset-pw` has the `totp`, and I understand why

tests/cmds/test_profile.py
```diff
@@ -530,3 +509,75 @@ def test_use_profile(runner, mock_cliprofile_namespace, profile):
...
+            "-n",
+            "foo",
+            "-s",
+            "bar",
+            "-u",
+            "baz",
+            "--password",
+            "testpass",
+            "--totp",
+            totp1,
```
unparalleled-js, line 534: I am surprised this test passes. Does the `totp` validation logic not happen in the test runner?

src/code42cli/cmds/profile.py
```diff
@@ -103,7 +113,9 @@ def create(name, server, username, password, disable_ssl_errors):
 @username_option()
 @password_option
 @disable_ssl_option
-def update(name, server, username, password, disable_ssl_errors):
+@debug_option
+@totp_option
```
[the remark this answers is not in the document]
timabrmsn: Yeah, if a user doesn't have a password saved and they don't provide the `--password` option, it'll prompt them if they want to save. Which won't use the `--totp` if they happen to have passed it, it'll just force a prompt (as it'll fail first but detect LOCAL_2FA LoginConfig and then prompt for totp). 

I'd rather ignore the option input since it'll prompt for it anyway rather than raise an error and tell the user they can't provide it as an option.

tests/cmds/test_profile.py
timabrmsn: why would it fail?

unparalleled-js: Because it uses the TOTP type which verifies that the value is an int with 6 digits?
Is that part mocked out or something?

timabrmsn: totp1 is an int (convertable from string) with 6 digits: https://github.com/code42/code42cli/pull/310/files/e15cb0006fb93dd1dd7e65d50f542f4bdeb03f78#diff-da10b77cdc04e2d1e4cee69562deba89b7e1eb31322da3129157212e08f3d2fcR517

timabrmsn:
> It fails to get a metavar.
updated.

tests/cmds/test_profile.py
unparalleled-js: Agh so sorry, I thought that was `"totp1"`

unparalleled-js (approved):

PR #314, opened 2021-08-19
Feature/query for risk indicators

tora-kozic:
Allowed for multiple `--risk-indicator` option flags to be provided to the `security-data search` command

### DIFF
--- a/src/code42cli/cmds/securitydata.py
+++ b/src/code42cli/cmds/securitydata.py
@@ -4,8 +4,8 @@ import click
 import py42.sdk.queries.fileevents.filters as f
 from c42eventextractor.extractors import FileEventExtractor
 from click import echo
-from py42.sdk.queries.fileevents.filters import RiskIndicator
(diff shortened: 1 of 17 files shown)


src/code42cli/cmds/securitydata.py
```diff
@@ -214,9 +214,11 @@
 def risk_indicator_callback(filter_cls):
     def callback(ctx, param, arg):
         if arg:
-            mapped_arg = (risk_indicator_map[arg[0]],)
+            mapped_args = ()
+            for i in range(len(arg)):
+                mapped_args += (risk_indicator_map[arg[i]],)
```
timabrmsn, line 219: This could be simplified to:

```
mapped_args = tuple(risk_indicator_map[i] for i in arg)
```

tora-kozic: Nice! I knew there was a better way than a for loop but I couldn't quite sort it out

timabrmsn (approved):

PR #319, opened 2021-09-08
Bugfix/improve invalid org uid error

peterbriggs42:
Testing:
`code42 users list --org-uid "invalid"`
`code42 devices list --org-uid "invalid"`
`code42 devices list-backup-sets --org-uid "invalid"`
Each of these commands should return error message `Error: Org with ID 'invalid' not found.`

### DIFF
--- a/src/code42cli/cmds/devices.py
+++ b/src/code42cli/cmds/devices.py
@@ -404,21 +404,15 @@ def _get_all_active_hold_memberships(sdk):
 def _get_device_dataframe(
     sdk, columns, active=None, org_uid=None, include_backup_usage=False
 ):
-    try:
-        devices_generator = sdk.devices.get_all(
-            active=active, include_backup_usage=include_backup_usage, org_uid=org_uid
-        )
-        devices_list = []
-        if include_backup_usage:
-            columns.append("backupUsage")
-        for page in devices_generator:
-            devices_list.extend(page["computers"])
-        return DataFrame.from_records(devices_list, columns=columns)
-    except Py42BadRequestError as err:
-        if "Unable to find org" in err.response.text:
-            raise Code42CLIError(f"Org with ID '{org_uid}' not found.")
-        else:
-            raise
+    devices_generator = sdk.devices.get_all(
+        active=active, include_backup_usage=include_backup_usage, org_uid=org_uid
+    )
+    devices_list = []
+    if include_backup_usage:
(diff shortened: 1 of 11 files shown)


src/code42cli/cmds/devices.py
```diff
@@ -403,15 +404,21 @@ def _get_all_active_hold_memberships(sdk):
...
+        devices_generator = sdk.devices.get_all(
+            active=active, include_backup_usage=include_backup_usage, org_uid=org_uid
+        )
+        devices_list = []
+        if include_backup_usage:
+            columns.append("backupUsage")
+        for page in devices_generator:
+            devices_list.extend(page["computers"])
+        return DataFrame.from_records(devices_list, columns=columns)
+    except Py42BadRequestError as err:
```
peterbriggs42, line 417: Ideally py42 would raise a Py42OrgNotFoundError here rather than Py42BadRequestError. Should we still go ahead with this fix for the device commands despite not having the preferred error abstraction available from py42?

src/code42cli/cmds/users.py
```diff
@@ -401,12 +402,16 @@ def _get_role_id(sdk, role_name):
...
-    return DataFrame.from_records(users_list, columns=columns)
+    try:
+        users_generator = sdk.users.get_all(
+            active=active, org_uid=org_uid, role_id=role_id
+        )
+        users_list = []
+        for page in users_generator:
+            users_list.extend(page["users"])
+        return DataFrame.from_records(users_list, columns=columns)
+    except Py42OrgNotFoundError:
```
timabrmsn, line 413: We can just add `Py42OrgNotFoundError` to the list of exceptions we handle in `code42cli.click_ext.groups.ExceptionHandlingGroup` then we don't need to handle it specifically in each command.

timabrmsn: Then we can update py42 to correctly raise a `Py42OrgNotFoundError` from `sdk.devices.get_page()` (and maybe a few other custom ones where a bad value could be passed). Then just maybe wait to release the CLI until the py42 issue has been fixed.

peterbriggs42: Yep, that sounds like the right move. I'll remove this handling from the device list fetch and open a PR against py42 to properly raise the exception there.

timabrmsn (approved):
LGTM

PR #320, opened 2021-09-08
Chore/remove extractor

tora-kozic:
Removing the extractor from the CLI.  It was mainly used when returning results for the `search()` and `send_to()` functions for `alerts` and `file events`. 

The extractor was replaced by making py42 calls directly and by moving some of the logic out from the extractor and into those `alerts` and `security-data` command modules, as well as a shared `util` file between them.

For `security-data` events, `eventId` page tokens were implemented for checkpointing, replacing the timestamps. 

For `alerts`, timestamp checkpoints were implemented in a similar fashion to the `audit-logs` checkpointing.  However, without the de-dupe logic due to the rarity of two alerts having an identical timestamp - but I don't think it'd be too hard to include that if it was preferred.


src/code42cli/cmds/securitydata.py
```diff
@@ -358,23 +365,83 @@ def search(
...
+        # otherwise it should set the initial value to ""
+        checkpoint = cursor.get(checkpoint_name) or ""
+    else:
+        checkpoint = ""
+
+    # older app versions stored checkpoint as float timestamp.
+    # we handle those here until the next run containing events will store checkpoint as the last eventId
+    if isinstance(checkpoint, (int, float)):
+        state.search_filters.append(InsertionTimestamp.on_or_after(checkpoint))
+        checkpoint = ""
```
timabrmsn, line 387: Can we move this up within the `if use_checkpoint` block since this only matters when that condition is met?

src/code42cli/cmds/securitydata.py
```diff
@@ -358,23 +365,85 @@ def search(
...
+        state.search_filters = convert_to_or_query(state.search_filters)
+    verify_filter_groups(state.search_filters)
+    query = FileEventQuery(*state.search_filters)
+    query.page_size = MAX_EVENT_PAGE_SIZE
+    query.sort_direction = "asc"
+    query.sort_key = "insertionTimestamp"
+    return query
+
+
+def _get_all_file_events(state, query, checkpoint=""):
```
tora-kozic, line 425: Because of the page tokens, would it be better for this to return a generator of either events or responses and then for it to handle each individually and update the checkpoint as it goes, rather than building the whole list right away?

src/code42cli/cmds/search/cursor_store.py
```diff
@@ -30,15 +30,20 @@ def get(self, cursor_name):
         try:
             location = path.join(self._dir_path, cursor_name)
             with open(location) as checkpoint:
-                return float(checkpoint.read())
+                try:
+                    return float(checkpoint.read())
+                except ValueError:
```
tora-kozic, line 35: I based the `alert` checkpoints off the existing `audit-log` checkpoints, and found that the audit logs threw an error that just directed you to the log when it ran out of checkpoints (when trying to convert an empty string to a float).

I added this catch to at least send a more specific message for now but I'm not sure if there's a different solution that would be better - For instance, checking if the checkpoint is empty and then just returning an empty string instead of trying to do the type conversion.

src/code42cli/cmds/securitydata.py
timabrmsn: Yeah, I would lean towards just yielding each event individually from `_get_all_file_events()`, then when using a checkpoint you just wrap the event generator with `_store_updated_checkpoint()` and pass it to `echo_formatted_generated_output()` or pass it there directly.

src/code42cli/cmds/search/cursor_store.py
timabrmsn: I'm not quite sure what you mean by `ran out of checkpoints`, since there should only be one saved from the last run (that is now the starting point for the next run). Could an empty string have been accidentally saved while testing the code? 

Although it isn't a bad idea to raise a helpful error message here if we do happen to get bad data into a checkpoint file. Maybe something like `Unable to parse checkpoint from {checkpoint_file_path}, expected a unix-epoch timestamp, got {bad_data}.` That way if a user was looking at the checkpoint file and accidentally added a character or something, they'd have the info they needed to fix it (or delete the file to reset).

src/code42cli/cmds/securitydata.py
tora-kozic: Yup! should be up there now

src/code42cli/cmds/search/cursor_store.py
tora-kozic: Thats a more helpful message, I'll adjust that! 

To your first point, maybe I should've said "ran out of events," as in there is no stored checkpoint because there are no more events - if that makes sense? Like if I run an `audit-log` search and make subsequent runs until all the events are returned, the checkpoint file is empty.  This throws an error when converting to a float.

timabrmsn: The checkpoint file should always have _something_ in it at the end of each CLI run (unless there aren't any events to be found on the very first run), as we're mainly checkpointing to be able to pick up where we left off between CLI invocations. 

Example: 

If I were an analyst wanting to just quickly check over the audit logs each day (but didn't want to have to see events I've already seen), I could run `code42 audit-log search -b 1d --use-checkpoint daily` today, look over the events, and the timestamp of the last event in the list would get saved to the `daily` checkpoint file. 

Then each subsequent day I could re-run `code42 audit-log search --use-checkpoint daily` and I'd only get events I haven't yet seen, cause the checkpoint is always updating at the end of each run with the latest seen timestamp (for file events, it'll now be the latest seen eventId).

alanag13: > unless there aren't any events to be found on the very first run

This could definitely be possible, especially depending on what the query was. Would we try to save an empty string in this scenario?

timabrmsn: Looks like it doesn't create the checkpoint file if no results are found, but if the command gets cancelled before the checkpoint is written we can end up with a blank checkpoint file, which then causes the `ValueError: could not convert string to float: ''`

So maybe we should handle the empty-file condition and just return None, but if there's _something_ in the checkpoint file but we can't parse it raise the `Unable to parse` message from above.

alanag13:
I haven't taken a super close look at this yet so maybe this is being handled, but just wanted to call out something we had to be careful about with the Splunk app:

Currently, checkpoints for file events are saved as timestamps. This will change to strings when we start using the page token instead -- If a user is is currently on a version of the CLI that uses timestamps and then upgrades, we need to make sure this change is backwards-compatible.

src/code42cli/cmds/alerts.py
```diff
@@ -242,21 +234,75 @@ def search(
...
+
+    if not alerts_list:
+        click.echo("No results found.")
+        return
+
+    if use_checkpoint:
+        checkpoint_name = use_checkpoint
+        # update checkpoint to alertId of last event retrieved
+        alerts_gen = _store_updated_checkpoint(cursor, checkpoint_name, alerts_list)
+        formatter.echo_formatted_generated_output(alerts_gen)
```
timabrmsn, line 259: the `audit-logs` command was broken because this method didn't account for the `JSON/RAW-JSON/CSV` formats, so I removed it altogether. But I realize here we're going to have the same problem that there's currently not a good way to handle a generator of events for the non-Table formats... 

Maybe I should invest the time and try to sort that out now.

timabrmsn: I think we'll eventually be able to make the csv, and json format types handle generators, but I don't think table will ever really work with a generator, since you need to know what the max length of each column value is to get the spacing of the table right, and for that you need to have all the data up front.

It still is useful to use a generator for send-to though, since we're actually processing each event one at a time.

So just calling `list` on the generator before printing in `search` is fine for now.

src/code42cli/cmds/alerts.py
```diff
@@ -16,18 +15,26 @@
...
 from code42cli.output_formats import OutputFormat
 from code42cli.output_formats import OutputFormatter
-
+from code42cli.util import hash_event
+from code42cli.util import parse_timestamp
+from code42cli.util import warn_interrupt
 
 ALERTS_KEYWORD = "alerts"
+MAX_ALERT_PAGE_SIZE = 500
+_ALERT_DETAIL_BATCH_SIZE = 100
```
timabrmsn, line 36: Could resolve NTEG-1917 while we're at it here.

src/code42cli/cmds/alerts.py
```diff
@@ -242,21 +235,92 @@ def search(
...
+    if advanced_query:
+        state.search_filters = advanced_query
+    else:
+        if begin or end:
+            state.search_filters.append(
+                create_time_range_filter(f.DateObserved, begin, end)
+            )
+    if or_query:
+        state.search_filters = convert_to_or_query(state.search_filters)
+    verify_filter_groups(state.search_filters)
```
timabrmsn, line 278: I don't think this check is necessary anymore, since it's just our code that is constructing the filters. The extractor would get them passed in, so it was possible to get something that wasn't actually a filter group.

src/code42cli/cmds/alerts.py
timabrmsn: I think just the detail batch, as that's where they were seeing performance issues.

src/code42cli/cmds/alerts.py
```diff
@@ -381,3 +461,17 @@ def _update_alert(sdk, alert_id, alert_state, note):
         sdk.alerts.update_state(alert_state, [alert_id], note=note)
     elif note:
         sdk.alerts.update_note(alert_id, note)
+
+
+def _get_alert_details(sdk, alert_summary_list):
```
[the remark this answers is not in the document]
timabrmsn: We can delete this function, delete `_get_all_alerts()` and just do:

```
alerts_list = list(cli_state.sdk.alerts.get_all_alert_details(query))
```

src/code42cli/cmds/securitydata.py
```diff
@@ -358,23 +365,77 @@ def search(
...
-    )
+    if use_checkpoint:
+        checkpoint_name = use_checkpoint
+        # if checkpoint name exists, checkpoint should be that eventId,
+        # otherwise it should set the initial value to ""
+        checkpoint = cursor.get(checkpoint_name) or ""
+
+        # older app versions stored checkpoint as float timestamp.
+        # we handle those here until the next run containing events will store checkpoint as the last eventId
+        if isinstance(checkpoint, (int, float)):
```
timabrmsn, line 383: Hrm, since the FileEventCursorStore.get() returns a string, this will never be true... We'll have to try calling `float()` on the checkpoint instead of `isinstance()`.

src/code42cli/cmds/securitydata.py
```diff
@@ -431,16 +492,36 @@ def send_to(
...
+        click.echo("No results found.")
+        return
 
-def _get_file_event_extractor(sdk, handlers):
-    return FileEventExtractor(sdk, handlers)
+    for event in events:
+        if use_checkpoint:
+            checkpoint_name = use_checkpoint
+            cursor.replace(checkpoint_name, event["eventId"])
+        with warn_interrupt():
```
timabrmsn, line 521: The `warn_interrupt()` context should wrap the whole for loop, otherwise the user would need to hit ctrl-c twice within the timeframe it takes to process a single event, otherwise the counter would get reset.

tests/cmds/test_alerts.py
```diff
@@ -415,26 +448,30 @@ def test_send_to_with_advanced_query_and_incompatible_argument_errors(
...
     begin_date = get_test_date_str(days_ago=89)
     end_date = get_test_date_str(days_ago=1)
 
     runner.invoke(
         cli, [*command, "--begin", begin_date, "--end", end_date], obj=cli_state,
     )
-    filters = alert_extractor.extract.call_args[0][0]
-    actual_begin = get_filter_value_from_json(filters, filter_index=0)
+    query = cli_state.sdk.alerts.search_all_pages.call_args[0][0]
+    query_dict = {k: v for k, v in query}
```
timabrmsn, line 460: you can actually just do `dict(query)` to convert it to a normal dict

timabrmsn:
I should note we don't want to release this until the `get_all_alert_details()` fix is released in py42.

tora-kozic:
> I should note we don't want to release this until the `get_all_alert_details()` fix is released in py42.

Yeah, I'll make sure to bump the py42 version requirement too when that happens.

PR #321, opened 2021-09-09
PATH troubleshooting helpers/docs

timabrmsn:
This change enables the CLI to be called as a module: `python -m code42cli <command>` is now equivalent to `code42 <command>` 

This makes it possible to still use the CLI if python hasn't installed its scripts directory into the user's PATH. 

Additionally, the `--script-dir` option has been added to the base command, which prints the directory the `code42` script had been installed into, which can then be used to add the correct directory to PATH so `code42` can be called directly.

GettingStarted guide has also been updated to include the above info.

### DIFF
--- a/docs/userguides/gettingstarted.md
+++ b/docs/userguides/gettingstarted.md
@@ -148,7 +148,7 @@ Add the above to your shell configuration file to persist the change.
 
 #### On Windows:
 
-```powershell 
+```powershell
 $env:Path += ";$(python -m code42cli --script-dir)"
 ```
 
--- a/docs/userguides/gettingstarted.md
+++ b/docs/userguides/gettingstarted.md
@@ -135,7 +135,7 @@ You can execute the CLI by calling the python module directly:
 python3 -m code42cli
 ```
 
-And the base `code42` command now has a `--script-dir` option that will print out the directory the `code42` script was 
+And the base `code42` command now has a `--script-dir` option that will print out the directory the `code42` script was
 installed into, so you can manually add it to your PATH, enabling the `code42` command to work.
 
 #### On Mac/Linux:
--- a/docs/userguides/gettingstarted.md
+++ b/docs/userguides/gettingstarted.md
@@ -123,7 +123,7 @@ To learn more about authenticating in the CLI, follow the [Configure profile gui
 
 ## Troubleshooting and support
 
-### PATH issues
+### Code42 command not found
 
 If your python installation has added itself to your environment's PATH variable, then running `code42` _should_ just work.
 
@@ -140,11 +140,11 @@ installed into, so you can manually add it to your PATH, enabling the `code42` c
 
 #### On Mac/Linux:
 
+Run the following to make `code42` visible in your shell's PATH (to persist the change, add it to your shell's configuration file):
+
 ```bash
 export PATH=$PATH:$(python3 -m code42cli --script-dir)
 ```
-
-Add the above to your shell configuration file to persist the change.
 
 #### On Windows:
 
@@ -152,7 +152,7 @@ Add the above to your shell configuration file to persist the change.
 $env:Path += ";$(python -m code42cli --script-dir)"
 ```
 
-Then to store the updated PATH variable permanently:
+To persist the change, add the updated PATH to your registry:
 
 ```powershell
 Set-ItemProperty -Path 'Registry::HKEY_LOCAL_MACHINE\System\CurrentControlSet\Control\Session Manager\Environment' -Name PATH -Value $env:Path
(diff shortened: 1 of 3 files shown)


docs/userguides/gettingstarted.md
```diff
@@ -123,6 +123,41 @@ To learn more about authenticating in the CLI, follow the [Configure profile gui
...
+
+```bash
+export PATH=$PATH:$(python3 -m code42cli --script-dir)
+```
+
+Add the above to your shell configuration file to persist the change.
+
+#### On Windows:
+
+```bash
```
alanag13, line 151: `bat` and `powershell` are supported keys for markdown syntax highlighting

src/code42cli/main.py
```diff
@@ -53,13 +55,28 @@ def exit_on_interrupt(signal, frame):
...
+    help="Print the directory the `code42` script was installed in (for adding to your PATH if needed).",
 )
 @sdk_options(hidden=True)
-def cli(state, python):
+def cli(state, python, script_dir):
     if python:
         click.echo(sys.executable)
         sys.exit(0)
+    if script_dir:
+        for root, _dirs, files in os.walk(site.PREFIXES[0]):
```
alanag13, line 71: `site-packages` can be global or per-user, so I would expect that you could get rid of the second for loop by just looping through `PREFIXES` instead of only looking at its first element.

timabrmsn: in all of the installs I've tested, site.PREFIXES had the same value for both items in the list (if there was more than one): 

```
/usr/bin/python -c 'import site; print(site.PREFIXES)'
['/System/Library/Frameworks/Python.framework/Versions/2.7', '/System/Library/Frameworks/Python.framework/Versions/2.7']

python3.9 -c 'import site; print(site.PREFIXES)'
['/Users/tim.abramson/.pyenv/versions/3.9.5', '/Users/tim.abramson/.pyenv/versions/3.9.5']
```
etc.

docs/userguides/gettingstarted.md
timabrmsn: 👍  Pycharm wasn't highlighting the contents when I tried it, but it works in the actual docs. Updated.

src/code42cli/main.py
alanag13: hmm, me too, I must be misunderstanding something about `PREFIXES`. What you have definitely works so I have no problem with getting this out, just thought I was eliminating some code here but I guess not

alanag13 (approved):

docs/userguides/gettingstarted.md
```diff
@@ -123,6 +123,41 @@ To learn more about authenticating in the CLI, follow the [Configure profile gui
 
 ## Troubleshooting and support
 
+### PATH issues
```
timabrmsn, line 126: @annie-payseur can I have you look over these new troubleshooting docs?

annie-payseur: For troubleshooting info, describe the symptom the user experiences, not the cause. So I'd recommend changing this sub-head from "PATH issues" to "Code42 command not found".

docs/userguides/gettingstarted.md
```diff
@@ -123,6 +123,41 @@ To learn more about authenticating in the CLI, follow the [Configure profile gui
...
+And the base `code42` command now has a `--script-dir` option that will print out the directory the `code42` script was
+installed into, so you can manually add it to your PATH, enabling the `code42` command to work.
+
+#### On Mac/Linux:
+
+```bash
+export PATH=$PATH:$(python3 -m code42cli --script-dir)
+```
+
+Add the above to your shell configuration file to persist the change.
```
annie-payseur, line 147: I'd move this above the code block and edit it to say "To persist the change, add the below to your shell configuration file."

docs/userguides/gettingstarted.md
```diff
@@ -123,6 +123,41 @@ To learn more about authenticating in the CLI, follow the [Configure profile gui
...
+
+#### On Windows:
+
+```powershell
+$env:Path += ";$(python -m code42cli --script-dir)"
+```
+
+Then to store the updated PATH variable permanently:
```
annie-payseur, line 155: Remove "Then".

docs/userguides/gettingstarted.md
```diff
@@ -123,6 +123,41 @@ To learn more about authenticating in the CLI, follow the [Configure profile gui
...
+#### On Mac/Linux:
+
+```bash
+export PATH=$PATH:$(python3 -m code42cli --script-dir)
+```
+
+Add the above to your shell configuration file to persist the change.
+
+#### On Windows:
+
```
annie-payseur, line 150: Consider adding the same statement you have for Mac/Linux ("To persist the change...") to facilitate scanning.

PR #324, opened 2021-09-24
note required nature of --begin

timabrmsn:


src/code42cli/options.py
```diff
@@ -165,7 +165,7 @@ def server_options(f):
 def begin_option(term, **kwargs):
     defaults = dict(
         type=MagicDate(rounding_func=round_datetime_to_day_start),
-        help=f"The beginning of the date range in which to look for {term}. {MagicDate.HELP_TEXT}",
+        help=f"The beginning of the date range in which to look for {term}. {MagicDate.HELP_TEXT} [required unless --use-checkpoint option passed]",
```
alanag13, line 168: Nitpick, but `passed` seems like odd terminology for the use of a terminal flag. Maybe just `used`?

alanag13 (approved):

PR #333, opened 2021-10-20
Exclude recently connected devices prior to filtering by date.

### LINKED ISSUE #268: [Bug] Combining filter options in "devices list" leads to unexpected behavior
### Description

The order of operations when filtering a device list can be confusing and lead to unexpected behavior. Specifically I've seen a case where a user had multiple devices, one of which had connected recently, and then found that "devices list --last-connected-before 30d --exclude-most-recently-connected 1" did not list a device that they felt it should have.

### Steps to Reproduce

1. Register two devices to a user in Code42
2. Prevent one from backing up for a period of time
3. Run `code42 devices list --last-connected-before <some time period> --exclude-most-recently-connected 1`

### Expected Behavior
The device that is not backing up is listed.

### Actual Behavior
Neither device is listed.

### Analysis

In `list_devices()` we process the connected timestamps first: https://github.com/code42/code42cli/blob/master/src/code42cli/cmds/devices.py#L339-L340

and then process things like dropping latest connected after:
https://github.com/code42/code42cli/blob/master/src/code42cli/cmds/devices.py#L339-L340

If we process "drop last connected" first, then we both have a more consistent result (the devices excluded don't change depending on your timestamps), and a more intuitive behavior.

### Basic Information

- code42cli version: 1.4.1
- python version: 3.9.2
- operating system: observed on Windows and Mac OS

ceciliastevens:
Closes #268.

This change moves the "exclude recently connected" filtering to _before_ the last-connected-before or created-before filtering. This better matches the expected behavior that we've been told about by our customers.

I also added a test for this change to document this behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 ### Fixed
 
 - Incorrect column title on `code42 trusted-activities bulk create` command help text.
+- `code42 devices list` will now process `--exclude-most-recently-connected` prior to `--last-connected-before` instead of after.
 
 ### Added
 


ceciliastevens:
Added. Let me know if you'd like me to update the wording at all.

PR #334, opened 2021-10-21
fix python version requirements

alanag13:
When running the cli on python 3.6.1 or 3.6.0, you will get

```
ImportError: cannot import name 'NoReturn'
```

On startup. This is because `typing.NoReturn` was not introduced until python 3.6.2, and ipython references this module.

This correctly updates our install requirements to require 3.6.2.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,8 +16,7 @@ Use the `code42` command to interact with your Code42 environment.
 
 ## Requirements
 
-- Python 3.6.0+
-- Code42 Server 6.8.x+
+- Python 3.6.2+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -16,8 +16,7 @@ Use the `code42` command to interact with your Code42 environment.
 
 ## Requirements
 
-- Python 3.6.0+
-- Code42 Server 6.8.x+
```
alanag13, line 20: got rid of this server reference because of on-prem deprecation

PR #341, opened 2021-11-09
Improved DataFrameOutputFormatter

timabrmsn:
This PR updates the `DataFrameOutputFormatter` to enable checkpointing of data as it is printed/processed, regardless of the output format. Just pass a `checkpoint_func` to the constructor of the `DataFrameOutputFormatter` and it will be called for each row of output printed/processed.

It also adds easy column filtering, so reducing the output to a default or user-provided set of columns can be easily added to _any_ output. An example of this is now in the `security-data` commands with the `--columns` option.

Ideally every command that outputs in table-like formats would be eventually converted over to use it, but I've only updated the `security-data` commands to keep this PR manageable.


timabrmsn:
There's an integration test failing because we don't have the risk properties (`riskIndicators`, `riskScore`, `riskSeverity`) on the mock response and the default column filter for table events is expecting them. I'm opening a PR to add those now.

src/code42cli/output_formats.py
```diff
@@ -85,57 +69,213 @@ def _requires_list_output(self):
...
+            raise Code42CLIError(
                 f"DataFrameOutputFormatter received an invalid format: {self.output_format}"
             )
 
-    def echo_formatted_dataframe(self, df, **kwargs):
-        str_output = self.get_formatted_output(df, **kwargs)
-        if len(df) <= OUTPUT_VIA_PAGER_THRESHOLD:
-            click.echo(str_output)
+    def echo_formatted_dataframes(
+        self, dfs, columns=None, force_pager=False, force_no_pager=False, **kwargs
```
tora-kozic, line 227: I haven't seen any way this `force_pager` variable is ever true - I assume this is just for potential future applications? (or I'm missing something)

src/code42cli/cmds/securitydata.py
```diff
@@ -362,84 +337,92 @@ def search(
...
 ):
 
     """Search for file events."""
-    output_header = try_get_default_header(
-        include_all, _create_search_header_map(), format
-    )
-    formatter = FileEventsOutputFormatter(format, output_header)
-    cursor = _get_cursor(state, use_checkpoint)
+    if format == FileEventsOutputFormat.CEF and columns:
+        raise click.BadOptionUsage("--columns option can't be used with CEF format.")
```
tora-kozic, line 347: this method takes the args (option name, message) so we you want the message to print the console I think it should be like -> 
`raise click.BadOptionUsage(columns, "--columns option can't be used with CEF format.")`

src/code42cli/output_formats.py
timabrmsn: yeah, it's just an arg someone can pass that overrides the 'only use pager when more than 10 rows' logic. We don't have any current use-cases for it though..

src/code42cli/output_formats.py
```diff
@@ -85,57 +69,213 @@ def _requires_list_output(self):
...
-            raise ValueError(
+            raise Code42CLIError(
                 f"DataFrameOutputFormatter received an invalid format: {self.output_format}"
             )
 
-    def echo_formatted_dataframe(self, df, **kwargs):
-        str_output = self.get_formatted_output(df, **kwargs)
-        if len(df) <= OUTPUT_VIA_PAGER_THRESHOLD:
-            click.echo(str_output)
+    def echo_formatted_dataframes(
```
tora-kozic, line 226: If no results are found, this prints nothing.  Not sure if we want to echo the usual "No results found." here or in the method where this `echo_formatted_dataframes()` gets called.

src/code42cli/cmds/securitydata.py
```diff
@@ -362,84 +337,92 @@ def search(
...
+        cursor = _get_file_event_cursor_store(state.profile.name)
+        checkpoint = _handle_timestamp_checkpoint(cursor.get(use_checkpoint), state)
 
-def _construct_query(state, begin, end, saved_search, advanced_query, or_query):
+        def checkpoint_func(event):
+            cursor.replace(use_checkpoint, event["eventId"])
 
-    if advanced_query:
-        state.search_filters = advanced_query
-    if saved_search:
```
tora-kozic, line 413: I don't think this really matters - ideally only exactly one of these would be true at a time, but this could probably be `elif` ? I think `advanced-search`, `saved-search` and `begin/end` are all exclusive from each other

src/code42cli/cmds/securitydata.py
```diff
@@ -470,74 +457,49 @@ def show(state, search_id):
...
-        checkpoint = cursor.get(checkpoint_name) or ""
-
-        # older app versions stored checkpoint as float timestamp.
-        # we handle those here until the next run containing events will store checkpoint as the last eventId
-        try:
+def _construct_query(state, begin, end, saved_search, advanced_query, or_query):
+
+    if advanced_query:
+        state.search_filters = advanced_query
+    if saved_search:
```
tora-kozic, line 468: I don't think this really matters (I wrote it actually I think) - but since exactly one of these should be true at a time, for consistency this could probably be `elif` or all three cases could be their own `if`? 

I think `advanced-search`, `saved-search` and `begin/end` are all exclusive from each other

timabrmsn: yeah, this is just showing up in the diff cause I moved things around a bit. Although an `elif` makes it a bit more explicit that these states shouldn't ever overlap, I'll update it.

PR #343, opened 2021-12-02
Enable json deserialization of backup set data

timabrmsn:
The filename inclusions/exclusions in a device's backup set data are stored as a [TrackedFileSelectionList](https://github.com/code42/py42/blob/main/src/py42/clients/settings/device_settings.py#L455), which the json serializer doesn't know how to handle. This causes an error when running `code42 devices list-backup-sets -f JSON`. 

Converting these to regular lists when building the DataFrame resolves this.


tora-kozic (approved):
😎

PR #350, opened 2022-02-01
add device rename and bulk rename cmds

tora-kozic:
Adds commands to rename devices with the CLI.

Adds commands:
- `code42 devices rename <guid> --new-device-name <name>`
- `code42 devices bulk rename <csv>`

`--new-device-name` can be aliased as `-n`
CSV should contain columns `guid,name`

### DIFF
--- a/src/code42cli/cmds/devices.py
+++ b/src/code42cli/cmds/devices.py
@@ -38,6 +38,10 @@ device_guid_argument = click.argument(
     "device-guid", type=str, callback=lambda ctx, param, arg: _verify_guid_type(arg),
 )
 
+new_device_name_option = click.option(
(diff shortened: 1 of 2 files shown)


src/code42cli/cmds/devices.py
```diff
@@ -60,6 +64,20 @@ def change_device_name_option(help_msg):
 )
 
 
+@devices.command()
+@device_guid_argument
+@new_device_name_option
+@sdk_options()
+def rename(state, device_guid, new_device_name):
```
tora-kozic, line 71: Let me know if we want to add any other checks here (ie. should users not be able to rename a device in legal hold?)

timabrmsn (approved):
LGTM

PR #352, opened 2022-02-04
Chore/allow optional cols for bulk cmds

tora-kozic:
- Switched over the `code42 departing-employee bulk remove` and `code42 high-risk-employee bulk remove` commands to use the CSV input rather than the flat file because we wanted the header the input file to be optional.  Previously they took flat files as input. 

The CSV file reader already has optional headers if there's an exact number of columns matching the argument and allows for more columns than necessary if headers are provided.

- Added unit tests to check the optional header functionality.
- Included a short user guide about bulk commands in general and how their headers work.

### DIFF
--- a/docs/userguides/bulkcommands.md
+++ b/docs/userguides/bulkcommands.md
@@ -1,0 +1,23 @@
+# Using Bulk Commands
+
+Bulk functionality is available for many Code42 CLI methods, more details on which commands have bulk capabilities can be found in the [Commands Documentation](../commands.md).
+
+All bulk methods take a CSV file as input.
+
+The `generate-template` command can be used to create a CSV file with the necessary headers for a particular command.
+
+For instance, the following command will create a file named `devices_bulk_deactivate.csv` with a single column header row of `guid`.
+```bash
+code42 devices bulk generate-template deactivate
+```
+
+The CSV file can contain more columns than are necessary for the command, however then the header row is **required**.
+
+If the CSV file contains the *exact* number of columns that are necessary for the command then the header row is **optional**.
+
+To run a bulk method, simple pass the CSV file to the desired command.  For example, you would use to following command to deactivate multiple devices within your organization at once:
+
+
+```bash
+code42 devices bulk deactivate devices_bulk_deactivate.csv
+```
--- a/docs/userguides/bulkcommands.md
+++ b/docs/userguides/bulkcommands.md
@@ -13,9 +13,9 @@ code42 devices bulk generate-template deactivate
 
 The CSV file can contain more columns than are necessary for the command, however then the header row is **required**.
 
-If the CSV file contains the *exact* number of columns that are necessary for the command then the header row is **optional**.
+If the CSV file contains the *exact* number of columns that are necessary for the command then the header row is **optional**, but columns are expected to be in the same order as the template.
 
-To run a bulk method, simple pass the CSV file to the desired command.  For example, you would use to following command to deactivate multiple devices within your organization at once:
+To run a bulk method, simple pass the CSV file path to the desired command.  For example, you would use to following command to deactivate multiple devices within your organization at once:
 
 
 ```bash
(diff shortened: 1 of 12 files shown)


tora-kozic:
I'm curious if we'd want to remove the flat-file related methods as they're no longer being used anywhere after these changes, or if there's a potential for a future use case with them.

docs/userguides/bulkcommands.md
```diff
@@ -0,0 +1,23 @@
...
+The `generate-template` command can be used to create a CSV file with the necessary headers for a particular command.
+
+For instance, the following command will create a file named `devices_bulk_deactivate.csv` with a single column header row of `guid`.
+```bash
+code42 devices bulk generate-template deactivate
+```
+
+The CSV file can contain more columns than are necessary for the command, however then the header row is **required**.
+
+If the CSV file contains the *exact* number of columns that are necessary for the command then the header row is **optional**.
```
timabrmsn, line 16: let's add: `the header row is **optional**, but columns are expected to be in the same order as the template.`

src/code42cli/file_readers.py
```diff
@@ -28,6 +28,12 @@ def read_csv(file, headers):
     else error is raised.
     """
     lines = file.readlines()
+
+    # check if header is commented for flat-file backwards compatability
+    if lines[0].startswith("#"):
```
tora-kozic, line 33: This seems like it could be a little bit hacky, might just be better to either strip the comment line (this would require them to have no unnecessary columns) or throw a message.

docs/userguides/bulkcommands.md
```diff
@@ -0,0 +1,23 @@
...
+
+To run a bulk method, simple pass the CSV file to the desired command.  For example, you would use to following command to deactivate multiple devices within your organization at once:
```
timabrmsn, line 18: simpl**y** pass the CSV file **path** to the desired command

src/code42cli/file_readers.py
```diff
@@ -28,6 +28,12 @@ def read_csv(file, headers):
...
+        # strip comment character
+        lines[0] = (lines[0].split("#"))[1].strip()
```
timabrmsn, line 35: you can just use `.lstrip("#")` here

src/code42cli/cmds/departing_employee.py
```diff
@@ -15,7 +15,6 @@
 from code42cli.cmds.shared import get_user_id
 from code42cli.errors import Code42CLIError
 from code42cli.file_readers import read_csv_arg
-from code42cli.file_readers import read_flat_file_arg
```
timabrmsn, line 18: We can also delete this arg + function since they're no longer used anywhere.

tora-kozic: I mentioned it above but do you think it'd be fine to just remove all the methods related to the flat-files then?

src/code42cli/file_readers.py
timabrmsn: Since the flat file expectation was only a single column anyway, it would have failed if they had extra columns before this change. So maybe just removing the commented line would be best.

src/code42cli/cmds/departing_employee.py
timabrmsn: oops, missed that comment. yeah, let's remove them since the csv reader can handle the single column case just as well now.

tora-kozic:
Can you check over the `test_bulk.py` tests I adjusted, I don't have a super clear understanding of how the flat-file vs csv file reading worked previously

docs/userguides/bulkcommands.md
```diff
@@ -0,0 +1,23 @@
...
+For instance, the following command will create a file named `devices_bulk_deactivate.csv` with a single column header row of `guid`.
+```bash
+code42 devices bulk generate-template deactivate
+```
+
+The CSV file can contain more columns than are necessary for the command, however then the header row is **required**.
+
+If the CSV file contains the *exact* number of columns that are necessary for the command then the header row is **optional**, but columns are expected to be in the same order as the template.
+
+To run a bulk method, simple pass the CSV file path to the desired command.  For example, you would use to following command to deactivate multiple devices within your organization at once:
```
timabrmsn, line 18: simpl<y>

timabrmsn (approved):
one last simple typo, then LGTM

PR #355, opened 2022-02-15
implement user alias management

ceciliastevens:
Adds several methods to enable management of user [cloud aliases](https://developer.code42.com/api/#operation/UserControllerV2_AddCloudUsernames):

```
users add-alias <username> <alias>
users remove-alias <username> <alias>
users list-aliases
users bulk add-alias <csv>
users bulk remove-alias <csv>
```

Also updates tests and changelog appropriately.

### DIFF
--- a/src/code42cli/cmds/users.py
+++ b/src/code42cli/cmds/users.py
@@ -215,6 +215,8 @@ _bulk_user_move_headers = ["username", "org_id"]
 
 _bulk_user_roles_headers = ["username", "role_name"]
 
+_bulk_user_alias_headers = ["username", "alias"]
+
 
 @users.command(name="move")
 @username_option("The username of the user to move.", required=True)
@@ -224,6 +226,24 @@ def change_organization(state, username, org_id):
     """Change the organization of the user with the given username
     to the org with the given org UID."""
     _change_organization(state.sdk, username, org_id)
+
+
+@users.command()
+@click.argument("username")
+@click.argument("alias")
+@sdk_options()
+def add_alias(state, username, alias):
+    """Add a cloud alias for a given user."""
+    _add_cloud_alias(state.sdk, username, alias)
+
+
+@users.command()
+@click.argument("username")
+@click.argument("alias")
+@sdk_options()
+def remove_alias(state, username, alias):
+    """Remove a cloud alias for a given user."""
+    _remove_cloud_alias(state.sdk, username, alias)
 
 
 @users.group(cls=OrderedGroup)
@@ -292,6 +312,8 @@ users_generate_template = generate_template_cmd_factory(
     commands_dict={
         "update": _bulk_user_update_headers,
         "move": _bulk_user_move_headers,
+        "add-alias": _bulk_user_alias_headers,
+        "remove-alias": _bulk_user_alias_headers,
     },
     help_message="Generate the CSV template needed for bulk user commands.",
 )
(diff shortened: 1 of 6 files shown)


tests/cmds/test_users.py
```diff
@@ -1356,3 +1402,137 @@ def test_orgs_show_when_invalid_org_uid_raises_error(runner, cli_state, custom_e
     result = runner.invoke(cli, ["users", "orgs", "show", TEST_ORG_UID], obj=cli_state)
     assert result.exit_code == 1
     assert f"Invalid org UID {TEST_ORG_UID}." in result.output
+
+
+def test_add_cloud_alias_calls_add_cloud_alias_with_correct_parameters(
```
tora-kozic, line 1407: In general we often add test cases for the commands if they return any specific py42 errors.  Looking at the py42 methods, looks like `add_cloud_alias` will throw an error if the name is too long, so we'd probably want an `add-alias` case that checks for that error message.

tora-kozic: Since these methods also implement the `sdk.get_user()` py42 method, we also want to check that we return a helpful error if that user doesn't exist or the customer doesn't have permission to view them

src/code42cli/cmds/users.py
```diff
@@ -226,6 +228,24 @@ def change_organization(state, username, org_id):
     _change_organization(state.sdk, username, org_id)
 
 
+@users.command()
+@click.argument("username")
+@click.argument("alias")
+@sdk_options()
+def add_alias(state, username, alias):
+    """Add a cloud alias for a given user."""
```
tora-kozic, line 236: I'm curious if theres a use case to add more than one alias at a time, there's functionality to provide multiple click options at once (https://click.palletsprojects.com/en/8.0.x/options/#multiple-options) but if the CLI is mainly aimed towards doing this in bulk, then this should be fine.

tora-kozic:
Hey Cecilia! 

Thanks for putting this up, made a couple notes but overall looks great! If you want me to push some unit test commits just let me know.

src/code42cli/cmds/users.py
ceciliastevens: Users can have a maximum of two aliases, and have one by default that is the same as their username. I think that we should probably keep this method using a single alias both in order to keep parity with the console UI, and also to make it less likely that users will try to add more aliases than are allowed.

ceciliastevens:
> Hey Cecilia!
> 
> Thanks for putting this up, made a couple notes but overall looks great! If you want me to push some unit test commits just let me know. (I'm also working on fixing the github actions integration tests **Update**: These should be fixed now, you'll just have to catch up to `main`)

I'll add the additional tests you mention when I get a chance, but if you get to them before I do please feel free to push.

src/code42cli/cmds/users.py
tora-kozic: Got it, I didn't know we had a two maximum, then yes I agree this is probably the right methood.

tests/cmds/test_users.py
tora-kozic: I'm going to switch this method to use the `users` module of the sdk for consistency, and so we can use an existing CLI error we have for this use case.

tora-kozic: So it does look like the aliases are applicable to all users regardless of if theyre on a watchlist, so maybe this is more a py42 structure question, disregard my previous comments, lets leave these in the user group.

ceciliastevens:
Updated error handling again; it turns out the character limit is enforced by the UI, so I added a test for that so that we don't end up adding aliases that are longer than other parts of the system expect. Additionally, updated the error text to clarify that a request has failed specifically because the user already has more than two aliases.

tests/cmds/test_users.py
ceciliastevens: Yeah the structure is definitely a little bit odd, since the methods to add aliases exist in the context of detection lists, but they are used in other parts of Incydr as well. I think it makes sense to have these in the `users` command in the CLI because of that broader use, even though the backend is dealing with the detection lists service.

tora-kozic:
Doing some final reviewing on this PR and turns out the error handling around exceeding the number of cloud aliases on the py42 side for `add_cloud_alias()` isn't correct, so I'll be fixing that on the py42 side before we merge this one.

timabrmsn (approved):

PR #357, opened 2022-03-01
better help messages

tora-kozic:
Improve documentation around cloud alias methods.

### DIFF
--- a/src/code42cli/cmds/users.py
+++ b/src/code42cli/cmds/users.py
@@ -236,7 +236,7 @@ def change_organization(state, username, org_id):
 def add_alias(state, username, alias):
     """Add a cloud alias for a given user.
 
-    A cloud alias is the username an employee uses to access cloud services such as Google Drive or Box. Adding a cloud alias allows Incydr to link a user's cloud activity with their Code42 username. Each user has a default cloud alias of their Code42 username. You can add up to two additional aliases."""
+    A cloud alias is the username an employee uses to access cloud services such as Google Drive or Box. Adding a cloud alias allows Incydr to link a user's cloud activity with their Code42 username. Each user has a default cloud alias of their Code42 username. You can add one additional alias."""
     _add_cloud_alias(state.sdk, username, alias)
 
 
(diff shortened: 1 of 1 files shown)


src/code42cli/cmds/users.py
```diff
@@ -251,7 +253,7 @@ def remove_alias(state, username, alias):
 @click.argument("username")
 @sdk_options()
 def list_aliases(state, username):
-    """List the cloud aliases for a given user."""
+    """List the cloud aliases for a given user.  Each user has a default cloud alias of their Code42 username with up to two additional aliases."""
```
timabrmsn, line 256: I'm pretty sure it's a _total_ of two aliases, including the code42 username. So unless they remove the code42 username as an alias (possible to do, but not sure of any use case for it), they can only add one additional alias.

tora-kozic: I was also a little confused by this, but based on my own testing of the API calls while I was adding these commands, it seems to be 2 + the username?

tora-kozic: `(venv) tora.kozic@C02G1662ML85 code42cli % code42 users add-alias tora.kozic+workato@code42.com mynewalias
Error: Cloud alias limit exceeded. A max of 2 cloud aliases are allowed.
(venv) tora.kozic@C02G1662ML85 code42cli % code42 users list-aliases tora.kozic+workato@code42.com
['torakozic', 'tora.kozic', 'tora.kozic@code42.com']
`

timabrmsn: It only lets me add one. Maybe you had removed the username from the one it let you add 2 more to?

timabrmsn: odd, maybe the new user risk profile service behaves differently, and now that detectionlists are just passing through to the new service it's only letting me add one now?

tora-kozic: Huh, weird.  I'll look into it and adjust the docs

tora-kozic: So it definitely is what you were saying, but I have no idea why I can add 3 to myself

timabrmsn (approved):

PR #360, opened 2022-03-29
Chore/rename exception catch

tora-kozic:
- added exception catch for trying to rename an incydr device 
- upgrade black to v22.3.0

**NOTE**: testing will fail until the device settings changed are released in py42


timabrmsn (approved):
Looks good, let's just wait to merge until we can bump the minimum py42 version once that's released.

PR #367, opened 2022-04-28
Feature/watchlists

timabrmsn:
- deprecates `code42 departing-employee` and `code42 high-risk-employee` command groups
- adds `code42 watchlists list|list-users|add|remove|bulk-add|bulk-remove` commands
- adds `code42 users update-start-date|update-departure-date|update-risk-profile-notes` commands
- adds `code42 users bulk update-risk-profile` command
- updates `code42 users add-alias|remove-alias|bulk-add-alias|bulk-remove-alias` commands to use new `userriskprofile` methods from py42

### DIFF
--- a/src/code42cli/cmds/departing_employee.py
+++ b/src/code42cli/cmds/departing_employee.py
@@ -37,23 +37,17 @@ filter_option = click.option(
 )
 
 
-@click.group(cls=OrderedGroup)
+@click.group(cls=OrderedGroup, help=f"{DEPRECATION_TEXT}\n\nAdd and remove employees from the Departing Employees detection list.")
 @sdk_options(hidden=True)
 def departing_employee(state):
-    f"""{DEPRECATION_TEXT}
-
-    Add and remove employees from the Departing Employees detection list."""
     pass
 
 
-@departing_employee.command("list")
+@departing_employee.command("list", help=f"{DEPRECATION_TEXT}\n\nLists the users on the Departing Employees list.")
 @sdk_options()
 @format_option
 @filter_option
 def _list(state, format, filter):
-    f"""{DEPRECATION_TEXT}
-
-    Lists the users on the Departing Employees list."""
     deprecation_warning(DEPRECATION_TEXT)
     employee_generator = _get_departing_employees(state.sdk, filter)
     list_employees(
@@ -63,7 +57,7 @@ def _list(state, format, filter):
     )
 
 
-@departing_employee.command()
+@departing_employee.command(help=f"{DEPRECATION_TEXT}\n\nAdd a user to the Departing Employees detection list.")
 @username_arg
 @click.option(
     "--departure-date",
@@ -74,30 +68,22 @@ def _list(state, format, filter):
 @notes_option
 @sdk_options()
 def add(state, username, cloud_alias, departure_date, notes):
-    f"""{DEPRECATION_TEXT}
 
-    Add a user to the Departing Employees detection list."""
     deprecation_warning(DEPRECATION_TEXT)
     _add_departing_employee(state.sdk, username, cloud_alias, departure_date, notes)
 
 
-@departing_employee.command()
+@departing_employee.command(help=f"{DEPRECATION_TEXT}\n\nRemove a user from the Departing Employees detection list.")
 @username_arg
 @sdk_options()
 def remove(state, username):
-    f"""{DEPRECATION_TEXT}
-
-    Remove a user from the Departing Employees detection list."""
     deprecation_warning(DEPRECATION_TEXT)
     _remove_departing_employee(state.sdk, username)
 
 
-@departing_employee.group(cls=OrderedGroup)
+@departing_employee.group(cls=OrderedGroup, help=f"{DEPRECATION_TEXT}\n\nTools for executing bulk departing employee actions.")
 @sdk_options(hidden=True)
 def bulk(state):
-    f"""{DEPRECATION_TEXT}
-
-    Tools for executing bulk departing employee actions."""
     pass
 
 
--- a/src/code42cli/cmds/departing_employee.py
+++ b/src/code42cli/cmds/departing_employee.py
@@ -17,12 +17,14 @@ from code42cli.errors import Code42CLIError
 from code42cli.file_readers import read_csv_arg
 from code42cli.options import format_option
 from code42cli.options import sdk_options
+from code42cli.util import deprecation_warning
 
 
 def _get_filter_choices():
     filters = DepartingEmployeeFilters.choices()
     return get_choices(filters)
 
+DEPRECATION_TEXT = "(DEPRECATED): Use `code42 watchlists` commands instead."
 
 DATE_FORMAT = "%Y-%m-%d"
 filter_option = click.option(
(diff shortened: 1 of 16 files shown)


timabrmsn:
Tests are failing because it's depending on py42 code that isn't released yet.

src/code42cli/cmds/users.py
```diff
@@ -252,13 +254,69 @@ def remove_alias(state, username, alias):
...
+    state.sdk.userriskprofile.update(user_id, start_date=date)
+    
+
+@users.command()
+@click.argument("username")
+@click.argument("date", type=click.DateTime(formats=["%Y-%m-%d"]), required=False)
+@click.option("--clear", is_flag=True, help="Clears the current `end_date` value.")
+@sdk_options()
+def update_departure_date(state, username, date, clear):
+    """Sets the `end_date` on a User's risk profile (useful for users on the Departing Employee
```
tora-kozic, line 279: Super small - probably irrelevant but when I had him review my py42 docs John pointed out the me that the `Departing Employee` watchlist is now just titled `Departing` in the UI

src/code42cli/main.py
```diff
@@ -24,8 +25,11 @@
 from code42cli.cmds.shell import shell
 from code42cli.cmds.trustedactivities import trusted_activities
 from code42cli.cmds.users import users
+from code42cli.cmds.watchlists import watchlists
 from code42cli.options import sdk_options
 
+warnings.simplefilter("ignore", DeprecationWarning)
```
tora-kozic, line 31: Do we want deprecation warnings off by default?

tora-kozic (commented):
I assume we're not doing anything with the `watchlist_members()` endpoints until they release create/update/delete methods around them?

src/code42cli/cmds/watchlists.py
```diff
@@ -0,0 +1,250 @@
...
+    except Py42NotFoundError:
+        raise Code42CLIError(f"User ID {user} not found.")
+
+
+@watchlists.command()
+@click.option("--watchlist-id", help="ID of the watchlist.")
+@click.option(
+    "--watchlist-type",
+    type=click.Choice(WatchlistType.choices()),
+    help="Type of watchlist to remove user from. For CUSTOM watchlists, --watchlist-id is required.",
```
tora-kozic, line 104: Do we want to talk about custom watchlists before they're released? Or does that not really matter

tests/cmds/test_users.py
```diff
@@ -305,14 +307,14 @@ def add_alias_success(mocker, cli_state):
...
 @pytest.fixture
 def add_alias_limit_failure(mocker, cli_state):
-    cli_state.sdk.detectionlists.add_user_cloud_alias.side_effect = (
+    cli_state.sdk.userriskprofile.add_cloud_aliases.side_effect = (
         Py42CloudAliasLimitExceededError(create_mock_http_error(mocker))
     )
 
 
 @pytest.fixture
 def remove_alias_success(mocker, cli_state):
```
tora-kozic, line 316: There's an `add_alias_success` mock fixture that I think my be unnecessary (and is currently using the detectionlists client`

tora-kozic (changes requested):
add method docs and user guides 🍎

src/code42cli/cmds/departing_employee.py
```diff
@@ -17,12 +17,14 @@
...
 from code42cli.options import format_option
 from code42cli.options import sdk_options
+from code42cli.util import deprecation_warning
 
 
 def _get_filter_choices():
     filters = DepartingEmployeeFilters.choices()
     return get_choices(filters)
 
+DEPRECATION_TEXT = "(DEPRECATED): Use `code42 watchlists` commands instead."
```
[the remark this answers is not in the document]
timabrmsn: I had thought about listing out each replacement in each command help, but since they aren't 1-1 in all features didn't want to have to get into the differences. 

I think maybe I'll update the [detections list user-guide](https://clidocs.code42.com/en/stable/userguides/detectionlists.html) and lay out the replacements there, and then add that link to `DEPRECATION_TEXT`. 

This also reminded me I need to add a watchlists user-guide!

src/code42cli/cmds/watchlists.py
timabrmsn: I suppose I can leave this off for now.

src/code42cli/cmds/watchlists.py
```diff
@@ -0,0 +1,250 @@
...
+    cls=incompatible_with("watchlist_id"),
+)
+@click.argument("user", metavar="USER_ID|USERNAME")
+@sdk_options()
+def remove(state, watchlist_id, watchlist_type, user):
```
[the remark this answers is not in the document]
timabrmsn: I think add/remove is ok (although I noticed I forgot to add the help text to these!). 

We do have `list` for the watchlists themselves already, but I don't really see create/delete for watchlists being a common enough use-case to put it in the CLI.

src/code42cli/main.py
timabrmsn: Yep, otherwise the user would get deprecation warnings to stderr for the py42 method names, which could be confusing.

timabrmsn:
> I assume we're not doing anything with the `watchlist_members()` endpoints until they release create/update/delete methods around them?

ugh, I just realized I had I was thinking about these endpoints backwards and that we wouldn't likely need `members`. But I was wrong. I'll add `list-members` and rename `list-users` to `list-included-users` to make it explicit.

src/code42cli/cmds/watchlists.py
```diff
@@ -0,0 +1,252 @@
...
+)
+@click.option(
+    "--watchlist-type",
+    type=click.Choice(WatchlistType.choices()),
+    help="Type of watchlist to list.",
+    cls=incompatible_with("watchlist_id"),
+)
+@format_option
+@sdk_options()
+def list_members(state, watchlist_type, watchlist_id, format):
```
tora-kozic, line 51: I feel like we may still want a `list_included_users` command (or some way to view that data, maybe even just a `--only-show-included-users` flag?) so that people managing adding/removing individual users will have a way to see only the individuals that can be managed. May be confusing if someone sees someone whose a member but isn't able to remove them? 


We'd also want to specify the difference between the two in that case

tora-kozic, line 51: I feel like we may still want a `list_included_users` command (or some way to view that data, maybe even just a `--only-show-included-users` flag?) so that people managing adding/removing individual users will have a way to see only the individuals that can be managed. May be confusing if someone sees someone whose a member but isn't able to remove them? 


We'd also want to specify the difference between the two in that case

timabrmsn: Yep, that totally makes sense. good call.

timabrmsn: added.

tora-kozic: We should probably add some directions around it in the user guide too

tests/cmds/test_watchlists.py
```diff
@@ -0,0 +1,666 @@
...
+        )
+        assert res.exit_code == 1
+        assert "Error: Watchlist ID 'INVALID' not found." in res.output
+
+
+class TestWatchlistsRemoveCmd:
+    def test_missing_watchlist_identifying_option_raises_cli_error(
+        self, runner, cli_state
+    ):
+        res = runner.invoke(cli, ["watchlists", "add", "1234"], obj=cli_state)
```
tora-kozic, line 444: should this be `watchlists remove 1234` ?

tora-kozic (approved):
LGTM other than those last couple changes (and pending the py42 release obviously)

PR #372, opened 2022-06-08
fix a bug with devices list --include-settings

ceciliastevens:
There's a bug with devices list --include-settings caused by the addition of row.pop() to _process_csv_row. This fixes it by passing in a list of dicts as expected, instead of a list of strings.

I'm not sure if we're passing lists of strings to run_bulk_process() anywhere else, but if we are those should probably be updated as well.


ceciliastevens:
Looks like an integration test is failing? I'm not sure why though, I don't think I touched anything that it'd look at

timabrmsn (approved):

PR #380, opened 2022-07-05
adjust checkpoint filters when you want to retrieve all events

tora-kozic:
Appends a `riskSeverity.exists()` filter in the case of using a checkpoint with just the `--include-non-exposure` option (a query to retrieve all events). 

Going to add more descriptive error message for an `INVALID_PAGE_TOKEN` 400 error to py42 and should automatically be raised here.


src/code42cli/cmds/securitydata.py
```diff
@@ -513,4 +508,6 @@ def _handle_timestamp_checkpoint(checkpoint, state):
         state.search_filters.append(InsertionTimestamp.on_or_after(checkpoint))
         return None
     except (ValueError, TypeError):
-        return checkpoint
+        if checkpoint is not None:
+            return checkpoint
+        return ""
```
tora-kozic, line 513: If the query sends None as a page token, the `nextPgToken` is returned as null.  You have to pass an empty string to get the first page.

timabrmsn: We should move this logic into `_get_all_file_events` as we want to pass the empty string instead of `None` even when --use-checkpoint isn't passed, as that would cap the results of any non-checkpoint query at 10k results.

PR #382, opened 2022-07-25
Add v2 file events

tora-kozic:
Adds V2 file events:

- adds `--use-v2-file-events` flag to CLI profile settings
- if enabled, uses V2 filter terms
- if enabled, passes the `use_v2=True` flag to saved search methods


src/code42cli/config.py
```diff
@@ -98,7 +110,16 @@ def _set_username(self, new_value, profile):
...
 
     def _set_ignore_ssl_errors(self, new_value, profile):
-        profile[self.IGNORE_SSL_ERRORS_KEY] = str(new_value)
+        if profile[self.IGNORE_SSL_ERRORS_KEY] == str(new_value):
+            profile[self.IGNORE_SSL_ERRORS_KEY] = str(False)
+        else:
+            profile[self.IGNORE_SSL_ERRORS_KEY] = str(new_value)
+
+    def _set_use_v2_file_events(self, new_value, profile):
+        if profile[self.USE_V2_FILE_EVENTS_KEY] == str(new_value):
```
timabrmsn, line 119: We should use `.get()` to try to read the key here (and other places where we access the value), because on upgrading, existing profiles won't have this key and we then throw a KeyError. 

Also, I'm not sure I like the "flip whatever the current value is" logic for these when the flag is provided. I think I'd expect running `profile create -n new --use-v2-file-events ...` and `profile update -n new --use-v2-file-events --password <updated>` to both result in the profile using v2 events.

We should probably just move from an `is_flag=True` option to an explicit `type=click.types.BOOL`, so the user would do `--use-v2-file-events=true/false` to set it explicitly, and default to None so if the option isn't passed we leave it unchanged.

I'm ok with changing the ssl option here to that too, even though it's technically a breaking change, that option probably gets used so little it's not worth worrying about.

src/code42cli/cmds/securitydata.py
```diff
@@ -35,8 +39,55 @@
...
+        return arg
+
+    return callback
+
+
+def get_all_events_callback():
+    def callback(ctx, param, arg):
+        if not arg:
+            if ctx.obj.profile.use_v2_file_events == "True":
+                risk_values = [
```
timabrmsn, line 75: This is missing `NO_RISK_INDICATED`. 

We could also just do `v2_filters.risk.Severity.exists()` and that future-proofs this in the event that we add any new severity levels.

timabrmsn:
The `_get_saved_search_option` callback needs to check for v2 in the profile, otherwise if you have v2 set you'll get a v1 query back that we then try to send to the v2 endpoint.

src/code42cli/config.py
tora-kozic: Okay, cool, I'd definitely prefer to move away from the flag so that sounds good to me. 👍

src/code42cli/cmds/securitydata.py
tora-kozic: Ah maybe I misunderstood this filter? I'm under the impression that if `--include-non-exposure` is _not_ called, then this callback happens.  In the V1 case we make sure there's an exposure type, and in the V2 case I think the equivalent would be making sure there's some level of severity attached.  For sure let me know if I'm wrong on this though, was a little unclear on it

timabrmsn: Nope you're right, I forgot how that worked. This is correct.

timabrmsn: Though we could do `v2_filters.risk.Severity.not_eq(v2_filters.risk.Severity.NO_RISK_INDICATED)` to handle any future severity changes.

tora-kozic: Oh thats a good call, I like it

src/code42cli/cmds/securitydata.py
```diff
@@ -342,41 +405,72 @@ def search(
...
+
+            def checkpoint_func(event):
+                cursor.replace(use_checkpoint, event["eventId"])
 
     else:
         checkpoint = checkpoint_func = None
 
     query = _construct_query(state, begin, end, saved_search, advanced_query, or_query)
-    dfs = _get_all_file_events(state, query, checkpoint)
+    dfs = _get_all_file_events(state, query, checkpoint, format == OutputFormat.TABLE)
```
timabrmsn, line 473: We need to flatten for CSV output too, so this should be:

```
flatten = format in (OutputFormat.TABLE, OutputFormat.CSV)
dfs = _get_all_file_events(state, query, checkpoint, flatten)
```

timabrmsn (approved):
Apart from the CSV flattening this looks good to me!

PR #389, opened 2022-09-13
Feature/api clients

tora-kozic:
Adds API client support
- adds `code42 profile create-api-client` command 
- adds `--secret` and `--api-client-id` options to `code42 profile update` 
- slight changes to `code42 legal-hold list` and `code42 legal-hold show`

### DIFF
--- a/docs/userguides/profile.md
+++ b/docs/userguides/profile.md
@@ -3,7 +3,9 @@
 Use the [code42 profile](../commands/profile.md) set of commands to establish the Code42 environment you're working
 within and your user information.
 
-First, create your profile:
+## User token authentication
+
+Use the following command to create your profile with user token authentication:
 ```bash
 code42 profile create --name MY_FIRST_PROFILE --server example.authority.com --username security.admin@example.com
 ```
@@ -14,6 +16,19 @@ the program prompts you about storing a password. If you agree, you are then pro
 Your password is not shown when you do `code42 profile show`. However, `code42 profile show` will confirm that a
 password exists for your profile. If you do not set a password, you will be securely prompted to enter a password each
 time you run a command.
+
+## API client authentication
+
+Once you've generated an API Client in your Code42 console, use the following command to create your profile with API client authentication:
+```bash
+code42 profile create-api-client --name MY_API_CLIENT_PROFILE --server example.authority.com --api-client-id "key-42" --secret "code42%api%client%secret"
+```
+
+```{eval-rst}
+.. note:: Remember to escape special characters in your API client secret like `!` with a backslash. ex: `"my!secret"` should be passed in as `"my\!secret"`
+```
+
+## View profiles
 
 You can add multiple profiles with different names and the change the default profile with the `use` command:
 
--- a/docs/userguides/profile.md
+++ b/docs/userguides/profile.md
@@ -25,7 +25,7 @@ code42 profile create-api-client --name MY_API_CLIENT_PROFILE --server example.a
 ```
 
 ```{eval-rst}
-.. note:: Remember to escape special characters in your API client secret like `!` with a backslash. ex: `"my!secret"` should be passed in as `"my\!secret"`
+.. note:: Remember to wrap your API client secret with single quotes to avoid issues with bash expansion and special characters.
 ```
 
 ## View profiles
--- a/src/code42cli/cmds/legal_hold.py
+++ b/src/code42cli/cmds/legal_hold.py
@@ -4,6 +4,7 @@ from pprint import pformat
 
 import click
 from click import echo
+from click import style
 
 from code42cli.bulk import generate_template_cmd_factory
 from code42cli.bulk import run_bulk_process
@@ -89,7 +90,7 @@ def add_user(state, matter_id, username):
 @sdk_options()
 def remove_user(state, matter_id, username):
     """Release a custodian from a legal hold matter."""
-    _remove_user_from_legal_hold(state.sdk, matter_id, username)
+    _remove_user_from_legal_hold(state, matter_id, username)
 
 
 @legal_hold.command("list")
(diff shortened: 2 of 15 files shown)


src/code42cli/cmds/profile.py
```diff
@@ -120,28 +149,75 @@ def create(
...
+@profile.command()
+@name_option(required=True)
+@server_option(required=True)
+@api_client_id_option(required=True)
+@secret_option(required=True)
+@yes_option(hidden=True)
+@disable_ssl_option
+@use_v2_file_events_option
+@debug_option
+def create_api_client(
```
tora-kozic, line 179: I chose to make this a separate `create` command for a couple reasons
- py42's sdk uses two different auth methods
- the `totp` option isn't applicable with api-client auth
- I didn't think an optional password/secret (and in turn prompting the user for authentication on every command) made sense in the case of api clients.

Whether or not this should be a separate command or if we should just add options to the original `create` command is definitely open for discussion.

src/code42cli/cmds/profile.py
```diff
@@ -120,28 +149,75 @@ def create(
...
+@api_client_id_option()
+@secret_option()
 @username_option()
 @password_option
 @totp_option
 @disable_ssl_option
 @use_v2_file_events_option
+@yes_option(hidden=True)
 @debug_option
 def update(
```
tora-kozic, line 216: I have made this method.. so ugly

src/code42cli/config.py
```diff
@@ -105,24 +120,6 @@ def delete_profile(self, name):
             self._internal[self.DEFAULT_PROFILE] = self.DEFAULT_VALUE
         self._save()
 
-    def _set_authority_url(self, new_value, profile):
```
tora-kozic, line 108: simplified these one liners

src/code42cli/cmds/legal_hold.py
```diff
@@ -227,12 +242,12 @@ def _add_user_to_legal_hold(sdk, matter_id, username):
     sdk.legalhold.add_to_matter(user_id, matter_id)
 
 
-def _remove_user_from_legal_hold(sdk, matter_id, username):
-    _check_matter_is_accessible(sdk, matter_id)
+def _remove_user_from_legal_hold(state, matter_id, username):
```
timabrmsn, line 245: We had these functions take the sdk directly so that in bulk commands it wouldn't have to initialize the sdk in each thread (so as to make only 1 auth call instead of 5).

src/code42cli/cmds/profile.py
timabrmsn: yeah, a separate command makes sense, given there's enough differences. Making it explicit makes it easier to document and harder for a user to do something wrong unintentionally.

src/code42cli/cmds/profile.py
```diff
@@ -152,24 +228,100 @@ def update(
...
-        )
-    cliprofile.update_profile(
-        c42profile.name, server, username, disable_ssl_errors, use_v2_file_events
-    )
-    if not password and not c42profile.has_stored_password:
-        password = _prompt_for_password(c42profile.name)
-    if password:
-        _set_pw(name, password, debug, totp=totp)
+    if c42profile.api_client_auth == "True":
+        if (
```
timabrmsn, line 232: the `any()` method can make this a bit more readable:

```
if not any([server, api_client_id, secret, disable_ssl_errors, use_v2_file_events]):
    raise click.UsageError(
       ...
```

timabrmsn: Oh, I guess that won't work if someone passes `--disable-ssl-errors=false`

src/code42cli/cmds/legal_hold.py
timabrmsn: ah, I see, we need the state to check the profile. I suppose we could just initialize the sdk by accessing the property on the state before passing it in on the bulk commands then.

src/code42cli/cmds/legal_hold.py
```diff
@@ -241,37 +256,50 @@ def _get_and_print_preservation_policy(sdk, policy_uid):
...
-        matter["creator_username"] = matter["creator"]["username"]
+def _get_all_active_matters(state):
+    matters_generator = state.sdk.legalhold.get_all_matters()
+    if state.profile.api_client_auth == "True":
+        matters = [
+            matter for page in matters_generator for matter in page if matter["active"]
+        ]
+        for matter in matters:
+            try:
+                matter["creator_username"] = matter["creator"]["user"]["email"]
```
timabrmsn, line 291: This will fail if the matter was created via an API client, right?

docs/userguides/profile.md
```diff
@@ -15,6 +17,19 @@ Your password is not shown when you do `code42 profile show`. However, `code42 p
...
 
+## API client authentication
+
+Once you've generated an API Client in your Code42 console, use the following command to create your profile with API client authentication:
+```bash
+code42 profile create-api-client --name MY_API_CLIENT_PROFILE --server example.authority.com --api-client-id "key-42" --secret "code42%api%client%secret"
+```
+
+```{eval-rst}
+.. note:: Remember to escape special characters in your API client secret like `!` with a backslash. ex: `"my!secret"` should be passed in as `"my\!secret"`
```
timabrmsn, line 28: let's just suggest always wrapping the secret in single-quotes to prevent any shell expansion issues

src/code42cli/cmds/legal_hold.py
tora-kozic: I see, yeah I can look into updating the bulk methods

src/code42cli/cmds/legal_hold.py
tora-kozic: I'm not sure what the conditions for the field being populated are (but that would make sense).  This `creator user email` field isn't even included on the response documentation and isn't on every matter, but it does exist and is populated on some of them.

src/code42cli/cmds/profile.py
tora-kozic: I could at least group them a little I bet

docs/userguides/profile.md
tora-kozic: I did not know this about single vs double quotes in bash.. thats a good tip

timabrmsn (approved):

PR #392, opened 2022-10-07
chore/update-click-8.0

tora-kozic:
Update click to latest major version >=8.0.0. Updated to `v8.0.4`. This major version was released in spring of 2021.

  It appears that the only changes that affected the CLI tests is trivial changes in error messages.  That being said the [list of changes](https://click.palletsprojects.com/en/8.1.x/changes/#version-8-0-0) is significant so I'm going to read through it carefully before release to make sure we didn't miss anything.

Worth noting that click 8.1.0 removes support for python <3.7 so we would have to up our python version requirement before further updating.


setup.py
```diff
@@ -32,7 +32,7 @@
     python_requires=">=3.6.2, <4",
     install_requires=[
         "chardet",
-        "click>=7.1.1, <8",
+        "click>=8.0.0, <8.1.0",
```
timabrmsn, line 35: We don't need to have an upper-limit here. If someone _is_ still running python 3.6 `click>=8.1.0` simply won't get installed for them. And if 8.1.+ doesn't break anything in our code we don't need to artificially limit it if they are on 3.7+

timabrmsn: and we can also keep it `>=7.1.1` since we're not depending on anything 8+ specific.

tora-kozic: oh very cool

tora-kozic: our unit tests will fail on 7.1.1 by error message asserting but if we're okay with that then sure

timabrmsn: We could just make the asserts be an `or`, like 
```
assert ("Could not open file: not_a_file" in result.stdout) or (" Invalid value for '--advanced-query': 'not_a_file': No such file or directory" in result.stdout)
```

tora-kozic: I have no idea why I didn't think of that

PR #394, opened 2022-12-05
drop 3.6 and add 3.10/3.11 to python version matrix

timabrmsn:


tora-kozic:
Should we just remove the CLA assistant check as well? That hasn't worked in forever

PR #416, opened 2025-03-18
deprecate incydr functionality and remove guides

ceciliastevens:
Incydr functionality is deprecated in the code42cli. Users should use the Incydr CLI (available at https://developer.code42.com/) instead.

This PR also removes user guides that referred to deprecated functionality.


src/code42cli/cmds/alert_rules.py
```diff
@@ -31,11 +34,14 @@ class AlertRuleTypes:
 _HEADER_KEYS_MAP["ruleSource"] = "Source"
 _HEADER_KEYS_MAP["isEnabled"] = "Enabled"
 
+DEPRECATION_TEXT = "Incydr functionality is deprecated. Use the Incydr CLI"
```
tora-kozic, line 37: Looks like this is missing the end of the sentence that the others have?

tests/cmds/test_auditlogs.py
```diff
@@ -619,30 +619,30 @@ def test_search_if_error_occurs_when_processing_event_timestamp_does_not_store_e
...
-    output = output.split(" ")
-    output = [s for s in output if s]
-    assert (
-        output.count("Timestamp")
-        == output.count("ActorName")
-        == output.count("ActorIpAddress")
-        == output.count("AffectedUserUID")
-        == 1
-    )
+# def test_search_when_table_format_and_using_output_via_pager_only_includes_header_keys_once(
```
tora-kozic, line 622: why is this commented out?

tora-kozic (approved):
We may want to include a link to the incydr sdk package or at least the package name, maybe in the deprecation text or somewhere else?

tests/cmds/test_auditlogs.py
ceciliastevens: The test broke with the changes because it is just reading all of the output, so the addition of a warning message caused it to miss its expected header. The header is still there properly, and I figured it was probably best not to spend time updating a test for functionality that we're deprecating and don't intend to update.

src/code42cli/cmds/alert_rules.py
ceciliastevens: Fixed!

ceciliastevens:
> We may want to include a link to the incydr sdk package or at least the package name, maybe in the deprecation text or somewhere else?

I added a link to the dev portal in the deprecation text.

jdullum (approved):

tora-kozic (approved):

PR #417, opened 2025-03-21
prep 1.19.0 release

ceciliastevens:


tora-kozic:
We'll probably want to include the py42 version dependency update in this, so then it will have to come after that release

tora-kozic (approved):

